PR #7, opened 2023-02-27
stack.yaml fix

vikrantkm1997:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Stack.yaml Fix

### Additional Changes

## Checklist

### DIFF
--- a/Backend/stack.yaml
+++ b/Backend/stack.yaml
@@ -79,7 +79,7 @@ extra-deps:
     subdirs:
       - app/gateway
       - app/mock-registry
-  -  ../../shared-kernel/lib/mobility-core
+  # -  ../../shared-kernel/lib/mobility-core
   - git: git@github.com:Namma-Company/atlas-shared-kernel.git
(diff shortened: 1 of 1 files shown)


Backend/stack.yaml
```diff
@@ -79,7 +79,7 @@ extra-deps:
     subdirs:
       - app/gateway
       - app/mock-registry
-  -  ../../shared-kernel/lib/mobility-core
+  # -  ../../shared-kernel/lib/mobility-core
```
olegromashin, line 82: Why comment? We don't need this, it should be removed completely

DonDokidon: > Why comment? We don't need this, it should be removed completely

We need it to run changed kernel code without pushing it (and changing commit link, downloading it, compiling it etc.). Personally i use it pretty often, that's why it is commented here.

PR #14, opened 2023-02-27
Nixify, phase 1

srid:
Resolves #9 

This represents initial phase of Nix work, and subsequent work will happen in #30.

Note that developers will still be using Stack. Nix will be used in CI, including building of docker image (main branch only).

**IMPORTANT** Merge the following first before merging this PR.

- https://github.com/nammayatri/shared-kernel/pull/12
  - Then we must change `flake.nix` to point to `main` branch
- https://github.com/nammayatri/nammayatri/pull/12
- https://github.com/nammayatri/nammayatri/pull/32
- https://github.com/nammayatri/nammayatri/pull/64

PR #15, opened 2023-02-27
Add Github Actions CI for Nix

srid:
To merge into #14

I'm using my personal cachix (https://srid-nammayatri.cachix.org/) cache for now (pushing manually from laptop), but eventually once @vimalkumar or somebody else creates the official nammayatri cache, we should switch to it (at which point, the CI can push to cachix automatically). 

Regardless, we still need to manually push (made convenient via the flake app, run as `nix run .#cachix`) from macOS machines because Github Actions has no[^ci] ARM macOS builder: https://github.com/actions/runner-images/issues/2187

[^ci]: Instead of waiting for Microsoft to implement this, we could as well switch to either Jenkins or Hercules CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,41 @@
+name: "CI"
+on:
+  push:
+  pull_request:
+jobs:
+  Nix:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


srid:
Re-opening due to corrupt commit history: #16

PR #25, opened 2023-02-28
backend/feat/BKN-2917: exotel-heartbeat-webhook

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Currently, our dashboard is serving the purpose of handling the drivers or customer queries. 

We have faced one more requirement which dashboard can only solve. For our telephony service i.e. connecting drivers to customer or connecting customer to drivers we use exotel exo phones. To check health of the exophones if they are up or not, exotel provides one webhook.

[Product Heartbeat](https://developer.exotel.com/api/heart-beat#heartbeat-webhook) 

But we have just one common account for both BAP and BPP in exotel, so we can’t configure two separate webhooks. To handle this, we can create the webhook in dashboard, and then call the webhooks of BAP and BPP with same data if status of exophones are not “OK”. Call BAP/BAP once health becomes OK again

Keep the Request and Response type of BAP/BPP webhooks same as dashboard.

### Additional Changes

- [x] This PR modifies dhall configs/environment variables

nammayatri/Backend/dhall-configs/dev/provider-dashboard.dhall

## How did you test it?
Tested locally

## Checklist


pavelzarubin (approved):

Backend/app/dashboard/provider-dashboard/src/API/Exotel.hs
```diff
@@ -0,0 +1,83 @@
...
+  T.buildTransaction (DT.ExotelAPI endpoint) (Just serverName) Nothing Nothing Nothing
+
+-- store request and call bap/bpp only when status is not OK or previous status was not OK
+exotelHeartbeat :: Common.ExotelHeartbeatReq -> FlowHandler APISuccess
+exotelHeartbeat req = withFlowHandlerAPI $ do
+  forM_ [DSN.APP_BACKEND, DSN.BECKN_TRANSPORT, DSN.DRIVER_OFFER_BPP] $ \serverName -> do
+    fork ("exotelHeartbeat:" <> show serverName) $ do
+      needToCallApp <-
+        if req.statusType /= Common.OK
+          then pure True
```
hkmangla, line 62: Here also we can have check if last status is OK then only return True. So, ideally we should notify BAP/BPP if status is changing from OK to not OK and vice versa.

hkmangla (approved):

Backend/app/dashboard/provider-dashboard/src/API/Exotel.hs
roman-bodavskiy: But what if we have status not OK for some other phone number, like:

request1:
```
{  
   "timestamp": "2018-08-22T15:19:23Z>",
   "status_type":"WARNING",
   "incoming_affected":[ 
      "07930061010"
   ],
   "outgoing_affected": null,
   "data":{  
      "07930061010":{  
         "phone_number": "+917930061010",
      }
   }
}
```

request2:

```
{  
   "timestamp": "2018-08-22T15:19:23Z>",
   "status_type":"WARNING",
   "incoming_affected":[ 
      "someOtherNumber"
   ],
   "outgoing_affected": null,
   "data":{  
      "07930061010":{  
         "phone_number": "someOtherNumber",
      }
   }
}
```

Maybe we should store all requests which are not OK?

roman-bodavskiy: We can notify bap/bpp every time when something changed. Like, if other phone number was affected, or status changed from not OK to OK. What do you think @hkmangla ?

roman-bodavskiy: > We can notify bap/bpp every time when something changed. Like, if other phone number was affected, or status changed from not OK to OK.

Implemented this one and rebased. Please check and merge.

hkmangla (approved):

PR #33, opened 2023-02-28
backend/fix: #20 avoid tolls

shanm-16:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Since Auto-Rickshaws are not allowed on the NICE road in Bangalore, rides that estimate the fare based on a route that goes through the NICE road end up being discrepant since the driver eventually has to take another route.
To solve for this, we need to supply the parameter “avoid tolls” to the distance matrix API while estimating the price.
Also add a fallback to use the existing mechanism if the avoid tolls API call returns “No Routes”.
Add a log to track number of times the fallback was used.

### Additional Changes

## Motivation and Context

Since Auto-Rickshaws are not allowed on the NICE road in Bangalore, rides that estimate the fare based on a route that goes through the NICE road end up being discrepant since the driver eventually has to take another route.
To solve for this, we need to supply the parameter “avoid tolls” to the distance matrix API while estimating the price.
Also add a fallback to use the existing mechanism if the avoid tolls API call returns “No Routes”.
Add a log to track number of times the fallback was used.

## How did you test it?

by calling the api

## Checklist

### DIFF
--- a/Backend/stack.yaml
+++ b/Backend/stack.yaml
@@ -81,7 +81,7 @@ extra-deps:
       - app/mock-registry
   # -  ../../shared-kernel/lib/mobility-core
   - git: https://github.com/nammayatri/shared-kernel.git
-    commit: 28bae0f39eb3d2158c74dd3aac7ed685805c5cee
+    commit: eb9ea61fdd6f4b08f7d1e3c47a102204a06a7e6e
     subdirs:
       - lib/mobility-core
   - git: https://github.com/juspay/passetto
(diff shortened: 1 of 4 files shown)


hkmangla:
@shanm-16 Please link the shared-kernel PR here if there are changes there.

shanm-16:
shared kernel 

https://github.com/nammayatri/shared-kernel/pull/21

Backend/stack.yaml
```diff
@@ -81,7 +81,7 @@ extra-deps:
       - app/mock-registry
   # -  ../../shared-kernel/lib/mobility-core
   - git: git@github.com:nammayatri/shared-kernel.git
-    commit: 28bae0f39eb3d2158c74dd3aac7ed685805c5cee
+    commit: b3019f54fd00c8ac92b521149f6d0aab87b94b5b
```
hkmangla, line 84: Commit id changed. Update as per latest shared-kernel commit.

hkmangla (approved):

PR #86, opened 2023-03-06
Backend/refactor/ #BKN-2787 apply finalizer refactor

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

### Additional Changes

## Motivation and Context
https://juspay.atlassian.net/browse/BKN-2787

Refactored this PR by Pavel Zarubin
https://github.com/nammayatri/shared-kernel/pull/26
In my solution we don't need to make `SqlDB` monad type polymorphic and change all existing queries.
Instead I use existential quantification:

```
data SqlDBEnv = forall m.
  Finalize m =>
  SqlDBEnv
  { currentTime :: UTCTime,
    actions :: m ()
  }
```
## How did you test it?
Tested locally. Tested transaction inside of transaction. All finalizer actions worked as expected.

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
@@ -98,7 +98,7 @@ import SharedLogic.DriverPool as DP
 import SharedLogic.FareCalculator
 import qualified SharedLogic.MessageBuilder as MessageBuilder
 import Storage.CachedQueries.CacheConfig
-import qualified Storage.CachedQueries.DriverInformation as QDriverInformation
+import qualified Storage.CachedQueries.DriverInformation as CQDriverInformation
 import Storage.CachedQueries.FarePolicy (findByMerchantIdAndVariant)
 import qualified Storage.CachedQueries.Merchant as CQM
 import qualified Storage.CachedQueries.Merchant.TransporterConfig as CQTC
@@ -364,7 +364,7 @@ createDriverDetails personId adminId = do
             updatedAt = now
           }
   QDriverStats.createInitialDriverStats driverId
-  QDriverInformation.create driverInfo
+  CQDriverInformation.create driverInfo
   where
     driverId = cast personId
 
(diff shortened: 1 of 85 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -453,8 +454,9 @@ changeDriverEnableState admin personId isEnabled = do
     QPerson.findById personId
       >>= fromMaybeM (PersonDoesNotExist personId.getId)
   unless (person.merchantId == admin.merchantId) $ throwError Unauthorized
-  QDriverInformation.updateEnabledState driverId isEnabled
-  unless isEnabled $ QDriverInformation.updateActivity driverId False
+  Esq.runTransactionF $ \finalize -> do
+    CQDriverInformation.updateEnabledState finalize driverId isEnabled
+    unless isEnabled $ CQDriverInformation.updateActivity (Esq.ignoreFinalize finalize) driverId False
```
DonDokidon, line 459: Why finalize has to be ignored here? Anyway it would be better if you just pass `pure ()` here/ create `CQDriverInformation.updateActivity` version without finalize requirement. So `ignoreFinalize` is not required IMO

roman-bodavskiy: > Why finalize has to be ignored here?
Its ignored because we don't need to clear the same cache twice.

> Anyway it would be better if you just pass `pure ()` here

Not just `pure ()`, it should be like `const @(m()) (pure ())`, and `m` should be the same as external monad, so I guess `Esq.ignoreFinalize` is more clear. Anyway I can remove, its not a big deal

DonDokidon: Okay, leave it be. Can you remove finalize as argument in it then? it's not used anyway

roman-bodavskiy: Of course I can add second version `updateActivity'` without finalize. But its not very good, because everyone will use it and cache will be not cleared again. In my solution cache will be cleared by default, but if you don't want it, then you should say it explicitly

hkmangla:
Not required after KV changes.

Issue #189, opened 2023-03-17
Separating select queries


vikrantkm1997:
**Describe the bug**
```
SELECT "booking"."id", "booking"."quote_id", "booking"."status", "booking"."provider_id", "booking"."bap_id", "booking"."bap_uri", "booking"."start_time", "booking"."rider_id", "booking"."from_location_id", "booking"."to_location_id", "booking"."vehicle_variant", "booking"."estimated_distance", "booking"."estimated_fare", "booking"."estimated_duration", "booking"."fare_parameters_id", "booking"."rider_name", "booking"."created_at", "booking"."updated_at", "booking_location"."id", "booking_location"."lat", "booking_location"."lon", "booking_location"."street", "booking_location"."city", "booking_location"."state", "booking_location"."country", "booking_location"."building", "booking_location"."area_code", "booking_location"."area", "booking_location"."created_at", "booking_location"."updated_at", "booking_location2"."id", "booking_location2"."lat", "booking_location2"."lon", "booking_location2"."street", "booking_location2"."city", "booking_location2"."state", "booking_location2"."country", "booking_location2"."building", "booking_location2"."area_code", "booking_location2"."area", "booking_location2"."created_at", "booking_location2"."updated_at", "fare_parameters"."id", "fare_parameters"."base_fare", "fare_parameters"."dead_km_fare", "fare_parameters"."extra_km_fare", "fare_parameters"."driver_selected_fare", "fare_parameters"."night_shift_rate", "fare_parameters"."night_coef_included", "fare_parameters"."waiting_charge_per_min"
FROM "booking" INNER JOIN "booking_location" ON "booking"."from_location_id" = "booking_location"."id" INNER JOIN "booking_location" AS "booking_location2" ON "booking"."to_location_id" = "booking_location2"."id" INNER JOIN "fare_parameters" ON "booking"."fare_parameters_id" = "fare_parameters"."id" INNER JOIN "driver_quote" ON "booking"."quote_id" = "driver_quote"."id"
WHERE "driver_quote"."search_request_id" = 'daa2a3e6-047e-42de-bf6a-ca7fe0ffc3ba'
```
Here we are using join query but in the end fetching only one row.
Instead we should do separate select queries which will be more optimal since it would run on index

PR #191, opened 2023-03-17
backend/fix: #189 Separating select queries

vikrantkm1997:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
Separated select queries

### Additional Changes

## How did you test it?
Tested locally

## Checklist


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Booking.hs
```diff
@@ -61,20 +63,29 @@ findById bookingId = buildDType $
...
+
+buildFullBooking ::
+  Transactionable m =>
+  BookingT ->
+  DTypeBuilder m (Maybe (SolidType FullBookingT))
+buildFullBooking bookingT@BookingT {..} = runMaybeT $ do
+  fromLocationT <- MaybeT $ Esq.findById' @BookingLocationT (fromKey fromLocationId)
+  toLocationT <- MaybeT $ Esq.findById' @BookingLocationT (fromKey toLocationId)
+  fareParamsT <- MaybeT $ Esq.findById' @Fare.FareParametersT (fromKey fareParametersId)
+  return $ extractSolidType @Booking (bookingT, fromLocationT, toLocationT, fareParamsT)
```
DonDokidon, line 88: And why do we need this change at all? I thought that we already settled that separate selects are not faster than joins in any way (and even vice versa)

roman-bodavskiy: > And why do we need this change at all? I thought that we already settled that separate selects are not faster than joins in any way (and even vice versa)

I helped @vikrantkm1997 to separate this query. But I agree with you, proper join query should be faster, than many little queries.

@vikrantkm1997 did you do some performance tests?

NikithShetty: @DonDokidon @roman-bodavskiy 
We were analyzing the queries that are taking lot of time, this was one such query. Following is the analysis output

*Query*
```
SELECT "booking"."id", "booking"."quote_id", "booking"."status", "booking"."provider_id", "booking"."bap_id", "booking"."bap_uri", "booking"."start_time", "booking"."rider_id", "booking"."from_location_id", "booking"."to_location_id", "booking"."vehicle_variant", "booking"."estimated_distance", "booking"."estimated_fare", "booking"."estimated_duration", "booking"."fare_parameters_id", "booking"."rider_name", "booking"."created_at", "booking"."updated_at", "booking_location"."id", "booking_location"."lat", "booking_location"."lon", "booking_location"."street", "booking_location"."city", "booking_location"."state", "booking_location"."country", "booking_location"."building", "booking_location"."area_code", "booking_location"."area", "booking_location"."created_at", "booking_location"."updated_at", "booking_location2"."id", "booking_location2"."lat", "booking_location2"."lon", "booking_location2"."street", "booking_location2"."city", "booking_location2"."state", "booking_location2"."country", "booking_location2"."building", "booking_location2"."area_code", "booking_location2"."area", "booking_location2"."created_at", "booking_location2"."updated_at", "fare_parameters"."id", "fare_parameters"."base_fare", "fare_parameters"."dead_km_fare", "fare_parameters"."extra_km_fare", "fare_parameters"."driver_selected_fare", "fare_parameters"."night_shift_rate", "fare_parameters"."night_coef_included", "fare_parameters"."waiting_charge_per_min"
FROM "booking" INNER JOIN "booking_location" ON "booking"."from_location_id" = "booking_location"."id" INNER JOIN "booking_location" AS "booking_location2" ON "booking"."to_location_id" = "booking_location2"."id" INNER JOIN "fare_parameters" ON "booking"."fare_parameters_id" = "fare_parameters"."id" INNER JOIN "driver_quote" ON "booking"."quote_id" = "driver_quote"."id"
WHERE "driver_quote"."search_request_id" = 'daa2a3e6-047e-42de-bf6a-ca7fe0ffc3ba'
```

*Analysis*
```
                                                                                 QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=1014.14..107091.41 rows=1 width=774) (actual time=117.513..146.223 rows=1 loops=1)
   ->  Nested Loop  (cost=1013.59..107090.61 rows=1 width=704) (actual time=117.488..146.197 rows=1 loops=1)
         ->  Nested Loop  (cost=1013.04..107089.76 rows=1 width=559) (actual time=117.465..146.173 rows=1 loops=1)
               ->  Gather  (cost=1012.49..107088.91 rows=1 width=414) (actual time=117.416..146.122 rows=1 loops=1)
                     Workers Planned: 2
                     Workers Launched: 2
                     ->  Hash Join  (cost=12.49..106088.81 rows=1 width=414) (actual time=129.269..138.016 rows=0 loops=3)
                           Hash Cond: (booking.quote_id = driver_quote.id)
                           ->  Parallel Seq Scan on booking  (cost=0.00..105123.15 rows=363115 width=414) (actual time=0.007..107.870 rows=290042 loops=3)
                           ->  Hash  (cost=12.46..12.46 rows=2 width=37) (actual time=0.022..0.022 rows=1 loops=3)
                                 Buckets: 1024  Batches: 1  Memory Usage: 9kB
                                 ->  Index Scan using idx_quote_search_req_id on driver_quote  (cost=0.43..12.46 rows=2 width=37) (actual time=0.018..0.018 rows=1 loops=3)
                                       Index Cond: (search_request_id = 'daa2a3e6-047e-42de-bf6a-ca7fe0ffc3ba'::bpchar)
               ->  Index Scan using booking_location_pkey on booking_location  (cost=0.55..0.85 rows=1 width=145) (actual time=0.042..0.043 rows=1 loops=1)
                     Index Cond: (id = booking.from_location_id)
         ->  Index Scan using booking_location_pkey on booking_location booking_location2  (cost=0.55..0.85 rows=1 width=145) (actual time=0.020..0.020 rows=1 loops=1)
               Index Cond: (id = booking.to_location_id)
   ->  Index Scan using fare_parameters_pkey on fare_parameters  (cost=0.55..0.80 rows=1 width=70) (actual time=0.021..0.021 rows=1 loops=1)
         Index Cond: (id = booking.fare_parameters_id)
 Planning Time: 1.728 ms
 Execution Time: 146.339 ms
(21 rows)
```

As you can see, the query is scanning the entire `Booking` table which is a hot spot which is taking lot of time.
We have indices on all the tables above, so running point select queries and joining the results in code will be faster than this join query.

This is the general finding we had, some join queries are not using indices on all tables which is causing the problem. Hence the understanding is that, for point lookup queries which return one row, we can use select queries and for bulk lookup queries that return multiple rows we can use join queries.

This also aligns with our long term plan to move to KV architecture where search/booking/tracking (for the duration of active booking) operations happen in KV store and then its synced to DB slowly, which allows for highly scalable system.
The constraint of this architecture is its difficult to model joins, but we its easier to model select queries.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Booking.hs
```diff
@@ -61,20 +63,29 @@ findById bookingId = buildDType $
...
-        from
-          ( baseBookingTable
-              `innerJoin` table @DriverQuote.DriverQuoteT
-              `Esq.on` ( \(rb :& _ :& _ :& _ :& dq) ->
-                           rb ^. BookingQuoteId ==. dq ^. DriverQuoteTId
-                       )
-          )
-      where_ $ dq ^. DriverQuoteSearchRequestId ==. val (toKey searchReqId)
-      pure (rb, bFromLoc, bToLoc, farePars)
+findBySearchReq' :: (Transactionable m) => Id DSR.SearchRequest -> m (Maybe Booking)
```
NikithShetty, line 66: why is `'` required in function name? There is only one function of this type, right?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverQuote.hs
```diff
@@ -100,3 +100,9 @@ deleteByDriverId personId =
   Esq.delete $ do
     driverQuotes <- from $ table @DriverQuoteT
     where_ $ driverQuotes ^. DriverQuoteDriverId ==. val (toKey personId)
+
+findDriverQuoteBySearchId' :: Transactionable m => Id DSReq.SearchRequest -> DTypeBuilder m (Maybe DriverQuoteT)
```
NikithShetty, line 104: same here

NikithShetty (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverQuote.hs
roman-bodavskiy: In this case we can consider function with `'` for internal use, because it returns `TType`, not domain type

NikithShetty (approved):

PR #261, opened 2023-03-22
Jenkins: push to cachix automatically

srid:
(We had disabled `cachix push` for CI security reasons)


srid:
Better approach incoming.

PR #262, opened 2023-03-22
Jenkins: simplify docker image name detection

srid:


srid:
Better approach incoming.

PR #284, opened 2023-03-23
Jenkins: use Groovy library for cachix/docker

srid:
Using: https://github.com/juspay/jenkins-nix-ci/pull/6

Going forward, we'll be using these library functions for doing common tasks. These functions, being located in [the same repo](https://github.com/juspay/jenkins-nix-ci), are automatically aware of the associated credentials; so we do not have to provide them ourselves. Thus it simplifies our `Jenkinsfile`.

### DIFF
--- a/Backend/nix/docker.nix
+++ b/Backend/nix/docker.nix
@@ -8,8 +8,6 @@ let
 in
 {
   config = {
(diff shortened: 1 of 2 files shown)


srid:
I'll do this on the other two repos shortly.

hkmangla (approved):

Issue #342, opened 2023-03-28
Like and View Count for drivers


Pranav161996:
In the current version of in app alerts/messages we do not have any engagement except comments from the driver. 

We need to integrate the feature that a driver can like a message along with comment, we will then show this view and like count to all the drivers just like instagram/twitter posts/reels. This can improve the adoption as drivers may open messages just because they see high like counts too.

**How it will work:**
They should be able to view the counts outside before opening the message, but can like only after opening the message (live +1 in count).

As we plan to introduce, driver made content in next phase which other drivers can check out and interact, this can be the start of that interactions

PR #379, opened 2023-03-30
backend/feat #342: like and view count for drivers

AshishSaini25:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

I have added  a feature like and view count for drivers

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Message.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Message.hs
@@ -29,6 +29,7 @@ import Kernel.Types.Id
 import Kernel.Utils.Common
 import Kernel.Utils.JSON (stripPrefixUnderscoreIfAny)
 import qualified Storage.Queries.MediaFile as MFQ
+import qualified Storage.Queries.Message.Message as MQ
 import qualified Storage.Queries.Message.MessageReport as MRQ
 import qualified Storage.Queries.Person as QP
 import Tools.Error
@@ -47,6 +48,8 @@ data MessageAPIEntityResponse = MessageAPIEntityResponse
     label :: Maybe Text,
     reply :: Maybe Text,
     readStatus :: Bool,
+    likeStatus :: Bool,
+    likeCount :: Int,
     messageId :: Id Domain.Message,
     mediaFiles :: [MediaFileApiResponse]
   }
@@ -75,6 +78,8 @@ messageList driverId mbLimit mbOffset = do
             reply = messageReport.reply,
             created_at = rawMessage.createdAt,
             readStatus = messageReport.readStatus,
+            likeStatus = messageReport.likeStatus,
+            likeCount = rawMessage.likeCount,
             messageId = rawMessage.id,
             mediaFiles = mediaFilesApiType
           }
@@ -99,6 +104,8 @@ getMessage driverId messageId = do
             reply = messageReport.reply,
             created_at = rawMessage.createdAt,
             readStatus = messageReport.readStatus,
+            likeStatus = messageReport.likeStatus,
+            likeCount = rawMessage.likeCount,
             messageId = rawMessage.id,
             mediaFiles = mediaFilesApiType
           }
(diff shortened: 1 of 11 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Message.hs
```diff
@@ -112,6 +116,16 @@ messageSeen :: Id SP.Person -> Id Domain.Message -> Flow APISuccess
...
   Esq.runTransaction $ MRQ.updateSeenAndReplyByMessageIdAndDriverId messageId (cast driverId) True Nothing
+
+  return Success
+
+messageLiked :: Id SP.Person -> Id Domain.Message -> Flow APISuccess
+messageLiked driverId messageId = do
+  _ <- Esq.runInReplica $ QP.findById driverId >>= fromMaybeM (PersonNotFound driverId.getId)
+  Esq.runTransaction $ MRQ.updateMessageLikeByMessageIdAndDriverIdAndReadStatus messageId (cast driverId)
+  messageDetails <- Esq.runInReplica $ MRQ.findByMessageIdAndDriverId messageId (cast driverId) >>= fromMaybeM (InvalidRequest "Message not found")
+  let val = if messageDetails.likeStatus then 1 else (-1)
```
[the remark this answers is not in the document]
AshishSaini25: if driver unlike it after like then we have to decrease the like count by 1 therefore -1

Backend/dev/migrations/dynamic-offer-driver-app/0135-add-like-count-in-message.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE atlas_driver_offer_bpp.message_report ADD COLUMN like_status BOOLEAN DEFAULT false;
+ALTER TABLE atlas_driver_offer_bpp.message ADD COLUMN like_count Int DEFAULT 0;
```
roman-bodavskiy, line 2: both should be `NOT NULL`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Message.hs
```diff
@@ -112,6 +116,16 @@ messageSeen :: Id SP.Person -> Id Domain.Message -> Flow APISuccess
...
   _ <- Esq.runInReplica $ QP.findById driverId >>= fromMaybeM (PersonNotFound driverId.getId)
   Esq.runTransaction $ MRQ.updateSeenAndReplyByMessageIdAndDriverId messageId (cast driverId) True Nothing
+
+  return Success
+
+messageLiked :: Id SP.Person -> Id Domain.Message -> Flow APISuccess
+messageLiked driverId messageId = do
+  _ <- Esq.runInReplica $ QP.findById driverId >>= fromMaybeM (PersonNotFound driverId.getId)
+  Esq.runTransaction $ MRQ.updateMessageLikeByMessageIdAndDriverIdAndReadStatus messageId (cast driverId)
+  messageDetails <- Esq.runInReplica $ MRQ.findByMessageIdAndDriverId messageId (cast driverId) >>= fromMaybeM (InvalidRequest "Message not found")
```
roman-bodavskiy, line 126: You shouldn't use read query from replica db right after transaction, because value can update very slowly.

Instead you can fetch all required values from replica DB, and then update all required values within one transaction

Backend/dev/migrations/dynamic-offer-driver-app/0135-add-like-count-in-message.sql
AshishSaini25: like every message in the beginning  will have 0 like and like_status for that particular message will be false , so if keep it NOT NULL then we have to explicitly assign these value.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Message.hs
AshishSaini25: ok

Backend/dev/migrations/dynamic-offer-driver-app/0135-add-like-count-in-message.sql
AshishSaini25: I have given default value for both the column so there will not be a Null value at any time so what is the need of this constraint ?

roman-bodavskiy: Why not? anyway it would not worse with constraint

AshishSaini25: understood

roman-bodavskiy (approved):

PR #361, opened 2023-03-28
Nixify development environment

srid:
Resolves #356 

What's left to do: remove stack, Makefile, shell scripts and ignore them in Git.

PR #395, opened 2023-03-31
Fix trailing whitespaces

srid:
Doesn't look like the existing pre-commit was actually run before the commit. That, or the shell script is buggy.

This will be checked as part of CI in #361


roman-bodavskiy (approved):

srid:
I restarted the Jenkins build; the CI passes.

hkmangla (approved):

Issue #415, opened 2023-04-04
Deletion of duplicate driver if checks pass during auth in alternate number


JanhaviSoni:
**Describe the bug**
A clear and concise description of what the bug is.Deletion of duplicate driver if checks pass during auth in alternate number

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**App (please complete the following information):**

**Additional context**

PR #422, opened 2023-04-04
backend/fix : #415 Delete driver based on checks and allow alternate number input

JanhaviSoni:
…number input

## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Delete driver based on check and allow alternate number input

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
@@ -69,6 +69,7 @@ import Domain.Types.Vehicle (VehicleAPIEntity)
 import qualified Domain.Types.Vehicle as SV
 import qualified Domain.Types.Vehicle as Veh
 import qualified Domain.Types.Vehicle.Variant as Variant
+import Environment
 import EulerHS.Prelude hiding (id, state)
 import GHC.Records.Extra
 import Kernel.External.Encryption
@@ -95,6 +96,7 @@ import Kernel.Utils.SlidingWindowLimiter
 import Kernel.Utils.Validation
 import SharedLogic.CallBAP (sendDriverOffer)
 import qualified SharedLogic.CancelSearch as CS
+import qualified SharedLogic.DeleteDriver as DeleteDriverOnCheck
 import SharedLogic.DriverPool as DP
 import SharedLogic.FareCalculator
 import qualified SharedLogic.MessageBuilder as MessageBuilder
@@ -854,7 +856,15 @@ validate personId phoneNumber = do
   altNoAttempt <- runInReplica $ QRegister.getAlternateNumberAttempts personId
   runRequestValidation validationCheck phoneNumber
   mobileNumberHash <- getDbHash phoneNumber.alternateNumber
-  duplicateCheck (QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId) "Person with this mobile number already exists"
+  merchant <- CQM.findById person.merchantId >>= fromMaybeM (MerchantNotFound person.merchantId.getId)
+  mbPerson <- QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId
+  deleteOldPersonCheck <- case mbPerson of
+    Nothing -> return False
+    Just oldPerson -> do
+      when (oldPerson.id == person.id) $ throwError $ InvalidRequest "Alternate number already linked"
+      DeleteDriverOnCheck.validateDriver merchant oldPerson
+  logDebug $ "Delete Driver Check" <> show deleteOldPersonCheck
+  when deleteOldPersonCheck $ throwError $ InvalidRequest "Alternate number can't be validated"
   smsCfg <- asks (.smsCfg)
   let useFakeOtpM = useFakeSms smsCfg
   otpCode <- maybe generateOTPCode return (show <$> useFakeOtpM)
@@ -874,8 +884,6 @@ validate personId phoneNumber = do
   let verified = False
   cacheAlternateNumberInfo personId phoneNumber.alternateNumber otpCode altNoAttempt verified
   return $ DriverAlternateNumberRes {attempts = altNoAttempt}
-  where
-    duplicateCheck cond err = whenM (isJust <$> cond) $ throwError $ InvalidRequest err
 
 validateAuthVerifyReq :: Validate DriverAlternateNumberOtpReq
 validateAuthVerifyReq DriverAlternateNumberOtpReq {..} =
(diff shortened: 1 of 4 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -854,7 +856,9 @@ validate personId phoneNumber = do
   altNoAttempt <- runInReplica $ QRegister.getAlternateNumberAttempts personId
   runRequestValidation validationCheck phoneNumber
   mobileNumberHash <- getDbHash phoneNumber.alternateNumber
-  duplicateCheck (QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId) "Person with this mobile number already exists"
+  merchant <- CQM.findById person.merchantId >>= fromMaybeM (InvalidRequest "Merchant not found")
+  mbPerson <- QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId
+  whenJust mbPerson $ DeleteDriverOnCheck.merchantAccessCheck merchant
```
hkmangla, line 861: merchantAccessCheck should just return one boolean and then error should be thrown accordingly instead of this function itself throwing the error.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -917,10 +911,19 @@ verifyAuth personId req = do
...
               SP.alternateMobileNumber = Just encNewNum
             }
+        mobileCountryCode = fromMaybe "+91" person.mobileCountryCode
+    mobileNumberHash <- getDbHash altMobNo
+    mbPerson <- QPerson.findByMobileNumberAndMerchant mobileCountryCode mobileNumberHash person.merchantId
     Esq.runTransaction $ do
       QPerson.updateAlternateMobileNumberAndCode driver
     expTime <- fromIntegral <$> asks (.cacheConfig.configsExpTime)
     Redis.setExp (makeAlternateNumberVerifiedKey personId) True expTime
+    when (isJust mbPerson) $ do
```
hkmangla, line 921: instead of using inJust using use case.

hkmangla (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -854,7 +856,18 @@ validate personId phoneNumber = do
   altNoAttempt <- runInReplica $ QRegister.getAlternateNumberAttempts personId
   runRequestValidation validationCheck phoneNumber
   mobileNumberHash <- getDbHash phoneNumber.alternateNumber
-  duplicateCheck (QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId) "Person with this mobile number already exists"
+  merchant <- CQM.findById person.merchantId >>= fromMaybeM (InvalidRequest "Merchant not found")
+  mbPerson <- QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId
+  deleteOldPersonCheck <-
+    if (isJust mbPerson)
```
hkmangla, line 862:
```
    if (isJust mbPerson)
      then do
        oldPerson <- case mbPerson of
          Nothing -> throwError $ InvalidRequest "Old Person not found"
          Just existingPerson -> return existingPerson
        DeleteDriverOnCheck.merchantAccessCheck merchant oldPerson
      else return False
```

```
      case mbPerson of
        Nothing -> return false
        Just person -> DeleteDriverOnCheck.merchantAccessCheck merchant person
```

hkmangla (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DeleteDriver.hs
```diff
@@ -0,0 +1,84 @@
...
+    QDriverQuote.deleteByDriverId reqDriverId
+    QSearchReqForDriver.deleteByDriverId reqDriverId
+    QDriverStats.deleteById (cast reqDriverId)
+    QDriverLocation.deleteById reqDriverId
+    QR.deleteByPersonId reqDriverId
+    QVehicle.deleteById reqDriverId
+    QDriverInfo.deleteById (cast reqDriverId)
+    QDriverFlowStatus.deleteById reqDriverId
+    QMessage.deleteByPersonId reqDriverId
+    QPerson.deleteById reqDriverId
```
DonDokidon, line 74: Why don't we have deletetion from Ride table among these queries? Don't we have Ride -> Person ref? If usage of this func expected to be only on drivers, which don't have any rides, then we should indicate it somehow in error or something like that.
Personally i think it's not a good idea to let anyone except Dashboard user to delete driver accounts, but that is up to what you want... We should add a feature to disable drivers without removing the entry from db entirely (mb with removing personal infos like a name and phone num)

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -917,10 +915,16 @@ verifyAuth personId req = do
...
             }
+        mobileCountryCode = fromMaybe "+91" person.mobileCountryCode
+    mobileNumberHash <- getDbHash altMobNo
+    mbPerson <- QPerson.findByMobileNumberAndMerchant mobileCountryCode mobileNumberHash person.merchantId
     Esq.runTransaction $ do
       QPerson.updateAlternateMobileNumberAndCode driver
     expTime <- fromIntegral <$> asks (.cacheConfig.configsExpTime)
     Redis.setExp (makeAlternateNumberVerifiedKey personId) True expTime
+    whenJust mbPerson $ \oldPerson -> do
+      merchant <- CQM.findById person.merchantId >>= fromMaybeM (InvalidRequest "Merchant not found")
```
DonDokidon, line 926: We have proper exception for this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -854,7 +856,13 @@ validate personId phoneNumber = do
   altNoAttempt <- runInReplica $ QRegister.getAlternateNumberAttempts personId
   runRequestValidation validationCheck phoneNumber
   mobileNumberHash <- getDbHash phoneNumber.alternateNumber
-  duplicateCheck (QPerson.findByMobileNumberAndMerchant phoneNumber.mobileCountryCode mobileNumberHash person.merchantId) "Person with this mobile number already exists"
+  merchant <- CQM.findById person.merchantId >>= fromMaybeM (InvalidRequest "Merchant not found")
```
DonDokidon, line 859: Exception

DonDokidon (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DeleteDriver.hs
JanhaviSoni: This function is used for the reason that during registration if the driver has completed login using one number and he tries to login using another number but while uploading DL/RC he uploads same as previous one so we don't allow him but the data is saved as another driver's data with mobile number (incomplete registration) so if he enters his this number as alternate number previously since the number is present in Db we were not allowing him but now we are using this function to delete the driver who passes the checks for deletion and allow driver to enter alternate number

hkmangla: @DonDokidon Correct. We should not delete any driver other than dashboard. I was also thinking the same. 
But this is very tricky case. We have introduced alternate number flow recently where a driver can have an alternate number considering that alternate number is not registered with us. But in case, driver mistakenly logged in using his alternate number then he will not be able to add this number as it is already registered. 

To handle this case, driver should call our operations team and get this extra entry deleted. But as this will be very operational heavy and we don't have enough number of operations staff, so we're automating this process by deleting it after OTP verification.

hkmangla (changes requested):

hkmangla (approved):

PR #458, opened 2023-04-07
Backend/fix: #98/added estimatedDistance parameter for shareRide api

prashant601:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Ride.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Ride.hs
@@ -29,7 +29,7 @@ import Kernel.Prelude
 import Kernel.Storage.Esqueleto hiding (isNothing)
 import Kernel.Types.Error
 import Kernel.Types.Id
-import Kernel.Utils.Error
+import Kernel.Utils.Common
 import Storage.CachedQueries.Merchant (findByShortId)
 import qualified Storage.Queries.Booking as QRB
 import qualified Storage.Queries.Person as QP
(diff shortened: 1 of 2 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Ride.hs
```diff
@@ -52,6 +53,9 @@ mkCommonBookingLocation BookingLocation {..} =
       ..
     }
 
+mkCommonEstimateDistance :: HighPrecMeters -> Common.HighPrecMeters
+mkCommonEstimateDistance HighPrecMeters {..} = Common.HighPrecMeters {..}
```
roman-bodavskiy, line 57: Isn't `HighPrecMeters` and `Common.HighPrecMeters` the same?
If not, why do we need two `HighPrecMeters` types?

prashant601: they are same , but we need to cast it because we are using it in the dashboard side.

hkmangla: Why to use different type in dashboard @prashant601 ? We can just use the same Kernel type.
Type case is mostly for the types which are defined in either bap and bpp but for the types defined in Kernel, we don't need to do type cast. We can just import them from kernel.

hkmangla (changes requested):

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Ride.hs
prashant601: Okay, got it. 
I have made the changes accordingly.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Ride.hs
```diff
@@ -73,6 +73,10 @@ shareRideInfo merchantId rideId = do
         DB.OneWayDetails locationDetail -> Just $ mkCommonBookingLocation locationDetail.toLocation
         DB.DriverOfferDetails driverOfferDetail -> Just $ mkCommonBookingLocation driverOfferDetail.toLocation
         _ -> Nothing
+  let mbDistance = case booking.bookingDetails of
+        DB.OneWayDetails locationDetail -> Just $ locationDetail.distance
+        DB.DriverOfferDetails driverOfferDetail -> Just $ driverOfferDetail.distance
+        _ -> Nothing
```
roman-bodavskiy, line 79: in `OneWaySpecialZoneDetails` details we also have `distance` field

prashant601: Done

hkmangla (approved):

Issue #417, opened 2023-04-04
Add app version in prometheus metrics


NikithShetty:
Prometheus metrics being generated by the container needs to have additional `version` label which will be used for filtering out metrics.

`version` can be fetched from `DEPLOYMENT_VERSION` env variable.

Magizhan:
@NikithShetty Are we adding both the apk and the bundle version ?

Issue #336, opened 2023-03-28
Separating select queries


vikrantkm1997:
**Separating Select queries for list of quotes**
```
SELECT "quote"."id", "quote"."fare_product_type", "quote"."request_id", "quote"."estimated_fare", "quote"."discount", "quote"."estimated_total_fare", "quote"."provider_id", "quote"."provider_url", "quote"."provider_name", "quote"."provider_mobile_number", "quote"."provider_completed_rides_count", "quote"."distance_to_nearest_driver", "quote"."vehicle_variant", "quote"."trip_terms_id", "quote"."rental_slab_id", "quote"."driver_offer_id", "quote"."merchant_id", "quote"."special_zone_quote_id", "quote"."created_at", "trip_terms"."id", "trip_terms"."descriptions", "rental_slab"."id", "rental_slab"."base_distance", "rental_slab"."base_duration", "driver_offer"."id", "driver_offer"."estimate_id", "driver_offer"."driver_name", "driver_offer"."duration_to_pickup", "driver_offer"."distance_to_pickup", "driver_offer"."valid_till", "driver_offer"."bpp_quote_id", "driver_offer"."rating", "special_zone_quote"."id", "special_zone_quote"."quote_id"
FROM "quote" LEFT OUTER JOIN "trip_terms" ON "quote"."trip_terms_id" = "trip_terms"."id" LEFT OUTER JOIN "rental_slab" ON "quote"."rental_slab_id" = "rental_slab"."id" LEFT OUTER JOIN "driver_offer" ON "quote"."driver_offer_id" = "driver_offer"."id" LEFT OUTER JOIN "special_zone_quote" ON "quote"."special_zone_quote_id" = "special_zone_quote"."id"
WHERE "driver_offer"."estimate_id" = '6d3282f6-b076-4a09-9925-6cb9bdba9b76'
```
Separate queries as `quote` table is sequentially scanned

```
                                                                                    QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Gather  (cost=1012.88..92373.57 rows=2 width=1921) (actual time=373.402..376.108 rows=2 loops=1)
   Workers Planned: 2
   Workers Launched: 2
   ->  Nested Loop Left Join  (cost=12.88..91373.37 rows=1 width=1921) (actual time=364.849..365.249 rows=1 loops=3)
[...]

PR #459, opened 2023-04-07
[WIP] backend/fix : #417 remove join queries

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

We are using join queries where we need data from multiple tables. This works well when we need list of rows from the tables, however we had noticed some queries don't use index on one of the tables in join which leads to higher latencies.

We fixed this by splitting the join queries with point select queries for some of the hot spot queries in earlier issues:
https://github.com/nammayatri/nammayatri/issues/189
https://github.com/nammayatri/nammayatri/issues/336

This solution can be applied to all the find one queries where only one row is required.

Tasks to be done
- [ ] Replace join with select queries in rider-app
- [ ] Replace join with select queries in dynamic-offer-driver-app

### Additional Changes

## Checklist


roman-bodavskiy:
wrong branch name

Issue #438, opened 2023-04-05
Remove join queries used for fetching single row of data from multiple tables


NikithShetty:
We are using join queries where we need data from multiple tables. This works well when we need list of rows from the tables, however we had noticed some queries don't use index on one of the tables in join which leads to higher latencies.

We fixed this by splitting the join queries with point select queries for some of the hot spot queries in earlier issues:
https://github.com/nammayatri/nammayatri/issues/189
https://github.com/nammayatri/nammayatri/issues/336

This solution can be applied to all the find one queries where only one row is required.

Tasks to be done
- [ ] Replace join with select queries in rider-app
- [ ] Replace join with select queries in dynamic-offer-driver-app

PR #337, opened 2023-03-28
backend/#fix: #336 Separating Select Queries

vikrantkm1997:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Separated Select Queries

### Additional Changes

- [ ] This PR modifies the database schema (database migration added)
- [ ] This PR modifies dhall configs/environment variables
- [ ] This PR contains API breaking changes

## How did you test it?
Tested Locally

## Checklist

- [x] I formatted the code and addressed linter errors `./dev/format-all-files.sh`
- [x] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable

PR #460, opened 2023-04-07
backend/refactor: #438 remove join queries

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

To be merged with https://github.com/nammayatri/shared-kernel/pull/75

We are using join queries where we need data from multiple tables. This works well when we need list of rows from the tables, however we had noticed some queries don't use index on one of the tables in join which leads to higher latencies.

We fixed this by splitting the join queries with point select queries for some of the hot spot queries in earlier issues:
https://github.com/nammayatri/nammayatri/issues/189
https://github.com/nammayatri/nammayatri/issues/336

This solution can be applied to all the find one queries where only one row is required.

Tasks to be done
- [x] Replace join with select queries in rider-app
- [x] Replace join with select queries in dynamic-offer-driver-app

Solution:
I use `MaybeT` transformer in each point query. So calculation will return `Nothing` if some point query will return `Nothing`.

### Additional Changes

## How did you test it?

Tested locally

## Checklist


DonDokidon (changes requested):
@NikithShetty @hkmangla  Could you guys explain me again, why we need this? In task description you mentioned, that the issue is that we join on fields. which do not have an indexes, why can't we just add them then? And how separate select, which will use the same unindexed field will help anything?
Did you maybe use Quote table for your tests, which has [this](https://github.com/nammayatri/nammayatri/pull/337/files/41d61ed9d89152e1885503f2364520a0f6537051#r1160099225) issue? If so, don't extrapolate results on all joins, that's very specific issue for this domain type ("either this or that type of entity" approach)

NikithShetty:
> @NikithShetty @hkmangla Could you guys explain me again, why we need this? In task description you mentioned, that the issue is that we join on fields. which do not have an indexes, why can't we just add them then? And how separate select, which will use the same unindexed field will help anything? Did you maybe use Quote table for your tests, which has [this](https://github.com/nammayatri/nammayatri/pull/337/files/41d61ed9d89152e1885503f2364520a0f6537051#r1160099225) issue? If so, don't extrapolate results on all joins, that's very specific issue for this domain type ("either this or that type of entity" approach)

@DonDokidon The query mentioned in the earlier issue has index on all the relevant fields. The join query constructed was not using it for some reason, specially doing a sequential scan on `quote` table which is big in comparison to other tables in the join statement. Also we are having to fix these issues query by query.

This along with the long term plan to add KV layer (which is basically generic cache infront of DB, which can power transactions directly from cache and sync to DB slowly), requires queries to be point queries (join queries are difficult to model in this architecture), we are choosing to do away from join queries in general.

cc: @vimalkumar @Magizhan

vimalkumar:
@DonDokidon 

Using in-memory KV provides us scaling, reliability & perf/cost benefits. Here is why: 

1. **Scaling**: Most of the KV stores (redis cluster in our case), provide out-of-the-box sharding, which allows for better scaling.
2. **Reliability**: Key-value stores like Redis have a simpler design than RDBMS, making it easier to reason about their reliability guarantees. It encourages upfront design of sharding/partition key spaces, which can help transition to a multi-cloud active-active architecture in the future.
3. **Performance/cost**: The simplicity of key-value stores prompts developers to focus on data locality and denormalization, leading to better performance characteristics.
4. **Performance/cost**: In-memory + change log approach: Combining Redis with a streaming platform like Kafka enables systems to operate on soft-state, retaining only recent data in memory. This is particularly useful for creating and manipulating recent session and trip data. In-memory data storage is fast and cost-effective.
5. **Cost**: Since soft-state data has a smaller size and shorter time-to-live (TTL) for durability, it facilitates auto-scaling of the Redis cluster, reducing costs.

_Note: We need a 10x reduction in cloud costs (to ~0.5 cents per trip!), to make this platform economical. Hence some strategic approaches for cost reduction are needed & move to in-memory KV could be an important one._

flake.nix
```diff
@@ -5,8 +5,7 @@
     flake-parts.follows = "common/flake-parts";
     systems.url = "github:nix-systems/default";
 
-    # Backend inputs
-    shared-kernel.url = "github:nammayatri/shared-kernel";
+    shared-kernel.url = "github:nammayatri/shared-kernel/e355d4aae4a46283bf0a01d4c6f38bd01153b3c2";
```
srid, line 8: shared-kernel should be updated without changing `flake.nix`; see https://github.com/nammayatri/nammayatri/tree/main/Backend#updating-flake-inputs

PR #491, opened 2023-04-10
frontend/fix/#490/handling-youtube video orientation in full screen mode

MercyQueen:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
- Added one for variable to store the YouTube video type ( SHORTS / VIDEO )
- Added a function that would check if the word "shorts" is present in the YouTube url that we get from backend
( as the links which are shorts will have the keyword shorts in it.)

### DIFF
--- a/Frontend/ui-driver/src/Helpers/Utils.js
+++ b/Frontend/ui-driver/src/Helpers/Utils.js
@@ -35,21 +35,21 @@ exports["getTimer"] = function (valType) {
       return function (action) {
           return function(){
               var callback = callbackMapper.map(function () {
-                    
+
                 console.log(valType+"<===>"+startTime+"<===>"+inputVal)
-                
+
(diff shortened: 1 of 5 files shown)


Frontend/ui-driver/src/Helpers/Utils.js
```diff
@@ -2390,4 +2390,17 @@ exports["getImageUrl"] = function (url) {
   }catch (e) {
     console.log("error in getImageUrl " + e);
   }
-};
\ No newline at end of file
+};
+
+exports["getVideoType"] = function (url) {
```
Vignesh-772, line 2395: can't we handle this in purescript itself

witcher-shailesh:
Need to change the approach, closing the PR for now!

PR #646, opened 2023-04-21
Include `nix` package manager in docker image

srid:
image


srid:
Note: this increases the docker image size from 0.7 to 1 GB.

PR #594, opened 2023-04-18
Frontend/chore/purescript-upgrade

belevy:
Update code and dependencies to compile with purescript v0.15
Additionally this PR adds a .#frontend development shell with nodejs, spago and purescript.

Currently packages.dhall is pointing at forks of the dependencies where needed. The open PR's to get the dependencies upstreamed are 
https://github.com/juspay/purescript-backtrack/pull/6
https://github.com/juspay/purescript-halogen-vdom/pull/9
https://github.com/juspay/purescript-presto/pull/47
https://github.com/juspay/purescript-presto-dom/pull/3
https://github.com/juspay/purescript-tracker/pull/5
https://github.com/juspay/purescript-otp-reader/pull/7

PR #685, opened 2023-04-25
Break compile time dependency between ui-common and apps

belevy:
Break the compile time dependencies between ui-common and ui-driver/ui-customer to support seperate compilation

* Extract common colors used by common components instead of relying on individual apps to define common colors
* Stop using Storage module to lookup LANGUAGE_KEY in the shared preferences store
* Move FareBreakup component into ui-customer app
* Move common animation and styles types into ui-common package
* Remove unnecessary dependency on `GlobalState` type and replace with a free type variable


belevy:
@srid I went ahead and broke out the trailing-ws change into its own commit as best as I could. Not sure if that is good enough.

srid:
A separate commit is good from commit history standpoint. For the PR review standpoint, people can use "Hide whitespace" option:

image

witcher-shailesh (approved):

srid (commented):
Since #594 is merged, this requires a rebase on top of `main`.

flake.nix
```diff
@@ -10,6 +10,8 @@
     shared-kernel.inputs.nixpkgs.follows = "nixpkgs";
     beckn-gateway.url = "github:nammayatri/beckn-gateway";
     beckn-gateway.inputs.shared-kernel.follows = "shared-kernel";
+    easy-purescript-nix.url = "github:justinwoo/easy-purescript-nix";
+    easy-purescript-nix.flake = false;
```
srid, line 14: Group these out with a `# Frontend inputs` comment (just like the backend inputs).

srid (approved):
(I haven't reviewed the PureScript changes)

PR #840, opened 2023-05-09
Frontend/chore/purifix

belevy:


Frontend/default.nix
```diff
@@ -1,28 +1,213 @@
...
+          export CONFIG_FILE
+          CONFIG_FILE=$(realpath "$TARGET/webpack.${platform}.js")
+
+          export PIPE=/tmp/${name}
+          mkfifo $PIPE
+          trap 'rm -f $PIPE' EXIT
+
+          ${lib.getExe config.proc.groups.frontend-dev.package}
+        '';
+      };
```
srid, line 102: I don't think we need this `start-app-devserver` indirection. It is used in two places, and can be inlined in proc-flake definition. Also `$PIPE` should be stored under project directory (eg: `./Frontend/tmp/${name}`). And proc-flake already cd's to FLAKE_ROOT.

This definition is much more straightforward to read as a new person in the team:

```nix
      proc.groups =
        let
          mkWatchProcesses = { target, platform, env }: {
            webpack-watch.command = ''
              ${lib.getExe webpack-dev-server} \
                ./tmp/pipes/${target} \
                --env ${env} \
                --config $(realpath "./Frontend/${target}/webpack.${platform}.js")
            '';
            purifix-watch.command = ''
              cd ./Frontend/${target}
              ${lib.getExe purifix-watch} ./tmp/pipes/${target} . ../ui-common
            '';
          };
        in
        {
          ui-customer-start.processes = mkWatchProcesses {
            target = "ui-customer";
            platform = "android";
            env = "master";
          };
          ui-driver-start.processes = mkWatchProcesses {
            target = "ui-driver";
            platform = "android";
            env = "master";
          };
        };
```

srid: Also, let's move all the watch related stuff to its own file `watch.nix` or something.

srid: Question: what happens if the user tries to start the watch twice in different terminals at the same time? Can the pipe be reused as a lock to prevent that?

Frontend/node/package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "devDependencies": {
+    "@babel/core": "^7.7.7",
+    "@babel/preset-env": "^7.7.7",
+    "babel-loader": "^8.0.6",
+    "filemanager-webpack-plugin": "^2.0.5",
+    "webpack": "^4.41.5",
+    "webpack-cli": "^3.3.10",
+    "webpack-dev-server": "^3.10.1",
+    "webpack-merge": "^5.8.0"
```
srid, line 21: Are these dependencies used only in `watch.js`? I'm wondering if we can make watch-dev-server a self-contained sub-directory along with the Nix function.

Frontend/node/watch.js
srid, line 1: Can you add comment at the top of this file explaining the purpose it serves?

Also a question: why must all of this code live in this mono repo?

srid (commented):
I've reviewed the Nix changes. Except for the `make-bundle` stuff ... could you update the README stating how it gets used and tested?

Frontend/default.nix
```diff
@@ -1,28 +1,213 @@
...
+            translator = "package-lock";
+            builder = "strict-builder";
+            subsystemInfo = {
+              nodejs = 14;
+            };
+          };
+        };
+      };
+
+      nodeDependencies = nodePackages.packages.${system}.atlas-ui.lib;
```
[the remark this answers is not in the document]
srid: (The goal is to decompose `Frontend/default.nix` such that that file itself is not so huge)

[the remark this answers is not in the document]
belevy: It was my understanding that a flake module is best used for adding outputs. This is really more of a function from `{dream2nix, system} -> nodeModules` would it not be better to just have something like
```
nodeDependencies = import ./node/nodeDependencies.nix { 
                                        inherit (inputs) dream2nix; 
                                        inherit system; 
                                 };
```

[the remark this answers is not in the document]
srid: `nodeDependencies` is a derivation. We can add outputs to the flake even if they are not directly used by the developer (the backend flake has quite a few of them). The CI enumerates all packages/checks/apps/devShells and builds them separately.

In this case what are proposing is probably okay as well.

Frontend/nix/bundle.nix
```diff
@@ -0,0 +1,55 @@
+{ ... }:
+{
+  imports = [
+    ../node
+  ];
```
srid, line 5: You don't need this because it is already imported at top-level.

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,109 @@
+{ ... }:
+{
+  imports = [
+    ../node
+  ];
```
srid, line 5: You don't need this because it is already imported at top-level.

Frontend/node/default.nix
```diff
@@ -0,0 +1,50 @@
...
+  options = {
+    perSystem = mkPerSystemOption
+      ({ config, self', lib, pkgs, system, ... }:
+        let
+          nodePackages = inputs.dream2nix.lib.makeFlakeOutputs {
+            systems = [ system ];
+            source = ./.;
+            projects = {
+              "atlas-ui" = {
+                name = "atlas-ui";
```
srid, line 18: Why `atlas-ui` (as naming)?

Frontend/node/default.nix
```diff
@@ -0,0 +1,50 @@
...
+                nodejs = lib.mkOption {
+                  type = types.package;
+                };
+              };
+            };
+          };
+          config.nammayatri = {
+            inherit nodeDependencies;
+            nodejs = pkgs.nodejs-14_x;
+          };
```
srid, line 47: Just put this in `config.packages.nodeDependencies` (or named suitably) and let it build as part of CI. We don't need a submodule here.

srid: And instead of `nodejs`, expose a devShell containing the node binary and [compose](https://haskell.flake.page/devshell#composing-devshells) it at higher level.

Frontend/nix/bundle.nix
```diff
@@ -0,0 +1,55 @@
...
+
+  perSystem = { config, self', lib, pkgs, system, ... }:
+    let
+      make-bundle = { mode ? "production", env ? "prod", target, platform }:
```
srid, line 9: These defaults are not used? Let's remove them.

Frontend/nix/bundle.nix
```diff
@@ -0,0 +1,55 @@
...
+          target = [ "ui-customer" "ui-driver" ];
+          platform = [ "android" ]; # TODO: Re-add ios
+        };
+
+    in
+    {
+      packages =
+        builtins.listToAttrs (map
+          (args: {
+            name = "${args.target}-${args.platform}-${args.env}-${args.mode}-js";
```
srid, line 50: You can access the `name` property from the package (returned by `make-bundle`) itself. No need to replicate it here.

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,109 @@
...
+            };
+          };
+        };
+
+      process-options =
+        lib.attrsets.cartesianProductOfSets {
+          env = [ "master" "sandbox" "prod" ];
+          target = [ "ui-customer" "ui-driver" ];
+          platform = [ "android" ]; # TODO: support iOS
+        };
```
srid, line 81: Looks like you want to factor out the "build config" separately, `./nix/build-config.nix` or something. The calculation of "name" happens in several places, and can be normalized here as well.

flake.nix
```diff
@@ -8,8 +8,12 @@
...
     beckn-gateway.inputs.shared-kernel.follows = "shared-kernel";
 
-    easy-purescript-nix.url = "github:justinwoo/easy-purescript-nix";
-    easy-purescript-nix.flake = false;
+    # Frontend inputs
+    purifix.url = "github:purifix/purifix";
+    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.inputs.nixpkgs.follows = "dream2nix-nixpkgs-fixed";
+    dream2nix-nixpkgs-fixed.url = "nixpkgs/c1989c17e2658f721df7b0027cbc3d8959f914cb"; # when this fix gets upstreamed remove
+    proc-flake.url = "github:srid/proc-flake";
```
srid, line 16: `proc-flake` no longer used?

flake.nix
```diff
@@ -8,8 +8,12 @@
     beckn-gateway.inputs.common.follows = "common";
     beckn-gateway.inputs.shared-kernel.follows = "shared-kernel";
 
-    easy-purescript-nix.url = "github:justinwoo/easy-purescript-nix";
-    easy-purescript-nix.flake = false;
+    # Frontend inputs
+    purifix.url = "github:purifix/purifix";
+    dream2nix.url = "github:nix-community/dream2nix";
+    dream2nix.inputs.nixpkgs.follows = "dream2nix-nixpkgs-fixed";
+    dream2nix-nixpkgs-fixed.url = "nixpkgs/c1989c17e2658f721df7b0027cbc3d8959f914cb"; # when this fix gets upstreamed remove
```
srid, line 15: please add a link to the github issue here, so as to remind us when to remove it in future.

srid (changes requested):
Just reviewed the Nix code in this PR. Please also see TODO in PR description, in particular VSCode IDE support.

flake.nix
srid: You can just link to https://github.com/nix-community/dream2nix/issues/522

Frontend/node/default.nix
belevy: that was the package name in ui-common from before. It can be whatever, happy to change it to common.

Frontend/nix/bundle.nix
belevy: I like the fact that it documents at the use site where the nodeDependencies/nodejs version are coming from. I can remove it but was fully aware that it was redundant.

Frontend/node/default.nix
belevy: how does this proposed devShell solution work for the bundle derivation? I can probably change the watch script to rely on being run in the devShell and add the node_path that way but I am unclear how to utilize this in a non-shell derivation.

srid: The bundle drv wouldn't use the devShell. Instead of `config.nammayatri.nodejs` it would use `pkgs.nodejs-14_x` (or `self'.packages.nodejs`).

Frontend/node/default.nix
srid: I'm trying to understand the organization, but I don't think I get it yet. 

Is `Frontend/node` more or less independent from the rest of the project? Also there is a `watch.js` in this directory, yet `webpack-dev-server` is define *outside*. You want to keep the Nix alongside the code they are packaging generally.

belevy: `webpack-dev-server` is specific to watch hence it living in `watch.nix`. Perhaps you would prefer `Frontend/nix/watch/default.nix` and `Frontend/nix/watch/watch.js`? Ultimately, I couldn't a good way to split the watch/bundle/node apart in a way that makes sense from every angle and so this is what I ended up with.

srid: I'd just put `watch.js` alongside `watch.nix` for now.

Still trying to wrap my head around the whole watch nix organization logic ...

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,109 @@
...
+          process-options);
+
+      mission-control-scripts =
+        builtins.listToAttrs (map
+          (args: rec {
+            name = process-name args;
+            value = {
+              category = "Frontend Watch";
+              description = "Start the dev server for ${args.target} on ${args.platform} with env set to ${args.env}";
+              exec = "${lib.getExe self'.packages.${name}} -t=false";
```
srid, line 100: Don't need `-t=false` here. Once #1254 is merged, you can set `tui = false;` in process-compose configuration.

Frontend/frontend.code-workspace
srid, line 1: Let's add recommended VSCode extensions here:

```diff
diff --git a/Frontend/frontend.code-workspace b/Frontend/frontend.code-workspace
index f87e6302a..b411ec606 100644
--- a/Frontend/frontend.code-workspace
+++ b/Frontend/frontend.code-workspace
@@ -14,5 +14,11 @@
                "purescript.addSpagoSources": false,
                "purescript.buildCommand": "purifix --json-errors \"./src/**/*.purs\" \"../ui-common/**/*.purs\"",
                "purescript.formatter": "none"
+       },
+       "extensions": {
+               "recommendations": [
+                       "nwolverson.ide-purescript",
+                       "bbenoist.nix"
+               ]
        }
 }
```

Frontend/default.nix
```diff
@@ -1,28 +1,69 @@
...
+      };
+
+    in
+    {
       treefmt.config = {
         # Suppress autoformatting of frontend dhall files.
         settings.formatter.dhall.excludes = [
           "Frontend/packages.dhall"
         ];
       };
```
srid, line 47: This can be removed; I don't see a `packages.dhall`.

Frontend/default.nix
```diff
@@ -1,28 +1,69 @@
...
+        nativeBuildInputs = [ pkgs.makeWrapper ];
+        paths = [
+          pkgs.nodePackages.typescript
+          pkgs.nodePackages.typescript-language-server
+        ];
+        postBuild = ''
+          wrapProgram $out/bin/typescript-language-server \
+            --add-flags "--tsserver-path $out/bin/tsserver"
+        '';
+      };
```
srid, line 38: Move this to `./nix/tsserver.nix`, so that we can keep `Frontend/default.nix` as simpler/smaller as possible.

Frontend/frontend.code-workspace
srid: Also, let's add the repo root itself as a workspace (otherwise we will not be able to edit the flake.nix/default.nix files).

```diff
--- a/Frontend/frontend.code-workspace
+++ b/Frontend/frontend.code-workspace
@@ -8,11 +8,20 @@
                },
                {
                        "path": "ui-common"
+               },
+               {
+                       "path": ".."
                }
        ],
        "settings": {
```

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,115 @@
...
+          process-options);
+
+      mission-control-scripts =
+        builtins.listToAttrs (map
+          (args: rec {
+            name = process-name args;
+            value = {
+              category = "Frontend Watch";
+              description = "Start the dev server for ${args.target} on ${args.platform} with env set to ${args.env}";
+              exec = "${lib.getExe self'.packages.${name}}";
```
srid, line 106:
```suggestion
              exec = lib.getExe config.process-compose.${name}.outputs.package;
```

It is confusing to refer to some arbitrary package in the flake output, and this is more clear as to what kind of package is being used.

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,115 @@
...
+        lib.attrsets.cartesianProductOfSets {
+          env = [ "master" "sandbox" "prod" ];
+          target = [ "ui-customer" "ui-driver" ];
+          platform = [ "android" ]; # TODO: support iOS
+        };
+
+      process-name = { target, platform, env }: "${target}-start-${platform}-${env}";
+
+      process-configs =
+        builtins.listToAttrs (map
```
srid, line 92: There are quite a few places -- here and in `bundle.nix` -- you are using `builtins.listToAttrs . map`. This is unnecessary when your build configurations info is essentially an attrset. What you want is something like:

```nix
let
   build-configs =
      let choices = lib.attrsets.cartesianProductOfSets {
          env = [ "master" "sandbox" "prod" ];
          target = [ "ui-customer" "ui-driver" ];
          platform = [ "android" ]; # TODO: support iOS
        };
     in builtins.listToAttrs (map (args: { name = "${args.target}-start-${args.platform}-${args.env}"; value = args; }) choices);
```

Then you can just `lib.mapAttrs _ build-configs` which is far more straightforward to understand on first glance.

srid (changes requested):
I now understand how the watch stuff works and is laid out in code. 

Overall I find the Nix code here to [complect](https://www.infoq.com/presentations/Simple-Made-Easy/) a bit which explains my original gut feeling. Here are some ideas on improving that.

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,115 @@
...
+            for SRC_FOLDER in "$@"; do
+              if [ -d "$SRC_FOLDER/src" ]; then
+                SRC_FOLDERS+=( -w "$SRC_FOLDER/src" )
+                SRC_GLOBS+=( "\"$SRC_FOLDER/src/**/*.purs\"" )
+              fi
+            done
+            watchexec "''${SRC_FOLDERS[@]}" -r -f "**/*.purs" -- \
+              "echo 'suspend' >$PIPE;purifix ''${SRC_GLOBS[*]} && echo 'ready' >$PIPE"
+          '';
+        };
```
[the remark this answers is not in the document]
belevy: buildInputs is to support running outside of the devShell, I have removed it as requested.

I disagree that this is really so decoupled from the project its self. The fact is most people are not going to want to use a pipe to communicate between two different watch processes. This whole two processes intertwined thing is really more of a hack necessitated by the way the existing project was using webpack. In an ideal world we could just use the existing webpack loader but I was unable to get it to work correctly with purifix. 

In general the reason this whole thing feels like its all intertwined is because it is inherently and breaking apart things that are inherently intertwined will not make anything clearer. The webpack-dev-server process and the purifix-watch process are engaged in a very specific dance where they are trading off work, they are necessarily decoupled via a pipe which has a fairly simple protocol but each piece independently makes no sense.

I also don't really believe that purifix will ever directly solve this problem. There are two things going on here that are usually not relevant. 
1. We have a multi-project repo where two of the packages are mutually exclusive due to overlapping namespacing. This is why we even need to do the globbing/src watching instead of just using purifix-project which will build your whole project instead of just acting like the `purs compile` command.
2. We are using webpack in a fairly abnormal way on the project. Webpack watch doesn't seem to debounce the changes coming from purescript well enough on its own so it will constantly recompile unless told to suspend its self. 

The fact is this whole long lived independent watch processes stuff is a hack necessitated by the slowness of webpack combined with the inability to integrate directly in a sensible way.

Frontend/nix/watch.nix
```diff
@@ -0,0 +1,115 @@
...
+      webpack-dev-server = pkgs.writeShellApplication {
+        name = "webpack-dev-server-wrapped";
+        runtimeInputs = [ self'.packages.nodejs ];
+        text = ''
+          export NODE_PATH="${self'.packages.nodeDependencies}/node_modules"
+          PIPE="$1"
+          shift
+          tail -n1 -f "$PIPE" | node ${./watch.js} "$@"
+        '';
+      };
```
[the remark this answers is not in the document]
belevy: Watch.nix is the unit of isolation in my view, see below for the more detailed comment.

The reason for a single monolithic nodeDependencies is because of webpack. The only thing that could actually be removed would be webpack-dev-server/webpack-cli. The way that webpack works means that we need to have a single node_modules folder with both the dev dependencies for running webpack and the runtime dependencies that will be bundled into the package. This is not really something that can be broken apart. Theoretically we could make two different package.json files one for each dev dependencies and runtime dependencies but dream2nix already independently compiles everything so we would not gain anything performance wise.

Frontend/nix/watch.nix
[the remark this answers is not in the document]
srid: The whole `purfix-watch` + (your custom) `webpack-dev-server` combination is [an area of volatility](https://www.informit.com/articles/article.aspx?p=2995357&seqNum=2) that NY frontend devs shouldn't have to concern themselves with when reading top-level Nix. When better solutions arise in future, we should be able to swap it in easily. The interface here really just says "take this directory, and compile it with common directory, watch for it as part of webpack dev server".

> In an ideal world we could just use the existing webpack loader but I was unable to get it to work correctly with purifix.

Yea, but why does it have to be an *existing* loader? I am doing some reading on how on webpack works, and it seems to me that we are not approaching this problem cleanly: we're writing custom hack when we could just use webpack's loader mechanism. Have you actually considered writing a _new_ (or reusing an existing) loader that works with our purifix project? 

> In an ideal world we could just use the existing webpack loader but I was unable to get it to work correctly with purifix.

By "the existing webpack loader" are you referring to https://github.com/ethul/purs-loader? What's the problem exactly? Can we extend it to work with purifix? I imagine this work would be useful beyond nammayatri, and in other PureScript projects anyway.

(As for overlapping namespacing, I thin we should fix it separately if it turns out to be a problem).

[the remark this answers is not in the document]
belevy: >The whole purfix-watch + (your custom) webpack-dev-server combination is [an area of volatility](https://www.informit.com/articles/article.aspx?p=2995357&seqNum=2) that NY frontend devs shouldn't have to concern themselves with when reading top-level Nix.
How is nix/webpack-dev-server/default.nix and nix/purifix-watch.nix any more or less top level than nix/watch.nix?

> Have you actually considered writing a new (or reusing an existing) loader that works with our purifix project?
Yes I already attempted this and determined that the amount of effort to make a new loader would be fairly significant and so disregarded it as a practical option.

> By "the existing webpack loader" are you referring to https://github.com/ethul/purs-loader? What's the problem exactly?
It fails to properly rewrite the imports post compilation. The main issue with webpack and purescript working together is that they have different expectations about their roles. Webpack assumes that it has a compiler that can convert a single file at a time to js. Purescript wants to compile everything all at once to js. To properly handle webpacks view where it wants to collect all the dependencies itsself the purs-loader rewrites the output js files to point back to the purs files. However, in our setup it appears to fail to properly determine where the src is located. In this process it mangles the file and causes the whole process to crash. 

I can try again to get it working but it really does seem like were getting into pareto principle territory over here.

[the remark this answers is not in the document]
srid: `nix/watch.nix` contains two things: the process-compose/mission-compose configuration (which qualifies as top-level Nix that devs are expected to modify) as well as the implementation details of this watch hack (which doesn't need to appear there).

> To properly handle webpacks view where it wants to collect all the dependencies itsself the purs-loader rewrites the output js files to point back to the purs files. However, in our setup it appears to fail to properly determine where the src is located. In this process it mangles the file and causes the whole process to crash.

I'm not sure if I completely understood that. In generally, whenever something like this happens the course of action I'd recommend is:

1. Open an upstream issue detailing the issue (in this case, on https://github.com/ethul/purs-loader)
2. Attempt to fix the issue and open a upstream PR
3. If the fix is difficult or impossible, explain so in that upstream issue, and then link all that information in the local hack (webpack-dev-server/ purifix-watch) so that future engineers modifying this Nix will have a context as to why all of this exists.

> I can try again to get it working but it really does seem like were getting into pareto principle territory over here.

Yea, please do. It is ideal to fix https://github.com/ethul/purs-loader itself - but [timebox](https://en.wikipedia.org/wiki/Timeboxing) the whole thing, as we don't want to be spending too much time there. Do open the upstream issue and link to it if that effort fails.

PR #885, opened 2023-05-15
Fix Kannada Language Translations

bskrksyp9:
Improvised with appropriate words to make it more meaningful

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates


witcher-shailesh (changes requested):
please follow contributing guidelines

bskrksyp9:
But this is just a minor change in a file doesn't involve any coding, could you please point me what needs to be followed here?

bskrksyp9:
It would be helpful if you could point me what needs to be followed here.

As per my knowledge, i may have to change branch naming convention? if yes, 

what needs to be mentioned for issues in number as?

and none of the template matches with the changes that i did hence I had to remove those template lines to avoid lengthy description.

PR #1096, opened 2023-06-02
BugFixes-in-RateCard

Akash2172:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates


Frontend/ui-customer/src/Components/RateCard/View.purs
```diff
@@ -379,7 +379,7 @@ driverAdditionView push config =
     , imageView
       [ height $ V 100
       , width MATCH_PARENT
-      , imageWithFallback "ny_ic_driver_addition_table,https://assets.juspay.in/beckn/nammayatri/user/images/ny_ic_driver_addition_table.png"
+      , imageWithFallback "ny_ic_driver_addition_table2,https://assets.juspay.in/beckn/nammayatri/user/images/ny_ic_driver_addition_table2.png"
```
Rohit4997, line 382: can you remove previous `ny_ic_driver_addition_table`  image if it is not used
Also just replace the image no need to change image name

witcher-shailesh: For now remove the older icon, as we want the image to S3 as well name is fine for now

witcher-shailesh (changes requested):
Added comments

Frontend/ui-customer/src/Components/RateCard/View.purs
kavya-shree-s: @witcher-shailesh image pushed to S3 can we overridden with latest image with same name

SrinivasNaikB: Answering on behalf of @Akash2172 
- ny_ic_driver_addition_table,png depicts that driver optional addition of upto 10 is only allowed for trips if distance 5km to 10 km (this was the previously planned Fare Policy and we might anytime go back to that if need be)
- ny_ic_driver_addition_table2,png- depicts a similar information as the earlier version of the image, but it also says drivers are allowed to add upto Rs.10 for less than 5km trips also.

We need to keep both in S3 as there is plan to shift between these two fair policies soon in future. Keeping both can allow us to achieve the shift smoothly with a simple hotpush when needed.

witcher-shailesh (approved):

Issue #1260, opened 2023-06-15
make email as case insensitive for dashboard login


saurav2401:
**Is your feature request related to a problem? Please describe.**
make email as case insensitive for dashboard login

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1261, opened 2023-06-15
backend/feature/: #1260/make email as case insensitive for dashboard …

saurav2401:
…login

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
make email as case insensitive for dashboard login

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
+++ b/Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
@@ -16,6 +16,7 @@
 module Storage.Queries.Person where
 
 import Control.Applicative (liftA2)
+import qualified Data.Text as T
 import Database.Esqueleto.PostgreSQL
 import Domain.Types.Merchant as Merchant
 import Domain.Types.Person as Person
@@ -44,7 +45,7 @@ findByEmail ::
   Text ->
   m (Maybe Person)
 findByEmail email = do
-  emailDbHash <- getDbHash email
+  emailDbHash <- getDbHash (T.toLower email)
   findOne $ do
     person <- from $ table @PersonT
     where_ $
@@ -57,7 +58,7 @@ findByEmailAndPassword ::
   Text ->
   m (Maybe Person)
 findByEmailAndPassword email password = do
-  emailDbHash <- getDbHash email
+  emailDbHash <- getDbHash (T.toLower email)
   passwordDbHash <- getDbHash password
   findOne $ do
     person <- from $ table @PersonT
--- a/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
+++ b/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
@@ -15,6 +15,7 @@
 module Domain.Action.Dashboard.Person where
 
 import Dashboard.Common
+import qualified Data.Text as T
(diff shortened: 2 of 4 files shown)


shanm-16 (approved):

Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
```diff
@@ -58,11 +59,12 @@ findByEmailAndPassword ::
   m (Maybe Person)
 findByEmailAndPassword email password = do
   emailDbHash <- getDbHash email
+  emailDbHashLower <- getDbHash (T.toLower email)
   passwordDbHash <- getDbHash password
   findOne $ do
     person <- from $ table @PersonT
     where_ $
-      person ^. PersonEmailHash ==. val emailDbHash
+      (person ^. PersonEmailHash ==. val emailDbHash ||. person ^. PersonEmailHash ==. val emailDbHashLower)
```
0utkarsh, line 67: You can make table's hash lower too and compare. Also, I doubt if hash is case insensitive and if it isn't, you should lowercase the email before fetching it and dumping to table too

DonDokidon: > You can make table's hash lower too and compare instead of OR condition. Also, I doubt that hash is case insensitive and if it isn't, you should lowercase the email before fetching and dumping to table too

How? It is hash, not the actual Email, you can't get hash of lower cased email out of whatever cased one.
It would be better to have all emails in lower case in db, but idk how to migrate it if it is not already the case...

0utkarsh: Yes, I'm saying that when a person is created, lower its email - take hash - and then dump to db.
Then while comparing - lower the email - take hash and compare without lowering anything else.

saurav2401: @0utkarsh ,I also thought the same , but max users are already created what to do with that ?

0utkarsh: I think the number of users should be very less rn, so we can create an API to update email which only admin will have access to. Admin can manually update such person's email.

Also, since external people are going to use dashboard, we should do this soon. If we are not doing this soon, at least make sure future users don't face this issue by making the necessary change in `createPerson` API at least.
cc: @KashishKhandelwal @hkmangla

hkmangla: Make sense. @saurav2401 Lets do one thing, in dashboard create person API, make email id always lower before creating entry in DB. Here just make this `emailDbHash <- getDbHash email` to this `emailDbHash <- getDbHash (T.toLower email)). Don't change anwhere else. In DB,  I will make already created email lower case manually.

hkmangla (changes requested):

Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
saurav2401: ok @hkmangla

hkmangla (dismissed):

hkmangla (approved):

PR #1344, opened 2023-06-21
Backend/#1312/ Split end ride thread

### LINKED ISSUE #1312: Split End Ride API into two threads
Currently, the `endride` API on BPP side returns success after completing the fare computation and updating the status of the Ride object with fare details. We need to split this behaviour into two, where a new endpoint called `endRide2` will be created that returns success immediately and the fare computation is triggered in a new thread parallelly. A new `rideStatus` API will be polled with `rideID` in the URL that returns the details for `one` ride, unlike rideList that returns the same for a list of rides.
The `endRide2` API's request body will have a new flag called `routeDeviated` :: Bool.

Additionally, the calls to `snapToRoad` API (which are used to calculate ride.TravelledDistance, in batches of 99 points) have to be moved to this new thread after endRide2, and made only if the above flag is true.

### **Psuedo Code:**

endRide2 is called:
  return success
  if routeDeviated
      Trigger snaptoRoad, recalculate fare and update Ride details
  else
       Update Ride details with estimate values
  
In response to the new `rideStatus` API, send the ride object details.

DonDokidon:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

fix #1312 
## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
@@ -42,7 +42,6 @@ import Environment (Flow)
 import EulerHS.Prelude hiding (id, pi)
 import Kernel.External.Maps
 import Kernel.Prelude (roundToIntegral)
-import qualified Kernel.Types.APISuccess as APISuccess
 import Kernel.Types.Common
 import Kernel.Types.Id
 import Kernel.Utils.CalculateDistance (distanceBetweenInMeters)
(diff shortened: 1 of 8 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -124,7 +124,7 @@ driverEndRide ::
   ServiceHandle m ->
   Id DRide.Ride ->
   DriverEndRideReq ->
-  m APISuccess.APISuccess
+  m ()
 driverEndRide handle rideId req =
```
umangj-28, line 128: why are we removing APISuccess?

DonDokidon: > why are we removing APISuccess?

Why not? It's not up to domain logic to return this anyway

0utkarsh (approved):

PR #1453, opened 2023-06-30
backend/refactor/Remove join from driverList

0utkarsh:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Person.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Person.hs
@@ -74,7 +74,7 @@ mkFullDriver :: (Person, DriverLocation, DriverInformation, Vehicle) -> FullDriv
 mkFullDriver (p, l, i, v) = FullDriver p l i v
 
 findAllDriversWithInfoAndVehicle ::
-  Transactionable m =>
+  (Transactionable m, EsqDBReplicaFlow m r) =>
   Id Merchant ->
   Int ->
   Int ->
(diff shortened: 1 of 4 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Person.hs
```diff
@@ -86,30 +86,53 @@ findAllDriversWithInfoAndVehicle ::
...
+
+mapDriverWithVehicle :: HashMap.HashMap Text Person -> HashMap.HashMap Text Vehicle -> DriverInformation -> Maybe (Person, DriverInformation, Maybe Vehicle)
+mapDriverWithVehicle driverHM vehicleHM info = do
+  let driverId = info.driverId.getId
+  person <- HashMap.lookup driverId driverHM
+  let vehicle = HashMap.lookup driverId vehicleHM
+  Just (person, info, vehicle)
+
+getDriversByAttributes :: Transactionable m => Id Merchant -> Maybe DbHash -> m [Person]
+getDriversByAttributes merchantId mbSearchPhoneDBHash = do
```
piyushKumar-1, line 114: can we add `limit`, `offset` here ?

0utkarsh: No, if driver is searched by vehicle number, he could be missed if we add limit and offset here

0utkarsh: Yes, I also thought of doing that but we'll have a lot of checks in that case plus there will be concept of priority too. Let's say 
1. Only `vehicleNumber` is provided - need to make `vehicle` table as the initial set
2. Only `phoneNumber` is provided - `person` table as initial set
3. Both given -> Any order is fine
4. Both not given and `mbSubscribed` is given -> `driverInformation` as initial set

Similarly there can be other cases to check.
For now, my assumption was that this query is mostly used to search by `phoneNumber` and hence I didn't want to complicate or mess up the coding to optimize it for cases which doesn't occur much. This can be changed though, if we have high number of queries with `vehicleNumber`

PR #2761, opened 2023-08-26
FCM/Device-Token-issue

KiranBKL:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/Frontend/ui-driver/src/Flow.purs
+++ b/Frontend/ui-driver/src/Flow.purs
@@ -91,8 +91,13 @@ import Types.ModifyScreenState (modifyScreenState, updateStage)
 import Types.ModifyScreenState (modifyScreenState, updateStage)
 import Common.Styles.Colors as Color
 import Effect.Uncurried (runEffectFn1)
-import Effect.Aff (makeAff, nonCanceler)
 import Engineering.Helpers.LogEvent (logEvent)
+import Engineering.Helpers.Commons (flowRunner)
+import JBridge as JB
+import Effect.Aff (makeAff, nonCanceler, launchAff)
+import Control.Monad.Except (runExceptT)
+import Control.Transformers.Back.Trans (runBackT)
+
 
 baseAppFlow :: Boolean -> FlowBT String Unit
 baseAppFlow baseFlow = do
@@ -417,7 +422,6 @@ updateDriverVersion dbClientVersion dbBundleVersion = do
       pure unit
     else pure unit
   else pure unit
-
 
 aadhaarVerificationFlow :: FlowBT String Unit
 aadhaarVerificationFlow = do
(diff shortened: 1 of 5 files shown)


Frontend/ui-driver/src/Flow.purs
```diff
@@ -407,7 +414,21 @@ updateDriverVersion dbClientVersion dbBundleVersion = do
...
     else pure unit
   else pure unit
-
+data FCMToken = FCMToken String
+
+checkAndUpdateToken :: Int -> FlowBT String Unit
+checkAndUpdateToken count = do
+    if count > 0 then do
+        let token = getValueToLocalStore FCM_TOKEN
+        when (any (_ == token) ["_failed","null","(null)",""]) do
```
witcher-shailesh, line 423: Where is masked token check? If it's coming as (["_failed","null","(null)",""])  while fetching from localStorage it's not necessary that it is wrong in DB as well. That's why we need to add this check in maskedDeviceToken which we receive when in profile response. Please refactor this as discussed

Frontend/ui-driver/src/Flow.purs
```diff
@@ -407,7 +414,21 @@ updateDriverVersion dbClientVersion dbBundleVersion = do
...
+            (FCMToken newToken) <- lift $ lift $ doAff $ makeAff \cb -> setFCMTokenWithTimeOut 5000 (cb <<< Right) FCMToken $> nonCanceler
+            if newToken == "NOT_FOUND" then checkAndUpdateToken (count - 1)
```
witcher-shailesh, line 425: If it's a recursive function there should be a delay of calling it again

witcher-shailesh (changes requested):

Frontend/ui-driver/src/Flow.purs
KiranBKL: But In wrong cases it will update with token in db and right cases will we overrided

KiranBKL: It will update the DB with token in wrong cases and if in db it is not correct then also it will be overrided with correct token

witcher-shailesh: No

witcher-shailesh: What does Profile response API gives you?

KiranBKL: For user side there is maskedTokenDevice in profileResponse API but not in driver profile Repsone API

KiranBKL: in DriverSide it is only available in VerifyTokenResp

Vignesh-772 (approved):

witcher-shailesh (approved):

PR #2810, opened 2023-08-30
backend/feat/#2613/Driver Go To Home: "A Homebound Ride Assignment fe…

ratnadeep-99:
…ature which when enabled allows drivers to be assigned rides heading towards their homes, enhancing driver convenience and satisfaction."

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
```diff
@@ -0,0 +1,143 @@
...
+  let ghkey = makeGoHomeReqKey driverId
+  currTime <- getLocalCurrentTime =<< ((CQTC.findByMerchantId merchantId >>= fromMaybeM (InternalError "Transporter config for timezone not found")) <&> (.timeDiffFromUtc))
+  Hedis.safeGet ghkey >>= \case
+    Just ghrData -> do
+      if maybe False (\validTill -> validTill > currTime || ghrData.isOnRide) ghrData.validTill
+        then return ghrData
+        else checkInvalidReqData ghrData currTime ghkey driverId merchantId ghCfg expTime
+    Nothing -> do
+      Hedis.setExp ghkey (templateGoHomeData Nothing initCnt Nothing Nothing False Nothing currTime) expTime
+      return $ templateGoHomeData Nothing initCnt Nothing Nothing False Nothing currTime
```
piyushKumar-1, line 46: we are initiating goHomeReqId as `Nothing`, and i can see at a lot of places you have validation on goHomeReqId not there then throw error. this looks risky can you avoid sending dummy values from this funciton, fields like Id and all shouldn't be Maybe.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
```diff
@@ -0,0 +1,143 @@
...
+  let ghKey = makeGoHomeReqKey driverId
+  expTime <- fromIntegral <$> asks (.cacheConfig.configsExpTime)
+  Hedis.setExp ghKey (templateGoHomeData ghInfo.status ghInfo.cnt (Just $ addUTCTime (fromIntegral goHomeConfig.activeTime) currTime) ghInfo.driverGoHomeRequestId False (Just ghInfo.goHomeReferenceTime) currTime) expTime
+
+increaseDriverGoHomeRequestCount :: (MonadFlow m, CacheFlow m r, EsqDBFlow m r) => Id Merchant -> Id DP.Driver -> m ()
+increaseDriverGoHomeRequestCount merchantId driverId = do
+  currTime <- getLocalCurrentTime =<< ((CQTC.findByMerchantId merchantId >>= fromMaybeM (InternalError "Transporter config for timezone not found")) <&> (.timeDiffFromUtc))
+  let ghKey = makeGoHomeReqKey driverId
+  expTime <- fromIntegral <$> asks (.cacheConfig.configsExpTime)
+  ghInfo <- getDriverGoHomeRequestInfo driverId merchantId Nothing
```
piyushKumar-1, line 134: what if this API gets called before driver has enabled go home for today ? this will give you dummy template with no gohomeReqId right ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Ride.hs
```diff
@@ -93,7 +94,10 @@ findAllRidesByDriverId ::
   m [Ride]
 findAllRidesByDriverId (Id driverId) = findAllWithKV [Se.Is BeamR.driverId $ Se.Eq driverId]
 
-findActiveByRBId :: MonadFlow m => Id Booking -> m (Maybe Ride)
+findCompletedRideByGHRId :: (MonadFlow m) => Id DDGR.DriverGoHomeRequest -> m (Maybe Ride)
+findCompletedRideByGHRId (Id ghrId) = findOneWithKV [Se.And [Se.Is BeamR.driverGoHomeRequestId $ Se.Eq (Just ghrId), Se.Is BeamR.status $ Se.Eq DRide.COMPLETED]] -- To Be added
```
piyushKumar-1, line 98: what is to be added ?

piyushKumar-1: @arjuna92  @vijaygupta18 KV will not throw error if multiple entries are returned for findOne right ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -575,6 +611,112 @@ setActivity (personId, _) isActive mode = do
...
+  driverInfo <- QDriverInformation.findById driverId >>= fromMaybeM DriverInfoNotFound
+  unless driverInfo.enabled $ throwError DriverAccountDisabled
+  when (driverInfo.blocked) $ throwError DriverAccountBlocked
+  driverLocation <- QDrLoc.findById driverId >>= fromMaybeM LocationNotFound
+  let currPos = LatLong {lat = driverLocation.lat, lon = driverLocation.lon}
+  driverHomeLocation <- QDHL.findById driverHomeLocationId >>= fromMaybeM (DriverHomeLocationDoesNotExist driverHomeLocationId.getId)
+  let homePos = LatLong {lat = driverHomeLocation.lat, lon = driverHomeLocation.lon}
+  unless (distanceBetweenInMeters homePos currPos > fromIntegral goHomeConfig.destRadiusMeters) $ throwError DriverCloseToHomeLocation
+  dghInfo <- CQDGR.getDriverGoHomeRequestInfo driverId merchantId (Just goHomeConfig)
+  whenM (fmap ((isJust dghInfo.status) ||) (isJust <$> QDGR.findActive driverId)) $ throwError DriverGoHomeRequestAlreadyActive
```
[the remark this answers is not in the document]
piyushKumar-1: `fmap ((isJust dghInfo.status) ||) (isJust <$> QDGR.findActive driverId)`

anyways the above code can be written as:

`fmap (isJust . (dghInfo.status <|>)) $ QDGR.findActive driverId

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -157,7 +167,7 @@ dashboardEndRide handle rideId req =
     $ DashboardReq req
 
 endRide ::
-  EndRideFlow m r =>
+  (EndRideFlow m r, CacheFlow m r, ServiceFlow m r, EsqDBFlow m r, MonadThrow m, Log m, MonadTime m, MonadGuid m) =>
```
piyushKumar-1, line 170: try `MonadFlow m`, in all above also.

piyushKumar-1 (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
```diff
@@ -423,6 +429,99 @@ getDriverAverageSpeed merchantId driverId = Redis.withCrossAppRedis $ do
...
+        }
+
+    makeDriverPoolResult QP.NearestGoHomeDriversResult {variant = variant', ..} =
+      DriverPoolResult
+        { distanceToPickup = distanceToDriver,
+          variant = variant',
+          ..
+        }
+
+getDistanceBetweenCoords :: LatLong -> LatLong -> Meters
```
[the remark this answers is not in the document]
ratnadeep-99: Yes I saw but this Alex wrote I am not sure though why...... If you say so I will remove it

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
```diff
@@ -423,6 +429,99 @@ getDriverAverageSpeed merchantId driverId = Redis.withCrossAppRedis $ do
...
+  logDebug $ "secondly filtered go home driver pool" <> show filtDriverPoolWithActualDist
+  return filtDriverPoolWithActualDist
+  where
+    filterFunc threshold estDist = getMeters estDist.actualDistanceToPickup <= fromIntegral threshold
+    getParallelSearchRequestCount now dObj = getValidSearchRequestCount merchantId (dObj.driverId) now
+
+    getRoutesForAllDrivers =
+      zipWithM
+        ( \driver ghrId -> do
+            dghReq <- QDGR.findById ghrId >>= fromMaybeM (InternalError "Driver go home request not found.")
```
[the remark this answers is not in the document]
ratnadeep-99: Without request there wont be any home lat lon and without that how will pooling proceed?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
```diff
@@ -423,6 +429,99 @@ getDriverAverageSpeed merchantId driverId = Redis.withCrossAppRedis $ do
...
+            routes <-
+              DRoute.getTripRoutes (driver.driverId, merchantId) $
+                Maps.GetRoutesReq
+                  { waypoints = getCoordinates driver :| [getCoordinates dghReq],
+                    mode = Just Maps.CAR,
+                    calcPoints = True
+                  }
+            route <- listToMaybe routes & fromMaybeM (InternalError "Zero routes result when expected exactly one.")
```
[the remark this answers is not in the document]
ratnadeep-99: This will fail if directions api returns 0 routes. Now without directions api response and without any route how will pooling proceed?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
```diff
@@ -669,8 +767,25 @@ computeActualDistance orgId pickup driverPoolResults = do
...
+        }
+
+refactorRoutesResp :: GoHomeConfig -> (QP.NearestGoHomeDriversResult, Maps.RouteInfo, Id DDGR.DriverGoHomeRequest) -> (QP.NearestGoHomeDriversResult, Maps.RouteInfo, Id DDGR.DriverGoHomeRequest)
+refactorRoutesResp goHomeCfg (nearestDriverRes, route, ghrId) = (nearestDriverRes, newRoute route, ghrId)
+  where
+    newRoute route' =
+      RouteInfo
+        { distance = route'.distance,
+          duration = route'.duration,
+          points = tail $ refactor [] route'.points,
```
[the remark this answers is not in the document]
ratnadeep-99: yeah but a route with a single point? atleast start and dest lat lon will be there right?
Anyways writing my own implementation of tail

tail' [] = []
tail' (_:xs) = xs

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
```diff
@@ -669,8 +767,25 @@ computeActualDistance orgId pickup driverPoolResults = do
...
+          snappedWaypoints = route'.snappedWaypoints,
+          boundingBox = Nothing
         }
+    refactor acc (p1 : p2 : ps) = if getDistanceBetweenCoords p1 p2 > goHomeCfg.goHomeWayPointRadiusRadius then refactor (p1 : acc) (getPointInBetween p1 p2 (fromIntegral (goHomeCfg.goHomeWayPointRadiusRadius.getMeters)) : p2 : ps) else refactor (p1 : acc) (p2 : ps)
```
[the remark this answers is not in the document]
ratnadeep-99: Co-pilot

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
ratnadeep-99: But if go home is not enabled what should I set it? Also, This design we both discussed and agreed upon? Also this is getting set using simple logic if enabled make a query to db and use the generated id to set it and if we have a go home request without a id that does not make any sense.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
ratnadeep-99: yes. But not dummy data as count will be there and reference time to keep track of count will be there.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/CancelRide.hs
```diff
@@ -72,51 +77,84 @@ data CancelRideReq = CancelRideReq
...
+
+cancelRideHandler :: (MonadHandler m, CacheFlow m r, MonadFlow m, EsqDBFlow m r) => ServiceHandle m -> RequestorId -> Id DRide.Ride -> CancelRideReq -> m CancelRideResp
+cancelRideHandler sh requestorId rideId req = withLogTag ("rideId-" <> rideId.getId) do
+  cancellationCnt <- cancelRideImpl sh requestorId rideId req
+  pure $ buildCancelRideResp cancellationCnt
+  where
+    buildCancelRideResp cancelCnt =
+      CancelRideResp
+        { result = "success",
+          goHomeCancellationCount = cancelCnt
```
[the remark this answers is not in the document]
ratnadeep-99: when cancellations are 2 thats the flag that go home will be disabled (from requirements) frontend will show a popup saying go to disabled due to too many cancellations. This count is important as when count is 1  frontend will show a warning.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/Ride.hs
ratnadeep-99: Secondary key for KV. Its already added forgot to remove the comment.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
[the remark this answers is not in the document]
piyushKumar-1: can you elaborate the logic behind fetching just the last point ? that would ideally be the destination point right ? it seems wrong can you verify the logic once ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool.hs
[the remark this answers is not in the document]
piyushKumar-1: suggestion:

```
refactor acc (p1 : p2 : ps) 
    | getDistanceBetweenCoords p1 p2 > goHomeCfg.goHomeWayPointRadiusRadius = refactor (p1 : acc) (getPointInBetween p1 p2 (fromIntegral (goHomeCfg.goHomeWayPointRadiusRadius.getMeters)) : p2 : ps)
    | otherwise = refactor (p1 : acc) (p2 : ps)
refactor acc [p1] = reverse (p1 : acc)
refactor _ [] = []
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
piyushKumar-1: > if we have a go home request without a id that does not make any sense

agreed, that why m saying don't even create this dummy object with nothing everywhere and some fields set to defaults. For now you have handled everywhere, but in future this may lead to wrong interpretation of what this function is doing and someone might use it wrongly.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/SendSearchRequestToDrivers/Handle/Internal/DriverPool.hs
```diff
@@ -78,56 +82,67 @@ prepareDriverPoolBatch driverPoolCfg searchReq searchTry batchNum = withLogTag (
...
-          else calculatePool batchSize sortingType onlyNewDrivers allNearbyDrivers intelligentPoolConfig transporterConfig
-      pure $ addDistanceSplitConfigBasedDelaysForDriversWithinBatch driverPoolWithoutBatchSplitDelays
+      goHomeConfig <- CQGHC.findByMerchantId searchReq.providerId
+      allNearbyGoHomeDrivers <-
+        if batchNum == 0 && goHomeConfig.enableGoHome
+          then calcGoHomeDriverPool goHomeConfig
+          else return []
+      currentDriverPoolBatch <-
+        if notNull allNearbyGoHomeDrivers
+          then calculateGoHomeBatch transporterConfig intelligentPoolConfig allNearbyGoHomeDrivers blockListedDrivers
```
piyushKumar-1, line 102: even if only one goHome valid driver is there then in first batch we will send only to that driver ? cc: @Prithivirajkota let's do go home drivers + (few batchSize drivers with slight delay in popup, 3-4 sec after goHome is sent)

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/Driver/GoHomeRequest.hs
ratnadeep-99: But whole point of creating this was to keep count in the redis and expire it every day. So its not dummy data count is there and reference time is keeping track which day's count it is. count is needed everywhere very frequently be it to dispaly or be it to show whether driver is eligible for go to request. So even when ghrid is null go to count will be there which is not dummy its an important parameter which is to be kept track of and expired/reset as per requirement time to time. Infact count was the reason you suggested to use redis you remember?

piyushKumar-1: sure, then lets change the type of object that we are storing to something like this:

```
data CachedGoHomeRequest = CachedGoHomeRequest
    {  goHomeRequest: Maybe GoHomeReqInfo
    ,  goHomeMeta: GoHomeMeta
    , status :: DriverGoHomeRequestStatus -> (Active | Inactive | Success | Failure)
    }
data GoHomeReqInfo = GoHomeReqInfo
    { driverGoHomeRequestId :: Id DriverGoHomeRequest
    , validTill :: UTCTime
    }
data GoHomeMeta = GoHomeMeta 
    {  count :: Int
    ,  goHomeReferenceTime :: UTCTime
    }
```

ratnadeep-99: Just sharing my perspective, here success and failure has no role this DS is for storing count, reqId and validity related info whether go to is success or failure same will be tracked in the DB. And here also if I want to access Go Home Request Id I have to do fromMaybeM only which was the issue as you were saying. In new one Like goHomeReqId and validTill here will be nothing (together) when status is not active and in the current one they will be nothing separately . I suggest lets just add proper commenting in the existing type so that in future if someone is reading it will have complete context. Also this data is being used very often throughout the code, I can restructure but that will be time consuming and Maybe affect is still there also now extra get fields will have to handle. Now it's up to you should we keep the current with proper commenting added in the type explaining meaning of each param or should we add this type, Not sure whether benefits will be worth the effort.

piyushKumar-1: Even if there is no significance in this data when its in redis the haskell type will have these constructors, otherwise how will you add them to DB and regarding the type change, i suggest to go with the types i proposed as it's more readable and adding inactive as a constructor will make it more explicit that go home is enabled or not. Also having few dependent fields as nothing separately can create issues in future, i agree it's handled rn, but if id and validTill can't exist without each other then ideally they both should he nothing or both should have a value that is ensured by the type i suggested this way you will not have to handle if validity exist and id doesn't etc kind of weird cases that should never come and if they came it would be hard to debug with the current type. If its very time consuming then maybe we can create another issue for refactoring and till then you can resolve other comments. @hkmangla  your views on this....should we do now or later ? Later we may have to backfill data from older type in redis to newer type if released during daytime or add backwards compatibility code...

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/CancelRide.hs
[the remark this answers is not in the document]
ratnadeep-99: Then endride api should have same thing as well i.e. show true when driver reaches home else show false.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/Utils.hs
```diff
@@ -410,6 +416,9 @@ buildRadiusWithin'' (lat, lon) rad =
 (<->.) :: (Double, Double) -> BQ.QGenExpr context Postgres s Double
 (<->.) (lat, lon) = BQ.QExpr (\_ -> PgExpressionSyntax (emit $ "point <-> " <> "ST_SetSRID (ST_Point (" <> show lon <> " , " <> show lat <> "),4326)"))
 
+byDist :: (Double, Double) -> BQ.QGenExpr context Postgres s Double
+byDist (lat, lon) = BQ.QExpr (\_ -> PgExpressionSyntax (emit $ "point <-> 'SRID=4326;POINT(" <> show lon <> " " <> show lat <> ")'"))
```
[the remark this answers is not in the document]
ratnadeep-99: So this was getting used by isWithinRange query which was getting used in search request. But that day we removed searchrequest implementation and made it direct from driverpool. So technically we should be able to safely remove byDist  as well as isWithinRange query. In the latest branch also byDist is not there. So should I delete it and is withinRange? As this in in Lib thats why asking for confirmation once. Please check once and let me know.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/Utils.hs
```diff
@@ -410,6 +416,9 @@ buildRadiusWithin'' (lat, lon) rad =
 (<->.) :: (Double, Double) -> BQ.QGenExpr context Postgres s Double
 (<->.) (lat, lon) = BQ.QExpr (\_ -> PgExpressionSyntax (emit $ "point <-> " <> "ST_SetSRID (ST_Point (" <> show lon <> " , " <> show lat <> "),4326)"))
```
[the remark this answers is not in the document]
ratnadeep-99: The time when I these queries we converted to beam I guess that time we were using byDist only which was later removed. I kept it till now thinking it is getting used by isWithinRange query.

piyushKumar-1 (approved):

PR #3378, opened 2023-09-26
backend/feat/enable rc check for express onboarding

max-keviv:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Earlier - RC check was being ignored on add vehicle 
Now - RC check is being enabled 

## Testing
earlier
Screenshot 2023-09-25 at 6 12 49 PM
Screenshot 2023-09-25 at 6 13 34 PM
Now
Screenshot 2023-09-26 at 9 52 45 AM
Screenshot 2023-09-26 at 9 52 59 AM


max-keviv:
@hemantmangla @0utkarsh  In testing the only check it was doing was the prefix check if merchant is NY but it was not check RC no is valid or not so should we add a check onVerifyRC that if output.registration_number is nothing then throwError

0utkarsh:
@max-keviv Only prefix checking is not enough. Did you check what response you get when RC is invalid? Based on that you can choose to throw error accordingly.

AshishSaini25 (approved):

PR #3533, opened 2023-10-07
Revert "backend/fix : rounding up driver rating (#3532)"

khuzema786:


khuzema786:
Handled with merchant config

Issue #3386, opened 2023-09-26
Driver Loyalty Program: Coins (Backend)


tejkranthi:
We are starting a loyalty program for Namma Yatri drivers. This program will be based on a closed-loop wallet that will introduced in the driver app called Yatri Coins. Driver will earn Yatri Coins whenever they exhibit good behaviour (Ex: ride completed, customer referral, etc.) and coins will be deducted from their wallet for bad behaviour (Ex: Cancellation). Each driver will earn coins for referrals. These coins can be used to get discount on their subscription plans.

PR #3535, opened 2023-10-07
backend/feat:#3386 driver earnings screen changes

Shikhar-24:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

- created a rideSummary API that will give the stats of driver on the basis of list of dates provided
- added a DialyStats table 

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/UI/RideSummary.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/UI/RideSummary.hs
@@ -1,0 +1,41 @@
+{-
+ Copyright 2022-23, Juspay India Pvt Ltd
+
+ This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License
+
+ as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program
+
+ is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+
+ or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details. You should have received a copy of
+
+ the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+-}
+
+module API.UI.RideSummary where
+
+import Data.Time (Day)
+import Domain.Action.UI.RideSummary as RDS
+import qualified Domain.Types.Merchant as Merchant
+import qualified Domain.Types.Merchant.MerchantOperatingCity as DM
+import qualified Domain.Types.Person as DP
+import Environment
+import Kernel.Prelude
+import Kernel.Types.Id
+import Kernel.Utils.Common
+import Servant
+import Tools.Auth
+
+type API =
+  "rideSummary"
+    :> ( "list"
+           :> TokenAuth
+           :> ReqBody '[JSON] [Day]
+           :> Post '[JSON] RDS.DriverRideSummaryListResp
+       )
+
+handler :: FlowServer API
+handler = listDailyStats
+
+listDailyStats :: (Id DP.Person, Id Merchant.Merchant, Id DM.MerchantOperatingCity) -> [Day] -> FlowHandler RDS.DriverRideSummaryListResp
+listDailyStats (driverId, merchantId, merchantOpCityId) = withFlowHandlerAPI . RDS.listDailyRidesSummary driverId merchantId merchantOpCityId
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
@@ -1,0 +1,58 @@
+{-
+ Copyright 2022-23, Juspay India Pvt Ltd
+
+ This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General Public License
+
+ as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version. This program
+
+ is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+
+ or FITNESS FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more details. You should have received a copy of
+
+ the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+-}
+{-# LANGUAGE DerivingStrategies #-}
+
+module Domain.Action.UI.RideSummary where
+
+import Data.Time (Day)
+import Domain.Types.DailyStats as DDS
+import qualified Domain.Types.Merchant as Merchant
+import qualified Domain.Types.Merchant.MerchantOperatingCity as DM
+import qualified Domain.Types.Person as DP
+import Kernel.Prelude
+import Kernel.Types.Common
+import Kernel.Types.Id
+import Kernel.Utils.Common
+import qualified Storage.Queries.DailyStats as SQDS
+
+data DriverRideSummaryResp = DriverRideSummaryResp
+  { earnings :: Money,
+    rideDistance :: Meters,
+    rideDate :: Day,
+    noOfRides :: Int
+  }
+  deriving (Generic, Show, FromJSON, ToJSON, ToSchema)
+
+newtype DriverRideSummaryListResp = DriverRideSummaryListResp
+  { list :: [DriverRideSummaryResp]
+  }
+  deriving (Generic, Show, FromJSON, ToJSON, ToSchema)
+
+listDailyRidesSummary :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => Id DP.Person -> Id Merchant.Merchant -> Id DM.MerchantOperatingCity -> [Day] -> m DriverRideSummaryListResp
+listDailyRidesSummary driverId _ _ req = do
+  dailyStatsList <- mapM (SQDS.findByDriverIdAndDate driverId) req
+  let list = mkRideSummaryList (catMaybes dailyStatsList)
+  return $ DriverRideSummaryListResp {..}
+
+mkRideSummaryList :: [DDS.DailyStats] -> [DriverRideSummaryResp]
+mkRideSummaryList =
+  map
+    ( \x ->
+        DriverRideSummaryResp
+          { earnings = x.totalEarnings,
+            rideDistance = x.totalDistance,
+            rideDate = x.merchantLocalDate,
+            noOfRides = x.numRides
+          }
+    )
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/DriverScore.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/DriverScore.hs
@@ -17,6 +17,8 @@ module Lib.DriverScore
   )
 where
 
+import Data.Time (utctDay)
+import qualified Domain.Types.DailyStats as DDS
 import qualified Domain.Types.DriverStats as DS
 import qualified Domain.Types.Merchant.MerchantOperatingCity as DMOC
 import qualified Domain.Types.Person as DP
@@ -26,12 +28,13 @@ import qualified Kernel.Beam.Functions as B
 import Kernel.Prelude
 import Kernel.Storage.Esqueleto as Esq
 import Kernel.Types.Id (Id, cast)
-import Kernel.Utils.Common (CacheFlow, Forkable (fork), Money (Money), fromMaybeM, getCurrentTime, getMoney, highPrecMetersToMeters, logDebug)
+import Kernel.Utils.Common (CacheFlow, Forkable (fork), MonadGuid (generateGUIDText), Money (Money), fromMaybeM, getCurrentTime, getLocalCurrentTime, getMoney, highPrecMetersToMeters, logDebug)
 import qualified Lib.DriverScore.Types as DST
 import qualified SharedLogic.DriverPool as DP
 import qualified Storage.CachedQueries.Merchant.TransporterConfig as CTCQ
 import qualified Storage.Queries.Booking as BQ
 import qualified Storage.Queries.BookingCancellationReason as BCRQ
+import qualified Storage.Queries.DailyStats as SQDS
 import qualified Storage.Queries.DriverInformation as QDI
 import qualified Storage.Queries.DriverStats as DSQ
 import qualified Storage.Queries.FareParameters as FPQ
@@ -83,9 +86,10 @@ eventPayloadHandler merchantOpCityId DST.OnDriverCancellation {..} = do
     nonZero (Just a)
       | a <= 0 = 1
       | otherwise = a
-eventPayloadHandler _ DST.OnRideCompletion {..} = do
+eventPayloadHandler merchantOpCityId DST.OnRideCompletion {..} = do
   mbDriverStats <- B.runInReplica $ DSQ.findById (cast driverId) -- always be just because stats will be created at OnNewRideAssigned
   -- mbDriverStats <- DSQ.findById (cast driverId) -- always be just because stats will be created at OnNewRideAssigned
+  updateDailyStats driverId merchantOpCityId ride
   whenJust mbDriverStats $ \driverStats -> do
     (incrementTotaEarningsBy, incrementBonusEarningsBy, incrementLateNightTripsCountBy, overallPickupCharges) <-
       if isNotBackFilled driverStats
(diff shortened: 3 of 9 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/UI/RideSummary.hs
```diff
@@ -0,0 +1,41 @@
...
+import Kernel.Utils.Common
+import Servant
+import Tools.Auth
+
+type API =
+  "rideSummary"
+    :> ( "list"
+           :> TokenAuth
+           :> MandatoryQueryParam "fromDate" Day
+           :> MandatoryQueryParam "toDate" Day
```
0utkarsh, line 33: Maybe not have these mandatory and send today's stats by default?

Backend/dev/migrations/dynamic-offer-driver-app/0285-dailyStats-table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS atlas_driver_offer_bpp.daily_stats (
+    id VARCHAR(36) NOT NULL PRIMARY KEY,
+    driver_id character(36) NOT NULL,
+    total_earnings INT,
+    number_of_rides INT,
+    total_distance INT,
+    of_date Date NOT NULL
```
0utkarsh, line 7:
```
date
num_rides
```
looks better but your call
Also, backfilling required

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
```diff
@@ -0,0 +1,76 @@
...
+  case maybeDailyStats of
+    Just a -> pure a
+    Nothing -> fetchDailyStatsfromKV day driverId
+
+fetchDailyStatsfromKV :: (CacheFlow m r, EsqDBFlow m r) => Day -> Id DP.Person -> m DDS.DailyStats
+fetchDailyStatsfromKV day driverId = do
+  mbDailyStats <- SQDS.findByDriverIdAndDate driverId day
+  case mbDailyStats of
+    Just a -> pure a
+    Nothing -> pure $ DDS.DailyStats {id = "", driverId = driverId, totalEarnings = 0, numberOfRides = 0, totalDistance = 0, ofDate = day}
```
0utkarsh, line 61: `Maybe DailyStats` can be sent from here and `catMaybe` can be used by the caller instead of sending dummy data

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/UI/RideSummary.hs
Shikhar-24: Actually the use case here is not like that 
We will be sending fromDate and toDate from frontend (always Monday to sunday), it doesn't matter if today is tuesday or wednesday.
you can see this from figma
https://www.figma.com/file/4zT4QhImxqphGf56DpAjcV/Beckn-Release-Version?type=design&node-id=82227-37499&mode=design&t=FPCQXmpWAatIQVmP-0

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
```diff
@@ -0,0 +1,76 @@
...
+
+newtype DriverRideSummaryListResp = DriverRideSummaryListResp
+  { list :: [DriverRideSummaryResp]
+  }
+  deriving (Generic, Show, FromJSON, ToJSON, ToSchema)
+
+listDriverRidesSummary :: (CacheFlow m r, EsqDBFlow m r, MonadFlow m) => Id DP.Person -> Id Merchant.Merchant -> Day -> Day -> m DriverRideSummaryListResp
+listDriverRidesSummary personId _ fromDay toDay = do
+  let allDates = datesBetween fromDay toDay
+  dailyStatsList <- mapM (`findStatsForDate` personId) allDates
```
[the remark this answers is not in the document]
Shikhar-24: i need value of each date in my output
So suppose if difference between fromDate and toDate is of seven days
My response would be of an array of length seven, with each object giving the stats of that particular date

Backend/dev/migrations/dynamic-offer-driver-app/0285-dailyStats-table.sql
Shikhar-24: We already discussed about it and backfilling is not needed 
We will start storing and caching data from the day of release itself

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/UI/RideSummary.hs
0utkarsh: In that case, why even take these as params, just send current week's data? Anyways, its fine. You can send weekly data by default and make these optional

Backend/dev/migrations/dynamic-offer-driver-app/0285-dailyStats-table.sql
0utkarsh: But why? Backfilling shouldn't be an issue. We can just write one query to fetch these details from ride table

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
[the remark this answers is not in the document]
0utkarsh: Sure, but you can do `rideDate >= fromDate && rideDate <= toDate` instead of generating all dates and comparing

Backend/dev/migrations/dynamic-offer-driver-app/0285-dailyStats-table.sql
Shikhar-24: The whole thing was done for not to fetch data from ride table
We will be showing data in bar graphs for last 4 weeks, i.e 28 days and backfilling for each driver for last 28 days and calculating total for each day would've been an extra load. Hence we went for this.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -317,7 +321,33 @@ endRide handle@ServiceHandle {..} rideId req = withLogTag ("rideId-" <> rideId.g
     let mbPaymentUrl = DMPM.getPostpaidPaymentUrl =<< mbPaymentMethod
     let mbPaymentMethodInfo = DMPM.mkPaymentMethodInfo <$> mbPaymentMethod
     withTimeAPI "endRide" "notifyCompleteToBAP" $ notifyCompleteToBAP booking updRide newFareParams mbPaymentMethodInfo mbPaymentUrl
-
+    let dateInIST = addUTCTime 19800 now
```
0utkarsh, line 324: Also, lets take this 19800 from `transporterConfig.timeDiffFromUtc`

0utkarsh (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
Shikhar-24: the reason i am sending sort of a dummydata with date here is because if I send data after using catmaybe then I will have to handle it afterwards searching for each date and then use that value and if that date is not present then I have to do the same.

0utkarsh: Don't know about frontend code but from the above statement it looks like you'll always assume fixed number of elements in response. Will UI break if `toDate - fromDate > 28`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
```diff
@@ -0,0 +1,76 @@
...
+  case mbDailyStats of
+    Just a -> pure a
+    Nothing -> pure $ DDS.DailyStats {id = "", driverId = driverId, totalEarnings = 0, numberOfRides = 0, totalDistance = 0, ofDate = day}
+
+mkRideSummaryList :: [DDS.DailyStats] -> [DriverRideSummaryResp]
+mkRideSummaryList =
+  map
+    ( \x ->
+        DriverRideSummaryResp
+          { earnings = Just x.totalEarnings,
```
pius-1, line 68: why is this maybe 🤔

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
pius-1: we can avoid any restriction of dates `to` `from` range. whatever this query gets should return back, with some basic sanity of not having a lot of dates. also can we make this API expect `[Dates]` from frontend itself and return data as `[Maybe Stats]` back, defaults can be handled at frontend side.

Also as we are planning to make it cached at frontend so if there is only a few days data to be fetched we can just send that date in the API and get the data for only those dates, i.e. when a week is ongoing then we can just fetch the new data everyday...?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -318,7 +321,32 @@ endRide handle@ServiceHandle {..} rideId req = withLogTag ("rideId-" <> rideId.g
...
+        pure
+          DDS.DailyStats
+            { id = dailyStats.id,
+              driverId = driverId,
+              totalEarnings = dailyStats.totalEarnings + finalFare,
+              numRides = dailyStats.numRides + 1,
+              totalDistance = dailyStats.totalDistance + chargeableDistance,
+              date = utctDay timeInIST
+            }
+    SQDS.upsertByDriverId dailyStatsOfDriver'
```
piyushKumar-1, line 349: in DB keep it as UTC Time instead of IST date and handle timestamps bounds to query for a day in a specific localZones using the timeDIffFromUTC Config.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Beam/DailyStats.hs
```diff
@@ -0,0 +1,44 @@
...
+
+instance B.Table DailyStatsT where
+  data PrimaryKey DailyStatsT f
+    = Id (B.C f Text)
+    deriving (Generic, B.Beamable)
+  primaryKey = Id . driverId
+
+type DailyStats = DailyStatsT Identity
+
+$(enableKVPG ''DailyStatsT ['driverId] [])
```
piyushKumar-1, line 42: it should be `$(enableKVPG ''DailyStatsT ['id] [['driverId]])`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
```diff
@@ -0,0 +1,61 @@
...
+  deriving (Generic, Show, FromJSON, ToJSON, ToSchema)
+
+newtype DriverRideSummaryListResp = DriverRideSummaryListResp
+  { list :: [DriverRideSummaryResp]
+  }
+  deriving (Generic, Show, FromJSON, ToJSON, ToSchema)
+
+listDailyRidesSummary :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => Id DP.Person -> Id Merchant.Merchant -> Id DM.MerchantOperatingCity -> [Day] -> m DriverRideSummaryListResp
+listDailyRidesSummary driverId _ _ req = do
+  dailyStatsList <- mapM (`findStatsForDate` driverId) req
```
piyushKumar-1, line 44: why not use `mapM (SQDS.findByDriverIdAndDate driverId) req` directly ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -318,7 +321,32 @@ endRide handle@ServiceHandle {..} rideId req = withLogTag ("rideId-" <> rideId.g
     let mbPaymentUrl = DMPM.getPostpaidPaymentUrl =<< mbPaymentMethod
     let mbPaymentMethodInfo = DMPM.mkPaymentMethodInfo <$> mbPaymentMethod
     withTimeAPI "endRide" "notifyCompleteToBAP" $ notifyCompleteToBAP booking updRide newFareParams mbPaymentMethodInfo mbPaymentUrl
-
+    transporterConfig <- QTConf.findByMerchantOpCityId booking.merchantOperatingCityId >>= fromMaybeM (TransporterConfigNotFound booking.merchantOperatingCityId.getId)
```
piyushKumar-1, line 324: do all this inside `driverScoreEventHandler` function, update the `OnRideCompletion` type to add any additional data required for these calculations.

piyushKumar-1 (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
Shikhar-24: I am storing date in form of "yyyy-mm-dd" and not in complete form of UTCTime. 
From frontend we are sending dates according to LocalTime and in backend if we will store dates in UTC then dates of two different timezones can't be compared accordingly as full UTCTime can't be stored in this case. 
What should I do here ?

piyushKumar-1: sure, you are already using `timeDIffFromUTC` config, just name the date column in Db as merchantLocalDate and also don't use IST in any variables you defined above, it gave an impression of being hardcoded to IST.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/RideSummary.hs
```diff
@@ -0,0 +1,58 @@
...
+  dailyStatsList <- mapM (SQDS.findByDriverIdAndDate driverId) req
+  let list = mkRideSummaryList (catMaybes dailyStatsList)
+  return $ DriverRideSummaryListResp {..}
+
+mkRideSummaryList :: [DDS.DailyStats] -> [DriverRideSummaryResp]
+mkRideSummaryList =
+  map
+    ( \x ->
+        DriverRideSummaryResp
+          { earnings = Just x.totalEarnings,
```
piyushKumar-1, line 53: Why is this maybe ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/DriverScore.hs
```diff
@@ -124,6 +128,35 @@ eventPayloadHandler _ DST.OnRideCompletion {..} = do
...
+      pure
+        DDS.DailyStats
+          { id = dailyStats.id,
+            driverId = driverId,
+            totalEarnings = dailyStats.totalEarnings + fromMaybe 0 ride.fare,
+            numRides = dailyStats.numRides + 1,
+            totalDistance = dailyStats.totalDistance + fromMaybe 0 ride.chargeableDistance,
+            merchantLocalDate = utctDay localTime
+          }
+  SQDS.upsertByDriverId dailyStatsOfDriver'
```
piyushKumar-1, line 158: You are already checking if ds for that day exist or not then why do it again in upsert query just do create or update in the case match above.

Backend/dev/migrations/dynamic-offer-driver-app/0285-dailyStats-table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE IF NOT EXISTS atlas_driver_offer_bpp.daily_stats (
+    id VARCHAR(36) NOT NULL PRIMARY KEY,
+    driver_id character(36) NOT NULL,
+    total_earnings INT,
+    num_rides INT,
+    total_distance INT,
+    merchant_local_date Date NOT NULL
+);
```
piyushKumar-1, line 8: Add create composite index query on (driver id , date) as well

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/DriverScore.hs
```diff
@@ -124,6 +128,36 @@ eventPayloadHandler _ DST.OnRideCompletion {..} = do
...
+      let dailyStatsOfDriver' =
+            DDS.DailyStats
+              { id = dailyStats.id,
+                driverId = driverId,
+                totalEarnings = dailyStats.totalEarnings + fromMaybe 0 ride.fare,
+                numRides = dailyStats.numRides + 1,
+                totalDistance = dailyStats.totalDistance + fromMaybe 0 ride.chargeableDistance,
+                merchantLocalDate = utctDay localTime
+              }
+      SQDS.upsertByDriverId dailyStatsOfDriver'
```
piyushKumar-1, line 159: name it update instead of upsert, its not doing upsert now. Also why send the whole driverStats object, just send what to update as args.

Shikhar-24: done

piyushKumar-1 (approved):

Issue #4066, opened 2023-10-30
Add logs for customer and driver for app loading time


ashkriti:
Add logs for customer and driver for app loading time

PR #4067, opened 2023-10-30
frontend/fix/#4066/add logs for app load time

ashkriti:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
Added logs for app loads time and sending to CleverTap

## Motivation and Context
https://github.com/nammayatri/nammayatri/issues/4066

## How did you test it?
Tested manually

## Checklist

### DIFF
--- a/Frontend/android-native/app/src/main/java/in/juspay/mobility/MainActivity.java
+++ b/Frontend/android-native/app/src/main/java/in/juspay/mobility/MainActivity.java
@@ -113,6 +113,7 @@ public class MainActivity extends AppCompatActivity {
     @SuppressLint("StaticFieldLeak")
     private static InAppNotification inAppNotification;
     ShowNotificationCallBack inappCallBack;
+    long onCreateTimeStamp = 0;
     SharedPreferences.OnSharedPreferenceChangeListener mListener = new SharedPreferences.OnSharedPreferenceChangeListener() {
         @Override
         public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
@@ -276,6 +277,8 @@ public class MainActivity extends AppCompatActivity {
         }
 
         super.onCreate(savedInstanceState);
+        onCreateTimeStamp = System.currentTimeMillis();
+        FirebaseAnalytics mFirebaseAnalytics = FirebaseAnalytics.getInstance(this);
         context = getApplicationContext();
         boolean isMigrated = migrateLocalStore(context);
         String clientId = context.getResources().getString(R.string.client_id);
@@ -457,7 +460,7 @@ public class MainActivity extends AppCompatActivity {
     }
 
     private void initApp(String viewParam, String deepLinkJSON) {
-
+        long initiateTimeStamp = System.currentTimeMillis();
         hyperServices = new HyperServices(this, findViewById(R.id.cl_dui_container));
         final JSONObject json = new JSONObject();
         JSONObject payload = new JSONObject();
@@ -471,6 +474,8 @@ public class MainActivity extends AppCompatActivity {
             if (viewParam != null) payload.put("viewParam", viewParam);
             if (viewParam != null) payload.put("view_param", viewParam);
             if (deepLinkJSON != null) payload.put("deepLinkJSON", deepLinkJSON);
+            payload.put("onCreateTimeStamp", onCreateTimeStamp);
+            payload.put("initiateTimeStamp" , initiateTimeStamp);
             json.put(PaymentConstants.PAYLOAD, payload);
         } catch (JSONException e) {
             e.printStackTrace();
--- a/Frontend/ui-common/src/Engineering/Helpers/Commons.js
+++ b/Frontend/ui-common/src/Engineering/Helpers/Commons.js
@@ -481,3 +481,24 @@ export const getImageUrl = function (url) {
     console.log("error in getImageUrl " + e);
   }
 };
+
+export const setEventTimestamp = function(string){
+  return function(){
+    if (!window.flowTimeStampObject[string]){
+      window.flowTimeStampObject[string] = Date.now() - window.prevTimeStamp;
+      window.prevTimeStamp = Date.now();
+    }
+  }
+}
+
+export const getTimeStampObject = function(){
+  return function(){
+    const keyValuePairArray = Object.keys(window.flowTimeStampObject).map(function(key) {
+      return {
+        key: key,
+        value: window.flowTimeStampObject[key]
+      };
+    });
+    return keyValuePairArray;
+  }
+}
--- a/Frontend/ui-driver/index.js
+++ b/Frontend/ui-driver/index.js
@@ -2,6 +2,8 @@ import "core-js";
 import "presto-ui";
 import "regenerator-runtime/runtime";
 
+const bundleLoadTime = Date.now();
+window.flowTimeStampObject = {};
 const blackListFunctions = ["getFromSharedPrefs", "getKeysInSharedPref", "setInSharedPrefs", "addToLogList", "requestPendingLogs", "sessioniseLogs", "setKeysInSharedPrefs", "getLayoutBounds"]
 
 if (window.JBridge.firebaseLogEventWithParams){  
@@ -175,6 +177,16 @@ window.onMerchantEvent = function (_event, payload) {
       // window.merchantID = clientPaylod.payload.clientId.toUpperCase();
       window.merchantID = "NAMMAYATRI";
     }
+    if (clientPaylod.payload && clientPaylod.payload.hasOwnProperty('onCreateTimeStamp') && clientPaylod.payload.hasOwnProperty('initiateTimeStamp'))
+    {
+      const onCreateTimeStamp = clientPaylod.payload.onCreateTimeStamp;
+      const initiateTimeStamp = clientPaylod.payload.initiateTimeStamp;
+      window.flowTimeStampObject["onCreateToBundle"] = window.prevTimeStamp - onCreateTimeStamp;
+      window.flowTimeStampObject["nativeIntiateToBundle"] = window.prevTimeStamp - initiateTimeStamp;
+    }
+    window.flowTimeStampObject["bundleLoadTime"] = bundleLoadTime - window.prevTimeStamp;
+    window.flowTimeStampObject["bundleToInitiate"] = Date.now() - bundleLoadTime;
+    window.prevTimeStamp = Date.now();
     callInitiateResult();
   } else if (_event == "process") {
     window.__payload.sdkVersion = "2.0.1"
--- a/Frontend/ui-driver/src/Flow.purs
+++ b/Frontend/ui-driver/src/Flow.purs
@@ -52,7 +52,7 @@ import Effect.Aff (makeAff, nonCanceler, launchAff)
 import Effect.Class (liftEffect)
 import Effect.Uncurried (runEffectFn1, runEffectFn5, runEffectFn2)
 import Engineering.Helpers.BackTrack (getState, liftFlowBT)
-import Engineering.Helpers.Commons (flowRunner, liftFlow, getNewIDWithTag, getVersionByKey, os, getExpiryTime, stringToVersion, setText, convertUTCtoISC, getCurrentUTC, getCurrentTimeStamp, clearTimer)
+import Engineering.Helpers.Commons (flowRunner, liftFlow, getNewIDWithTag, getVersionByKey, os, getExpiryTime, stringToVersion, setText, convertUTCtoISC, getCurrentUTC, getCurrentTimeStamp, clearTimer, setEventTimestamp, getTimeStampObject)
 import Engineering.Helpers.Commons as EHC
 import Engineering.Helpers.LogEvent (logEvent, logEventWithParams, logEventWithMultipleParams)
 import Engineering.Helpers.Suggestions (suggestionsDefinitions, getSuggestions)
@@ -125,6 +125,7 @@ import ConfigProvider
 
 baseAppFlow :: Boolean -> Maybe Event -> FlowBT String Unit
 baseAppFlow baseFlow event = do
+    liftFlowBT $ setEventTimestamp "baseAppFlow"
     versionCode <- lift $ lift $ liftFlow $ getVersionCode
     checkVersion versionCode
     checkTimeSettings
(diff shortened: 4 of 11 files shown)


Frontend/ui-driver/src/Flow.purs
```diff
@@ -1744,8 +1776,8 @@ onBoardingSubscriptionScreenFlow onBoardingSubscriptionViewCount = do
         Nothing -> onBoardingSubscriptionScreenFlow (state.props.screenCount-1)
   pure unit
 
-homeScreenFlow :: FlowBT String Unit
-homeScreenFlow = do
+homeScreenFlow :: Boolean -> FlowBT String Unit
+homeScreenFlow check = do
```
ashishsingh101, line 1780: why we need to pass boolean value to check wheather to call event or not ?

ashishsingh101: instead just pass this boolean to js function which will update the key and make flag true, so next when we check and if the flag is true will call the event

ashishsingh101 (changes requested):

Frontend/ui-driver/src/Flow.purs
ashkriti: We need to call this after all the API calls are done

ashishsingh101: maintain seperate flag for this

Frontend/ui-driver/src/Flow.purs
```diff
@@ -234,13 +236,18 @@ ifNotRegistered _ = getValueToLocalStore REGISTERATION_TOKEN == "__failed"
 isTokenValid :: String -> Boolean
 isTokenValid = (/=) "__failed"
 
-loginFlow :: FlowBT String Unit
-loginFlow = do
+loginFlow :: Boolean -> FlowBT String Unit
+loginFlow checkLog = do
```
ashishsingh101, line 240: why this `checkLog` was required ?

Frontend/ui-driver/src/Flow.purs
```diff
@@ -412,21 +419,28 @@ checkStatusAndStartLocationUpdates = do
   setDriverStatusInLocal (show (not isOffline)) $ show $ getDriverStatusFromMode currentMode
   liftFlowBT $ if isOffline then stopLocationPollingAPI else startLocationPollingAPI
 
-onBoardingFlow :: FlowBT String Unit
-onBoardingFlow = do
+onBoardingFlow :: LazyCheck -> FlowBT String Unit
+onBoardingFlow check = do
```
ashishsingh101, line 423: why we need lazy check here ?

ashishsingh101: other places as well

ashishsingh101 (changes requested):

Frontend/ui-driver/src/Flow.purs
ashkriti: made the changes

Frontend/ui-driver/src/Flow.purs
```diff
@@ -241,6 +242,11 @@ loginFlow = do
   runInternetCondition
   void $ pure $ setCleverTapUserProp [{key : "Preferred Language", value : unsafeToForeign $ getValueFromConfig "defaultLanguage"}]
   setValueToLocalStore LANGUAGE_KEY $ getValueFromConfig "defaultLanguage"
+  if (getToBeLog unit) then do
```
ashishsingh101, line 245: this is not required

ashishsingh101: we can have two flags in purs global state, one will tell should we send the log and another flag for should we add timestamp in object if we have not sended the log

Frontend/ui-driver/src/Flow.purs
```diff
@@ -414,11 +420,18 @@ checkStatusAndStartLocationUpdates = do
...
   (DriverRegistrationStatusResp resp ) <- driverRegistrationStatusBT (DriverRegistrationStatusReq { })
   lift $ lift $ doAff do liftEffect hideSplash
   GlobalState globalState <- getState
   if (resp.dlVerificationStatus == "NO_DOC_AVAILABLE" && resp.rcVerificationStatus == "NO_DOC_AVAILABLE") then do
+    let showLog = getToBeLog unit
+    if showLog then do
+      _ <- pure $ getTimestampObject "onBoardingFlow"
+      let logData = getData unit
+      liftFlowBT $ logEventWithMultipleParams logField_ "sending_logs" logData
+      else pure unit
```
ashishsingh101, line 434: `if-else` could be replace with `when`

ashishsingh101: but only in flow type not global events

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,26 @@ export const getImageUrl = function (url) {
...
 };
+
+export const getTimestampObject = function(string){
+  window.jsonObject[string] = Date.now() - window.prevTimeStamp;
+  window.prevTimeStamp = Date.now();
+  if (string == "baseAppFlow")
+    window.tobeLog = true;
+  else
+    window.tobeLog = false;
+}
```
ashishsingh101, line 470: add check to skip if key name is already present

ashishsingh101 (changes requested):

Frontend/ui-driver/src/Flow.purs
ashkriti: We are sending the log to CleverTap only after we are getting the timestamp, adding the timestamp or not is being covered inside the `tobelog` check in the flow file at every place required. So why do we need 2 checks if we want to log just after we get the timestamp?

Frontend/ui-common/src/Engineering/Helpers/Commons.js
ashkriti: Why do we need this? Even if the user comes back to the same flow the `tobeLog` check will handle this right as it will not come inside this function only.

Frontend/ui-driver/src/Flow.purs
ashkriti: Discussed on this

Frontend/android-native/app/src/main/java/in/juspay/mobility/MainActivity.java
```diff
@@ -480,12 +482,15 @@ private void initApp(String viewParam, String deepLinkJSON) {
             @Override
             public void onEvent(JSONObject jsonObject, JuspayResponseHandler juspayResponseHandler) {
                 Log.d(LOG_TAG, "onEvent: " + jsonObject.toString());
+                long initiateTimeStamp = System.currentTimeMillis();
```
ashishsingh101, line 485: initiate timestamp should be outside onEvent, it should be when we about to initiate hyperservice

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,30 @@ export const getImageUrl = function (url) {
...
 };
+
+export const getTimestampObject = function(string){
+  window.jsonObject[string] = Date.now() - window.prevTimeStamp;
+  window.prevTimeStamp = Date.now();
+  if (string == "baseAppFlow"){
+    window.tobeLog = true;}
+  else{
+    window.tobeLog = false;}
+}
```
ashishsingh101, line 470: add a check if key is already present otherwise it will override value

Frontend/ui-common/src/Engineering/Helpers/Commons.js
ashishsingh101: we should always deal with any edge cases in present that may cause issue in future, because if some code refactor happens then it might happen that it will not log accurate result

ashishsingh101 (changes requested):

Frontend/ui-driver/src/Flow.purs
ashishsingh101: @ashkriti how we will deal with cases if we are supposed to go to some other flow where this check is added and then to homescreen then both the timestamp is required, but only one will work out here

Frontend/ui-common/src/Engineering/Helpers/Commons.js
ashishsingh101, line 470: it should be independent of "baseAppFlow" check

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,30 @@ export const getImageUrl = function (url) {
     console.log("error in getImageUrl " + e);
   }
 };
+
+export const getTimestampObject = function(string){
+  window.jsonObject[string] = Date.now() - window.prevTimeStamp;
```
Vignesh-772, line 464: update the variable name to relevant name

Frontend/ui-driver/index.js
```diff
@@ -154,6 +157,16 @@ window.onMerchantEvent = function (_event, payload) {
     window.__payload.sdkVersion = "2.0.1"
     console.warn("Process called");
     const parsedPayload = JSON.parse(payload);
+    if (parsedPayload.payload && parsedPayload.payload.hasOwnProperty('onCreateTimeStamp') && parsedPayload.payload.hasOwnProperty('initiateTimeStamp'))
```
Vignesh-772, line 160: we don't need initiate called time ?

Frontend/ui-driver/index.js
```diff
@@ -154,6 +157,16 @@ window.onMerchantEvent = function (_event, payload) {
...
+      {
+        const onCreateTimeStamp = parsedPayload.payload.onCreateTimeStamp;
+        const initiateTimeStamp = parsedPayload.payload.initiateTimeStamp;
+        window.jsonObject["onCreateToBundle"] = prevTimeStamp - onCreateTimeStamp;
+        window.jsonObject["bundleToInitiate"] = initiateTimeStamp - window.prevTimeStamp;
```
Vignesh-772, line 165: this is `bundleToProcess` . the above if block is `bundleToInitiate`

Vignesh-772, line 165: `bundleToInitiate` has to be handled in js. `initiateTimeStamp - window.prevTimeStamp` this will give correct timings.

Frontend/ui-driver/src/Flow.purs
```diff
@@ -122,6 +122,7 @@ import Resource.Constants as RC
 
 baseAppFlow :: Boolean -> Maybe Event -> FlowBT String Unit
 baseAppFlow baseFlow event = do
+    _ <- pure $ getTimestampObject "baseAppFlow"
```
Vignesh-772, line 125: use `liftFlowBT` in all places instead of `pure`

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,30 @@ export const getImageUrl = function (url) {
...
+        value: jsonObject[key]
+      };
+    });
+    return keyValuePairArray;
+  }
+}
+
+export const getToBeLog = function(){
+  return function(){
+    return window.tobeLog;
```
Vignesh-772, line 486: Whats the use of `tobeLog` ?

Vignesh-772 (changes requested):

Frontend/ui-common/src/Engineering/Helpers/Commons.js
ashkriti: This checks if the timestamp has to be taken for that flow or not, It is based on if first screen is loaded and timestamp has been taken then it becomes false for that session

Vignesh-772 (approved):

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,24 @@ export const getImageUrl = function (url) {
     console.log("error in getImageUrl " + e);
   }
 };
+
+export const getTimestampObject = function(string){
```
ashishsingh101, line 463: can we please rename this to `setEventTimestamp` as per behaviour

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,24 @@ export const getImageUrl = function (url) {
...
+  return function(){
+    if (!window.flowTimeStampObject || !(string in window.flowTimeStampObject)){
+      window.flowTimeStampObject[string] = Date.now() - window.prevTimeStamp;
+      window.prevTimeStamp = Date.now();
+    }
+  }
+}
+
+export const getData = function(){
```
ashishsingh101, line 472: can we rename this to `getTimeStampObject` as per behaviour

Frontend/ui-driver/index.js
```diff
@@ -2,6 +2,8 @@ import "core-js";
 import "presto-ui";
```
ashishsingh101, line 2: can we please include bundle load time timestamp which is very much required, you can add a timeStamp in main becknbase.html which will be the timeStamp from `native initiate to bundle` time and in first line of `index.js` which will be the time for `bundle load`
flow will be like `native initiate -> becknbase.html -> index.js`

Frontend/ui-common/src/Engineering/Helpers/Commons.js
```diff
@@ -459,3 +459,24 @@ export const getImageUrl = function (url) {
     console.log("error in getImageUrl " + e);
   }
 };
+
+export const getTimestampObject = function(string){
+  return function(){
+    if (!window.flowTimeStampObject || !(string in window.flowTimeStampObject)){
+      window.flowTimeStampObject[string] = Date.now() - window.prevTimeStamp;
```
ashishsingh101, line 466: how will know the timestamp order ?

Frontend/ui-driver/src/Flow.purs
```diff
@@ -1803,6 +1834,16 @@ homeScreenFlow = do
...
+  let (GlobalState state) = globalState
+  when (state.globalProps.addTimestamp) $ do
+    liftFlowBT $ getTimestampObject "homeScreenFlow"
+    if state.homeScreen.props.tobeLogged then do
+      logData <- liftFlowBT $ getData unit
+      liftFlowBT $ logEventWithMultipleParams logField_ "sending_logs" logData
+      modifyScreenState $ HomeScreenStateType (\homeScreen -> homeScreen { props {tobeLogged = false}})
+      modifyScreenState $ GlobalPropsType $ \globalProps -> globalProps{addTimestamp = false}
+      pure unit
+      else pure unit
```
ashishsingh101, line 1846: in all these cases we are logging after screen render only right ?

ashishsingh101: can use `when` inside this body instead of `if-else`

ashishsingh101 (changes requested):

Frontend/ui-common/src/Engineering/Helpers/Commons.js
ashkriti: Currently inside the flow first we'll have the baseappflow and further other flows, so later if we want some other timestamps we can name them accordingly

Frontend/ui-driver/index.js
ashkriti: done

ashishsingh101 (dismissed):

witcher-shailesh (approved):

PR #4093, opened 2023-10-31
Backend/fix/compile times optimizations - Adds Parallel Jobs and Makes using development builds convenient 

arjunkathuria:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Major changes in this PR:- 

- Adds 6 parallel jobs by default when building the backend project.
- Adds a development section in cabal.project file with flags that can be simply be un-commented to get local optimized development builds that build very quickly, this makes using them (or turning them off) much more convenient, quick and low mental-overhead. 
 
### Additional Changes

Adds README sections documenting and explaining these features and elucidating usage.  

## Motivation and Context

1. We lacked parallel jobs setting. This also makes building more deterministic.
2. using the local developments builds and turning them on and off easily wasn't quite ergonomic. 

## How did you test it?

Built it locally and verified the settings work and the build times tally.

## Checklist

### DIFF
--- a/Backend/cabal.project
+++ b/Backend/cabal.project
@@ -36,3 +36,38 @@ packages:
   test
   lib/producer
 
+-- ## 6 parallel jobs by default.
+--
+-- # since we now use "-j4" in "ghc-options" for all packages,
+-- # allowing them to use up-to 4 cores each to build.
+-- # A number like 6 here means we can have 6 packages building parallel, with each allowed to use up-to 4 cores.
+-- # which means a maximum of 6 * 4 = 24 cores, which should be manageable in most cases.
+-- # A higher number of parallel jobs here could saturate the system easily.
+-- # This Upper-limits the max resources required while still giving us the maximum benefits of parallel builds,
+-- # without saturating the system completely.
+-- # This number should work for both laptops and desktops.
+--
+-- # you can increase this number if on a more powerful machine.
+jobs: 6
+
+-- ### DEVELOPMENT FLAGS ###
+-- Note: you may comment/un-comment these while doing local development for faster compile & build times
+
+-- Flag defaults configured as recommended here https://github.com/nammayatri/nammayatri/pull/4093
+
+-- ## The optimization level flag, range is between 0 to 2.
+-- # Level 0 disables optimization.
+-- # Level 1 is the default.
+-- # Level 2 is higher optimization if the compiler supports it.
+--
+-- # Note: Level 2 is likely to lead to longer compile times and bigger generated code.
+-- #       If you are not planning to run code,
+-- #       turning off optimization will lead to better build times
+-- #       and less code to be rebuilt when a module changes.
+
+-- optimization: 0
+
+-- ## Enables flags in the project.
+-- #  +flagname enables the flag.
+-- #  -flagname disables it.
+flags: +Local
--- a/Backend/README.md
+++ b/Backend/README.md
@@ -89,6 +89,16 @@ cabal run lib/location-updates
 , ghcid lib/location-updates
 ```
 
+#### Faster Local Development builds
+
+For much shorter compile times and quicker feedback cycles while developing, you may want to use the builds optimized for local development.
+
+This is now easily & quickly achieved by simply un-commenting the flags under ["DEVELOPMENT FLAGS" section in cabal.project file](cabal.project) when developing.
+
+
+#### Parallel Jobs
+To speed up the compilation times, we use 6 parallel jobs by default. If you have a powerful computer with lots of cores and memory, you can increment the `jobs` setting in [cabal.project](cabal.project) file to run more parallel jobs for faster results. Inversely, if its a low-powered machine, you may consider lowering that number.
+
 #### Running external services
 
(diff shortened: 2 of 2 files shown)


Backend/cabal.project
```diff
@@ -36,3 +36,12 @@ packages:
   test
   lib/producer
 
+-- 6 parallel jobs by default
+-- you can increase this number if on a more powerful CPU
+jobs: 6
```
srid, line 41: Does `cabal.project` not support [`$ncpus`](https://cabal.readthedocs.io/en/3.4/cabal-project.html#cfg-flag---jobs)?

Backend/cabal.project
```diff
@@ -36,3 +36,12 @@ packages:
...
+
+-- ### DEVELOPMENTENT FLAGS ###
+-- Note: you may uncomment these while doing local development for faster compile & build times
+
+-- optimization: 0
+-- flags: +Local
```
srid, line 47: Why not leave this uncommented by default? AFAICT, this won't affect Nix builds, anyway. haskell-flake uses cabal.project for `packages`; everything else is ignored.

Backend/cabal.project
arjunkathuria: it does, but we also now use `-j4` in `ghc-options` of all packages, which allows them to use upto 4 cores to build.

using a number like 6 here means we can have 6 packages building in parallel, with each allowed to use upto 4 cores which means a maximum of 6 * 4 = 24 cores at a time, which is a lot more manageable.

if we allow `$ncpus` and then also allow each of those to use upto 4 cores each, that would saturate the system really quickly.

This was a number that should work for both laptops and desktops. if someone has a more powerful machine they can then explicitly bump this if required, or decrement it in case of an under-powered machine as mentioned in the relevant README sections.

srid: Okay, you want to state this (especially the ghc-options -j4 part) in the comment on lines 39,40.

Backend/cabal.project
arjunkathuria: We can if we want to. 
I just deduced from the previous PR that introduced this flag and conditional logic that we would want these to be opt-in or explicitly set by the user.

srid: Let's first confirm why we actually need it to be conditional. @shreyaxh @hemantmangla ?

shreyaxh: We had conditional logic because we want optimized code when we are in prod, but in local our concern is faster build time only.

srid: Ok, so we can, in fact, leave this uncommented since prod builds (built via Nix) won't use `cabal.project` (non-packages) settings anyway.

srid: Aside: I do wonder why we are doing all of this, instead of simply disabling optimizations **by default** and then re-enabling them only in Nix.

Backend/cabal.project
arjunkathuria: Added the commentary.

Backend/cabal.project
arjunkathuria: LMK guys, to leave this on or off by default.
Resolved the other issues.

arjunkathuria:
Bump

arjunkathuria:
Hi guys, can we merge or close this? 

cc:- @shreyaxh @hemantmangla

srid (changes requested):
See https://github.com/nammayatri/nammayatri/pull/4093#discussion_r1377680234 - that's about the only change I see that can be done on this PR.

arjunkathuria:
> See [#4093 (comment)](https://github.com/nammayatri/nammayatri/pull/4093#discussion_r1377680234) - that's about the only change I see that can be done on this PR.

confirm please @hemantmangla @shreyaxh

shreyaxh:
> > See [#4093 (comment)](https://github.com/nammayatri/nammayatri/pull/4093#discussion_r1377680234) - that's about the only change I see that can be done on this PR.
> 
> confirm please @hemantmangla @shreyaxh
After using the build time optimization flags for a few days, we have identified an issue with the `scheduler` and `scheduler-example` take up almost all the disk space when built with `O0` flag, but this gets fixed when we build them with `O1`. So I will be raising another PR reverting them to `O1`. 

I believe we can't just use `optimization: 0` as a default. About the `flags: +Local`, as @srid mentioned that this won't affect Nix builds, so we can uncomment that.

arjunkathuria:
> This is why I believe we can't just use `optimization: 0` as a default. About the `flags: +Local`, as @srid mentioned that this won't affect Nix builds, so we can uncomment that.

if i got that correct:- 

1. leave `flags +Local` on/uncommented by default.
2. leave `optimization: 0` off/commented by default.

Would ^that work for you @srid ?
I have no opinions on the matter and would make the changes whichever way you guys request.

srid:
> Would ^that work for you @srid ?

👍🏿

arjunkathuria:
Updated.

arjunkathuria:
Bump

arjunkathuria:
Bump Again.

hkmangla (approved):

arjunkathuria:
Updated: rebased and squashed commits.

Issue #4150, opened 2023-11-04
Make DriverInfoCard as Maybe value


Vignesh-772:
**Describe**
Sometimes we are seeing the dummy data in Driver Info Card. The solution I'm approaching is even the data is not there but still polling is going in background where we are have the data. but it is not begin used. so we are going make use of that data. Here we can't update the data for each api call which is not a good solution. so we need to know when to update.As it is a object we can't check each keys. Its better to maintain it as Maybe Type. So when we call ride/list if we are not getting the data we can the data from polling and if driver info is nothing then we are updating the state.
While checking the issue found another case where if bookingId == "" then polling it self is not happening which can be fixed by calling ride/list and get the bookingID and use it exponentially. This will fix that even if are not getting the data but we are getting it in future at least screen updates after 1 or 2 sec. as the percentage of this issue is very less. Once bottom sheet revamp is done we can have shimmer until we get the data.

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**App (please complete the following information):**

**Additional context**

PR #4151, opened 2023-11-04
frontend/fix: #4150 Made Driver info card as Maybe type

Vignesh-772:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Sometimes we are seeing the dummy data in Driver Info Card. The solution I'm approaching is even the data is not there but still polling is going in background where we are have the data. but it is not begin used. so we are going make use of that data. Here we can't update the data for each api call which is not a good solution. so we need to know when to update.As it is a object we can't check each keys. Its better to maintain it as Maybe Type. So when we call `ride/list` if we are not getting the data we can the data from polling and if driver info is nothing then we are updating the state.
While checking the issue found another case where if bookingId == "" then polling it self is not happening which can be fixed by calling `ride/list` and get the bookingID and use it. This will fix that even if are not getting the data but we are getting it in future at least screen updates after 1 or 2 sec. as the percentage of this issue is  very less. Once bottom sheet revamp is done we can have shimmer until we get the data.

### Additional Changes

## Checklist


witcher-shailesh:
> Made Driver info card as Maybe type
but this we mean that info card itself will be hidden in case of Nothing?

witcher-shailesh:
- Is it the StateData which will be of maybe type? If yes we have to maintain a dummy value but that dummy values shouldn't be shown to users, to solve for this we might need shimmer in the first place. Any idea how will you solve it otherwise?
- Can we have a caching method also per session as this data won't be changing once ride is assigned other than the status. So we can cache the data and use it so that even if there's a gap of 2-3 seconds we can show some data, instead of blank/dummy.
- For polling or fallback to ride/list or other API let's have a exponential retry mechanism, so that in cases of failure it don't spam with multiple requests

Vignesh-772:
> * Is it the StateData which will be of maybe type? If yes we have to maintain a dummy value but that dummy values shouldn't be shown to users, to solve for this we might need shimmer in the first place. Any idea how will you solve it otherwise?

This will be the permanent solution. As we are doing UI revamp of the bottom sheet we can add shimmer after it is completed. Even we put shimmer still we have to maintain it as a maybe type. I'm thinking adding shimmer now is a not required


>     * Can we have a caching method also per session as this data won't be changing once ride is assigned other than the status. So we can cache the data and use it so that even if there's a gap of 2-3 seconds we can show some data, instead of blank/dummy.

Yes. We can cache the data per session. But the issue we are facing now is we are not getting the data when the session starts and we are getting it later in polling thread block. That's why we are updating the main thread block state once we got the data in polling thread block. and we need to know when to update the state also. If we want we can cache the whole response. we can cache in home screen state. The better solution for cache the data will be storing the response in local table via SQLite and fetch the data using bookingId. And this will be a shared resource for the any thread block. Comment your thoughts.

 
>     * For polling or fallback to ride/list or other API let's have a exponential retry mechanism, so that in cases of failure it don't spam with multiple requests

Sure. will implement this.

witcher-shailesh:
> This will be the permanent solution. As we are doing UI revamp of the bottom sheet we can add shimmer after it is completed. Even we put shimmer still we have to maintain it as a maybe type. I'm thinking adding shimmer now is a not required

Sure, let's roll @PraveenGongada also then for the visibility of these changes

> The better solution for cache the data will be storing the response in local table via SQLite

Yes, let's fast track the dhruv's PR for the same it will also help handling the offline cases

PR #4291, opened 2023-11-11
frontend/#2031/refactored code and moved translations to purs

ashishsingh101:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description

1. moved translation from JS to PURS to reduce zipped bundle size for :
    customer : 506.69 KB -> 471.88 KB
    driver : 588.69 KB -> 548.33 KB
    
    here we have to define the STR constructor for all the languages even for the one we define in configuration, same as 
    before getString will first lookup for translations with respective chosen language into common and then local if not able 
    to find the key.

2. refactored specialZone config and remove its direct dependency from JS.

### Additional Changes

## Checklist


Frontend/ui-customer/src/Resources/Localizable/Strings.purs
```diff
@@ -16,27 +16,27 @@
...
-getString key = getStringFromConfig $ trim $ getKeyString key
-
-getVarString :: STR -> Array String -> String
-getVarString key vals = getStringWithVar (trim $ getKeyString key) vals
-
-getEN :: STR -> String
-getEN key = getENStrings $ trim $ getKeyString key
\ No newline at end of file
+getString key = 
+  let language = getKeyInSharedPrefKeysConfig "LANGUAGE_KEY"
```
witcher-shailesh, line 33: `getString` function is called almost for every string we use in our screens. Assuming the frequency of this function, is it okay to call `SharedPref` everytime? Can we cache the language for each session in our state?

Frontend/ui-customer/src/Resources/Localizable/Strings.purs
```diff
@@ -16,27 +16,27 @@
...
+      stringValue = case (getStringFromConfig key Just Nothing) of
+                      Just value -> value
+                      Nothing    -> case language of
```
witcher-shailesh, line 36: this can be part of a separate function which do case match and return strings

Frontend/ui-driver/src/Resource/Localizable/Strings.purs
```diff
@@ -0,0 +1,48 @@
...
+import Resources.Localizable.BN (getBN)
+import Resources.Localizable.EN (getEN)
+import Resources.Localizable.HI (getHI)
+import Resources.Localizable.KN (getKN)
+import Resources.Localizable.ML (getML)
+import Resources.Localizable.TA (getTA)
+
+data Language = EN_US | KN_IN | HI_IN | TA_IN 
+
+getString :: STR -> String
```
witcher-shailesh, line 33: if this function/ file is same can it be under common library?

ashishsingh101: `Language` has no dependency so removed and `getString` has some differences in both customer and driver so it is required to be seperate

Frontend/ui-customer/src/Resources/Localizable/Strings.purs
ashishsingh101: yes we can cache per session but depending on state may not be optimal as we have to make sure state is updated all the time

ashishsingh101: we can store the value in `js` for that while updating native shared pref will update `js` also, so while fetching value we will call `js` and not native

witcher-shailesh: what about keeping it in GlobalProps? is it not safer to store in GlobalProps?

Frontend/ui-driver/src/Resource/Localizable/Strings.purs
witcher-shailesh: what's the difference in `getString` I tried finding out one but looks similar

ashishsingh101: `getString` is using `STR` which is different for both, also it is calling `getStringFromLocal ` which requires language string which is also different for both

witcher-shailesh: these functions can be passed as params to getString from driver or customer app independently

Issue #1361, opened 2023-06-22
Cleanup Stuck Rides/Bookings Flow


sidsethujuspay:
Create and Modify APIs as needed to facilitate the following flow of clearing stuck bookings and rides across BAP and BPP. The IDs of Rides/Bookings will be provided as input to the API and all APIs should support array (and single) input. If the operation fails for any one of the IDs the loop should continue for the rest of the IDs and the responses should be stored in a separate list to be displayed at the end.

We need to be able to do the following actions:
1. With BPP Ride IDs as input, `cancel` the rides by providing appropriate cancellation reason. Update the corresponding Booking in BPP and send both updates to BAP. If Ride doesn't exist in BAP, create a `cancelled` ride entry. **End Goal: BPP and BAP to be in sync.** 
2. With BPP Ride IDs as input, `end` the Rides by fetching the intended drop location from the Booking table and supplying it in the endRide API. Update the corresponding Booking in BPP and send the update to BAP. If Ride doesn't exist in BAP, create a `completed` ride entry.
3. With BPP Booking IDs as input, sync the status from its corresponding Ride in BPP. If no Ride exists, cancel the booking. In both cases, send update to BAP.
4. With BAP Ride IDs as input, call the Beckn `status` API and pull the Ride status from BPP, update it in BAP, update the corresponding booking.
5. With BAP Booking IDs as input, sync with corresponding BAP Ride, if no ride exists, then sync with BPP by calling status API such that both Booking and Ride reflect the status as in BPP.

We need these five Dashboard APIs. Review deletion of any unused APIs post this change.

PR #4322, opened 2023-11-16
backend/feat: #1361 on_status improving rebase

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Earlier `on_status` api used for sync only `booking` and ride `statuses`. After this PR we can do via `on_status`:

- update the same bap `ride` fields as we update via `on_update` (when bpp `ride.status` is not the same as bap `ride.status`)
- create bap `ride` if it is not there
- create `cancellation_reason` and `fare_breakup`s on bap side

### Additional Changes

## How did you test it?
Tested locally

## Checklist

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnStatus.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnStatus.hs
@@ -35,6 +35,7 @@ import Kernel.Types.Common hiding (id)
 import Kernel.Types.Error
 import Kernel.Types.Id (Id)
 import Kernel.Utils.Common
+import qualified Storage.CachedQueries.Merchant as CQM
 import qualified Storage.Queries.Booking as QB
 import qualified Storage.Queries.BookingCancellationReason as QBCR
 import qualified Storage.Queries.FareBreakup as QFareBreakup
@@ -92,6 +93,10 @@ data RideDetails
       { mbNewRideInfo :: Maybe NewRideInfo,
         cancellationSource :: DBCR.CancellationSource
       }
+  | BookingReallocationDetails
+      { newRideInfo :: NewRideInfo,
+        reallocationSource :: DBCR.CancellationSource
+      }
 
 -- the same as OnUpdateFareBreakup
 data OnStatusFareBreakup = OnStatusFareBreakup
@@ -113,7 +118,8 @@ buildRideEntity booking updRide newRideInfo = do
   mbExistingRide <- B.runInReplica $ QRide.findByBPPRideId newRideInfo.bppRideId
   case mbExistingRide of
     Nothing -> do
-      newRide <- buildNewRide booking newRideInfo
+      mbMerchant <- CQM.findById booking.merchantId
+      newRide <- buildNewRide mbMerchant booking newRideInfo
       pure $ RenewedRide (updRide newRide)
     Just existingRide -> do
       unless (existingRide.bookingId == booking.id) $ throwError (InvalidRequest "Invalid rideId")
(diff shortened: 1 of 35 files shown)


Backend/dev/rest-client/dynamic-offer-driver-app/auto-flow.http
```diff
@@ -1,6 +1,11 @@
 @token1 = {{driver-offer-bpp-auto-token1}}
```
khuzema786, line 1: @roman-bodavskiy are this changes required to be pushed ?

Backend/dev/migrations/mock-registry/0000-init.sql
```diff
@@ -0,0 +1,2 @@
+-- duplicated from seed, app is not running without migration
+ALTER TABLE atlas_registry.subscriber OWNER TO atlas_registry_user;
```
khuzema786, line 2: This also is it required to be pushed ?

Backend/dev/rest-client/dynamic-offer-driver-app/auto-flow.http
roman-bodavskiy: this changes affect only local testing with rest client vs code extension, I was not able to test without them, so can be pushed I think

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnStatus.hs
```diff
@@ -11,42 +11,256 @@
...
+                  rideStartTime = Just rideStartedInfo.rideStartTime,
+                  driverArrivalTime = rideStartedInfo.driverArrivalTime
+                 }
+    RideCompletedDetails {newRideInfo, rideStartedInfo, rideCompletedInfo} -> do
+      rideEntity <- buildRideEntity booking updateRideCompleted newRideInfo
+      rideBookingTransaction bookingNewStatus rideNewStatus booking rideEntity
+      when (isStatusChanged booking.status bookingNewStatus rideEntity) $ do
+        breakups <- traverse (buildFareBreakup booking.id) rideCompletedInfo.fareBreakups
+        QFareBreakup.deleteAllByBookingId booking.id
+        QFareBreakup.createMany breakups
```
khuzema786, line 179: Also here we are deleting FareBreakups by booking.id and then creating new ones. Can we just update the existing fare breakups in this case instead of deleting and creating new ones.

Backend/dev/migrations/mock-registry/0000-init.sql
roman-bodavskiy: This is also required for local testing. Mock-registry failed to run locally because of empty directory `dev/migrations/mock-registry.` 
If it bothers us, I can remove

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnStatus.hs
roman-bodavskiy: I was thinking about that. then we need to update each breakup by description (id generated on bap side, did not come from bpp). What if changed number of breakups, or descriptions? logic will be too complicated

Anyway this logic works when status changed, so most likely we will not delete anything, only create

khuzema786: Sure @roman-bodavskiy Works

Backend/dev/migrations/mock-registry/0000-init.sql
khuzema786: Ohh yes, someone faced this issue too, we just created a directory to fix it. Sure we can have this local fixes to solve such issues

khuzema786 (approved):

PR #4586, opened 2023-12-04
frontend/fix: 3525-disability banner storage updation issue fix

MercyQueen:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

Issue - The banner to update the disability is showing up on Home Screen when user kills and reopens the application even if the value is updated. This is because of incorrect value updation in the flow.


kavya-shree-s:
@MercyQueen Can you add description what is the issue ?

kavya-shree-s (approved):

Issue #1169, opened 2023-06-08
Adding Call logs to Kafka


PraveenGongada:
**Is your feature request related to a problem? Please describe.**
To check the status of the exotel calls

**Describe the solution you'd like**
Adding the exotel call data to kafka and check for the serviceability of the exotel

PR #4784, opened 2023-12-15
backend/feat:#1169 Made Ride id maybe in call event

PraveenGongada:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Made rideId maybe in call event

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/CallEvent.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/CallEvent.hs
@@ -30,7 +30,6 @@ import Kernel.Types.Id
 import Kernel.Utils.Common
 import Lib.SessionizerMetrics.Types.Event
 import qualified Storage.Queries.Ride as QRide
-import Tools.Error
 import Tools.Event
 
 data CallEventReq = CallEventReq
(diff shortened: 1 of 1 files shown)


0utkarsh (approved):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/CallEvent.hs
```diff
@@ -48,6 +47,8 @@ sendCallDataToKafka :: (EncFlow m r, CacheFlow m r, EsqDBFlow m r, EsqDBReplicaF
 sendCallDataToKafka vendor mRideId callType callSid callStatus triggeredBy exophoneNumber = do
   case mRideId of
     Just rideId -> do
-      ride <- runInReplica $ QRide.findById rideId >>= fromMaybeM (RideDoesNotExist rideId.getId)
-      triggerExophoneEvent $ ExophoneEventData vendor callType (Just rideId) callSid callStatus ride.merchantId triggeredBy (Just ride.driverId) exophoneNumber
+      mRide <- runInReplica $ QRide.findById rideId
+      case mRide of
+        Just ride -> triggerExophoneEvent $ ExophoneEventData vendor callType (Just rideId) callSid callStatus ride.merchantId triggeredBy (Just ride.driverId) exophoneNumber
+        Nothing -> triggerExophoneEvent $ ExophoneEventData vendor callType (Just rideId) callSid callStatus Nothing triggeredBy Nothing exophoneNumber
```
RitikaHanish, line 53: Is it  suppose to be like this ?
Don't you need to mention that it is RC related call ?

PraveenGongada: We are checking if the callService is active or not using this right? won't it same for RC calls too?

PR #5100, opened 2024-01-18
backend: Refactor `run-mobility-stack.nix` for composability

srid:
This file is too complex to manage. Refactor this so that we can compose it with the load test process group.

uses https://github.com/Platonic-Systems/process-compose-flake/pull/55


srid:
Tested by @soumyajit2304

PR #5140, opened 2024-01-22
Add script to kill processes occupying external services’ port

shivaraj-bh:
This script will be useful when `process-compose` doesn’t gracefully shutdown, leaving services like postgres, kafka, nginx etc. occupying the ports which are required for `, run-svc` to start.


srid:
On macOS, I get:

image

It is true that this package doesn't have `kill`:

```
❯ ls /nix/store/9k37gqn2r05i1sv8wz8xhjh5sq7yafyj-procps-1003.1-2008/bin/
 ps   sysctl   top   watch
```

`kill` appears to be a shell-builtin that you can just use. Using the builtin, 

```diff
diff --git a/Backend/nix/scripts.nix b/Backend/nix/scripts.nix
index 5bd6510486..491673d34e 100644
--- a/Backend/nix/scripts.nix
+++ b/Backend/nix/scripts.nix
@@ -101,7 +101,7 @@ _:
               if [ -n "$pid" ]; then
                 echo "Sending SIGTERM to processs running on ${builtins.toString port}:"
                 echo "$pid"
-                echo "$pid" | ${pkgs.findutils}/bin/xargs ${pkgs.procps}/bin/kill
+                echo "$pid" | ${pkgs.findutils}/bin/xargs kill
               else
                 echo "No processes found on port ${builtins.toString port}"
               fi
```

it works, and this is what I see:

```
❯ , kill-svc-ports
Sending SIGTERM to processs running on 5434:
37559
No processes found on port 5435
No processes found on port 29092
Sending SIGTERM to processs running on 8080:
4600
No processes found on port 5422
No processes found on port 8021
No processes found on port 6379
No processes found on port 30001
No processes found on port 30002
No processes found on port 30003
No processes found on port 30004
No processes found on port 30005
No processes found on port 30006
No processes found on port 2181
```

When I ran it again, it tried to send SIGTERM to 37559 (so it wasn't killed the last time).

Running it the third time, it still sends SIGTERM to 37559.

srid:
Last lines from rider-app-exe.log:

```
{"level":"info","process":"rider-app-exe","replica":0,"message":"Shutting down by sigTERM"}
{"level":"info","process":"rider-app-exe","replica":0,"message":"Disposing async logger..."}
```

It seems to be stuck after this point.

shivaraj-bh:
Considering this script will be used in extreme cases, I would think it will be okay to use `kill -9` (SIGKILL)?

srid:
Yes.

shivaraj-bh:
> When I ran it again, it tried to send SIGTERM to 37559 (so it wasn't killed the last time).

rider-app should not be a process that is occupying port `5432`, I should probably just kill the process that has `(LISTEN)
` in the `NAME` column output of `lsof`, which will be the `db-primary` (postgres) process.

shivaraj-bh:
> rider-app should not be a process that is occupying port 5432, I should probably just kill the process that has (LISTEN)  in the NAME column output of lsof, which will be the db-primary (postgres) process.

 I don’t think I will have to bother about this now, in a case where this command has to be used it doesn’t matter if it kills all the processes associated with the port (whether it is listening or not). I think it looks good to merge.

PR #5261, opened 2024-01-31
backend/feat/namma-dsl/Generate code only if spec file hash changed

ErasedSoul:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

**, run-generator** : Will only generate specs which are changed or new spec file. This is done by getting the current hash of the spec file and comparing with the file hash of the latest HEAD commit.

**, run-generator --all** : Will generate all specs. 

**See Motivation and Context for detailed reasons why this is required.**
                             

### Additional Changes

## Motivation and Context

This change is required due to following reasons: 

- **Developers don't have to see other project's generated code. Only his/her changed spec will be generated.**
- **Sometime it we don't want to generate the code untill a feature of dsl is pending. (For example monadic transformers WIP)**
- **With this after a spec is commited and not changed after the code will not be generated for the spec, thus reducing the chance of breaking other's code**
- **After all to sync with dsl we can use the --all args with , run-generator command to generate all.**

## Checklist

### DIFF
--- a/Backend/app/alchemist/src/Main.hs
+++ b/Backend/app/alchemist/src/Main.hs
@@ -5,6 +5,7 @@ import qualified Data.Yaml as Yaml
 import Kernel.Prelude
 import qualified NammaDSL.App as NammaDSL
 import System.Directory
+import System.Environment (getArgs)
 import System.FilePath
 
 findGitRoot :: FilePath -> IO (Maybe FilePath)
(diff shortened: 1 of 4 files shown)


Backend/app/alchemist/src/Main.hs
```diff
@@ -46,36 +50,73 @@ applyDirectory dirPath processFile = do
...
-      NammaDSL.mkDomainHandler (src </> "Domain/Action/UI") inputFile
+    processApp :: Bool -> String -> FilePath -> FilePath -> FilePath -> IO ()
+    processApp isGenAll dbName rootDir appPath appName = do
+      applyDirectory (rootDir </> appPath </> "spec" </> "Storage") (processStorageDSL isGenAll dbName rootDir appPath appName)
+      applyDirectory (rootDir </> appPath </> "spec" </> "API") (processAPIDSL isGenAll rootDir appPath)
+
+    processStorageDSL isGenAll dbName' rootDir appPath appName inputFile = do
+      fileState <- getFileState inputFile
+      putStrLn $ show fileState ++ " " ++ inputFile
+      when (isGenAll || fileState == NEW || fileState == CHANGED) $ do
```
hkmangla, line 102: What about fileState doesn't exist?

hkmangla (changes requested):

Backend/app/alchemist/src/Main.hs
ErasedSoul: Does not matter as it will take provide only the existing files using this fuction : 
applyDirectory :: FilePath -> (FilePath -> IO ()) -> IO ()
applyDirectory dirPath processFile = do
  exists <- doesDirectoryExist dirPath
  when exists $ do
    files <- listDirectory dirPath.  <<<<<<<<<< This one 
    let yamlFiles = filter (\file -> takeExtension file `elem` [".yaml", ".yml"]) files
    mapM_ (processFile . (dirPath </>)) yamlFiles

hkmangla (dismissed):

hkmangla (approved):

PR #5520, opened 2024-02-14
backend/refactor: remove partial fields

roman-bodavskiy:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
Remove partial fields

### Additional Changes

## Motivation and Context
suppose we have type
`data Foo = Bar {bar :: Int, cat :: Int} | Baz {baz :: Text}`
then `bar` will be partial field, i.e.

- if record created with `Bar` constructor, then all is okay;

- if record created with `Baz` constructor, and then somewhere in code we try to reach this field, we get error in runtime, which is not appear in logs and hard to debug:

```
let foo = Baz {baz = "Baz"}
...
foo.bar -- error in runtime
```
So I think we should remove this partial fields from our code, adding extra layer of definitions:

```
data Foo = Bar Bar' | Baz Baz'
data Bar' = Bar' {bar :: Int, cat :: Int} 
newtype Baz' = Baz' {baz :: Text}

let foo = Baz $ Baz' {baz = "Baz"}
...
case foo of
  Bar bar -> Just bar.bar
  Baz baz -> Nothing
```
With this approach unsafe code can't be compiled
 

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Update.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Update.hs
@@ -11,8 +11,6 @@
 
  the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
 -}
-{-# OPTIONS_GHC -Wwarn=incomplete-record-updates #-}
-{-# OPTIONS_GHC -Wwarn=incomplete-uni-patterns #-}
 
 module Domain.Action.Beckn.Update where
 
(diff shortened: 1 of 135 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Update.hs
```diff
@@ -42,32 +40,46 @@ import Tools.Error
...
+getBookingId (UAddStopReq req) = req.bookingId
+getBookingId (UEditStopReq req) = req.bookingId
 
 data PaymentStatus = PAID | NOT_PAID
 
 handler :: DUpdateReq -> Flow ()
-handler req@PaymentCompletedReq {} = do
-  unless (req.paymentMethodInfo.paymentType == DMPM.ON_FULFILLMENT) $
+handler (UPaymentCompletedReq req@PaymentCompletedReq {}) = do
+  unless (req.paymentMethodInfo.paymentType == DMPM.POSTPAID) $
```
hkmangla, line 82: It needs to be `ON_FULFILLMENT`. Lets remove POSTPAID, it has been removed from production.

hkmangla (changes requested):

hkmangla (approved):

hkmangla (approved):

PR #5741, opened 2024-02-27
backend/feat : added delete api for person from dashboard

ashishsaini25:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

added delete api for person from dashboard
### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
+++ b/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
@@ -105,6 +105,11 @@ newtype ReleaseRegisterReq = ReleaseRegisterReq
   {token :: Text}
   deriving (Show, Generic, FromJSON, ToJSON, ToSchema)
 
+newtype ChangeEnabledStatusReq = ChangeEnabledStatusReq
+  { enabled :: Bool
+  }
+  deriving (Generic, ToJSON, FromJSON, ToSchema)
+
 data ReleaseRegisterRes = ReleaseRegisterRes
   { username :: Text,
     token :: Text,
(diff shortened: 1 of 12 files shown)


Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
```diff
@@ -418,6 +418,22 @@ changeEmailByAdmin _ personId req = do
...
+deletePerson ::
+  ( BeamFlow m r,
+    Redis.HedisFlow m r,
+    HasFlowEnv m r '["authTokenCacheKeyPrefix" ::: Text]
+  ) =>
+  TokenInfo ->
+  Id DP.Person ->
+  m APISuccess
+deletePerson _ personId = do
+  void $ QP.findById personId >>= fromMaybeM (PersonNotFound personId.getId)
```
0utkarsh, line 430: We can use replica

0utkarsh (approved):
Approving as code wise everything is correct. Can we implement soft delete in these cases(when we are not forced to delete according to play store compliance etc in case of drivers/customers)?

ashishsaini25:
> Approving as code wise everything is correct. Can we implement soft delete in these cases(when we are not forced to delete according to play store compliance etc in case of drivers/customers)?

yes , i am thinking to add enable and disable functionality.

ashishsaini25:
> > yes , i am thinking to add enable and disable functionality.
> 
> Sure, this can be used if enable-disable will go back and forth. If we want to soft delete, we can just add a column like `deleted_at` and while listing users or performing anything, we can keep a check on this field being not null

sure that can also be the way , but currently i have added enabled in registration token where it will handle this case.

vijaygupta18 (approved):

PR #5815, opened 2024-03-04
backend/ONDC-added track and ontrack changes

RitikaHanish:
## Type of Change
backend/ONDC-added track and ontrack changes

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
backend/ONDC-added track and ontrack changes

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnTrack.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnTrack.hs
@@ -15,36 +15,68 @@
 module Domain.Action.Beckn.OnTrack
   ( onTrack,
     OnTrackReq (..),
+    TrackingLocation (..),
     validateRequest,
   )
 where
 
+import qualified Beckn.OnDemand.Utils.Common as Common
 import Domain.Types.Ride
 import EulerHS.Prelude hiding (id)
+import Kernel.Beam.Functions as B
+import Kernel.Storage.Esqueleto.Config (EsqDBReplicaFlow)
+import qualified Kernel.Storage.Hedis as Hedis
+import qualified Kernel.Types.Beckn.Gps as Gps
 import Kernel.Types.Common hiding (id)
 import Kernel.Types.Id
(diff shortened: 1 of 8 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnTrack.hs
```diff
@@ -15,36 +15,68 @@
...
-    ValidatedOnTrackReq
-      { ..
-      }
+  booking <- B.runInReplica $ QRB.findById ride.bookingId >>= fromMaybeM (BookingDoesNotExist ride.bookingId.getId)
+  isValueAddNP <- CQVAN.isValueAddNP booking.providerId
+  if isValueAddNP && isNothing trackUrl
+    then throwError $ InvalidRequest "TrackUrl is required for ValueAddNP"
+    else
+      if not isValueAddNP && isNothing trackingLocation
+        then throwError $ InvalidRequest "TrackingLocation is required for non ValueAddNP"
```
NikithShetty, line 77: Update the error message to `TrackingLocation is required` since this error will be transmitted over the network

Backend/lib/shared-services/package.yaml
```diff
@@ -1,3 +1,8 @@
+# hpack defaults
```
NikithShetty, line 1: skip this diff?

NikithShetty (changes requested):

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnTrack.hs
RitikaHanish: @NikithShetty We need to do error handling anyway as per spec. it can be taken with that ?

Backend/lib/shared-services/package.yaml
RitikaHanish: This is generated by pre-commit hook, it will keep on coming every time.

PR #5943, opened 2024-03-13
backend/nix: Avoid unnecessary rebuilds

srid:
With this change, when non-Haskell-related files change, the Haskell nix packages won't be rebuilt.

cf. https://github.com/srid/haskell-flake/issues/269


srid:
n/m - this is already the case for packages in subdir.

PR #6102, opened 2024-03-26
backend/fix: race condition of multiple booking creation

jaypalmudaliyar24:
## Type of Change

- [x] Bugfix

## Description

Before creating a booking in `on_select`, after above change we will first try to acquire lock on the basis of `personId`, to avoid multiple booking creation for same person.

## Motivation and Context

Fixes race condition of multiple booking creation for same person at same time when consecutive `on_select` callbacks are received. 

## How did you test it?

Tested locally.

## Checklist

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnSelect.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnSelect.hs
@@ -110,9 +110,9 @@ onSelect OnSelectValidatedReq {..} = do
       let lowestFareQuote = selectLowestFareQuote quotes
       case lowestFareQuote of
         Just autoAssignQuote -> do
-          isLockAcquired <- SConfirm.tryInitTriggerLock person.id
+          isLockAcquired <- SConfirm.tryInitTriggerLock autoAssignQuote.requestId
           when isLockAcquired $ do
-            let dConfirmReq = SConfirm.DConfirmReq {personId = person.id, quoteId = autoAssignQuote.id, paymentMethodId = searchRequest.selectedPaymentMethodId}
+            let dConfirmReq = SConfirm.DConfirmReq {personId = person.id, quote = autoAssignQuote, paymentMethodId = searchRequest.selectedPaymentMethodId}
             dConfirmRes <- SConfirm.confirm dConfirmReq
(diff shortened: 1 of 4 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnSelect.hs
```diff
@@ -110,11 +110,13 @@ onSelect OnSelectValidatedReq {..} = do
       let lowestFareQuote = selectLowestFareQuote quotes
       case lowestFareQuote of
         Just autoAssignQuote -> do
-          let dConfirmReq = SConfirm.DConfirmReq {personId = person.id, quoteId = autoAssignQuote.id, paymentMethodId = searchRequest.selectedPaymentMethodId}
-          dConfirmRes <- SConfirm.confirm dConfirmReq
-          becknInitReq <- ACL.buildInitReqV2 dConfirmRes
-          handle (errHandler dConfirmRes.booking) $
-            void . withShortRetry $ CallBPP.initV2 dConfirmRes.providerUrl becknInitReq
+          isLockAcquired <- SConfirm.tryInitTriggerLock person.id
```
hkmangla, line 113: Why both locks are `tryInitTriggerLock`?

hkmangla (changes requested):

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnSelect.hs
hkmangla: And lock shouldn't be on person.id, it should be on search request id.

jaypalmudaliyar24: > Why both locks are tryInitTriggerLock?
as we are trying to acquire lock thats why try is used, in case we don't get lock we need to throwError in autoAssign false case (for UI API [here](https://github.com/nammayatri/nammayatri/pull/6102/files#:~:text=isLockAcquired%20%3C%2D%20SConfirm.tryInitTriggerLock%20personId))
and wanted to keep same locking function for both.

> And lock shouldn't be on person.id, it should be on search request id.
fixed.

hkmangla (approved):

PR #6511, opened 2024-04-22
Fixed drivers Lat Lon not coming for other elements other than head i…

MercyQueen:


Backend/app/rider-platform/rider-app/Main/src/Beckn/OnDemand/Utils/OnSearch.hs
```diff
@@ -251,11 +251,36 @@ makeLatLong provider vehicleVariant location = do
...
+            >>= ( \itemfullfillment ->
+                    provider.providerFulfillments
+                      >>= ( \provFul ->
+                              Just (filterFulfillmentsByFulfillmentId provFul itemfullfillment)
+                          )
+                )
+        )
+
+    filterFulfillmentsByFulfillmentId providerFulfillments arrFullFIllment = do
+      let result = filter (\fulf -> listToMaybe arrFullFIllment == fulf.fulfillmentId) providerFulfillments
```
jaypalmudaliyar24, line 272: This will always return single fulfilment, 
because of this check `listToMaybe arrFullFIllment == fulf.fulfillmentId`
so instead of `filter`, you can use `find` here. 
This will also make return type as `Maybe VehicleVariant` and solve your issue of list of lists.

jaypalmudaliyar24 (changes requested):

jaypalmudaliyar24 (dismissed):

jaypalmudaliyar24 (approved):

PR #6548, opened 2024-04-25
backend/nix: Update common (haskell-flake)

srid:
- [x] Update haskell-flake
- [ ] Simplify `backend/default.nix` correspondingly


srid:
> * Simplify `backend/default.nix` correspondingly

Will do this in separate PR.

PR #6840, opened 2024-05-12
frontend/fix- try again issue fix

MercyQueen:
Fixes the following : 
- Local Stage Updation issue when flow goes inside retry_finding_quotes.


Frontend/ui-customer/src/Components/QuoteListModel/Controller.purs
```diff
@@ -60,4 +60,6 @@ type QuoteListModelState = {
   , selectedEstimatesObject :: CVC.Config
   , showAnim :: Boolean
   , animEndTime :: Int
+  , currentStage :: Stage
```
ashishsingh101, line 63: not required, is it added on purpose ?

MercyQueen: was added on purpose only, not required anymore. I'll remove it.

witcher-shailesh (approved):

PR #7097, opened 2024-05-30
backend/fix/validate-Past-booking-date-for-ticket-booking-service

max-keviv:
1
added order status api call if any status is in pending refund to update the status
2
Earlier-
there was no check for past ticket booking request so past tickets could also be booked
now -
added the check if current time is greater than start time (visit date and start time of business hour) throw error

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/TicketService.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/TicketService.hs
@@ -293,6 +293,14 @@ postTicketPlacesBook (mbPersonId, merchantId) placeId req = do
       now <- getCurrentTime
       ticketService <- QTS.findById ticketServicesReq.serviceId >>= fromMaybeM (TicketServiceNotFound ticketServicesReq.serviceId.getId)
       businessHour <- QBH.findById bHourId >>= fromMaybeM (BusinessHourNotFound bHourId.getId)
+
+      let startTime = case businessHour.btype of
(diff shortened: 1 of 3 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/TicketService.hs
```diff
@@ -293,6 +293,14 @@ postTicketPlacesBook (mbPersonId, merchantId) placeId req = do
       now <- getCurrentTime
       ticketService <- QTS.findById ticketServicesReq.serviceId >>= fromMaybeM (TicketServiceNotFound ticketServicesReq.serviceId.getId)
       businessHour <- QBH.findById bHourId >>= fromMaybeM (BusinessHourNotFound bHourId.getId)
+
+      let startTime = case businessHour.btype of
+            Domain.Types.BusinessHour.Slot time -> time
+            Domain.Types.BusinessHour.Duration startTime' _ -> startTime'
+      let visitDateTime = UTCTime visitDate (timeOfDayToTime startTime)
+      
+      when (visitDateTime < now) $ throwError $ InvalidRequest "Cannot book for past date"
```
0utkarsh, line 302: Can we allow till some time of start of the show? For example, when we are late for movies, we get tickets from counter but not online, we can allow online bookings too for some period(10 mis let's say?). Any issue with that @sidsethu?

0utkarsh (approved):
Can be merged after build passes

0utkarsh (approved):

PR #7833, opened 2024-07-13
frontend/fix - source pointer locate on map issue

MercyQueen:
fixes the following issue: 
- chat not working properly after ride is started
- wait time not visible if user tries to edit destination
- Source pointer not getting updated on source change in few cases


aryan-13 (approved):

Frontend/ui-customer/src/Flow.purs
```diff
@@ -1027,7 +1027,7 @@ homeScreenFlow = do
                     , nearByPickUpPoints = pickUpPoints
                     }
                   , props
-                    { defaultPickUpPoint = ""
+                    { defaultPickUpPoint = (fromMaybe HomeScreenData.dummyLocation (pickUpPoints !! 0)).place
```
aryan-13, line 1030: We can make this value nothing instead of using dummy value

ashishsingh101 (approved):

PR #7920, opened 2024-07-18
frontend/api optimisation customer side

RitikaHanish:
## Type of Change
frontend/api optimisation customer side

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## Description
frontend/api optimisation customer side

### DIFF
--- a/Frontend/ui-customer/src/Flow.purs
+++ b/Frontend/ui-customer/src/Flow.purs
@@ -1417,7 +1417,7 @@ homeScreenFlow = do
           homeScreenFlow
     FCM_NOTIFICATION notification state -> fcmHandler notification state
     LOGOUT -> do
-      (LogOutRes resp) <- Remote.logOutBT LogOutReq
+      (APISuccessResp resp) <- Remote.logOutBT LogOutReq
       removeChatService ""
       void $ pure $ deleteValueFromLocalStore REGISTERATION_TOKEN
       void $ pure $ deleteValueFromLocalStore REGISTRATION_APPROVED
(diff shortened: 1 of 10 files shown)


Frontend/ui-customer/src/Flow.purs
```diff
@@ -1435,8 +1435,12 @@ homeScreenFlow = do
       enterMobileNumberScreenFlow -- Removed choose langauge screen
     SUBMIT_RATING state -> do
       liftFlowBT $ logEventWithMultipleParams logField_ "ny_user_ride_give_feedback" $ [ { key: "Rating", value: unsafeToForeign state.data.ratingViewState.selectedRating } ]
-      void $ lift $ lift $ fork $ HelpersAPI.callApi $ Remote.makeRideFeedBackReq state.data.rideRatingState.rideId state.data.rideRatingState.feedbackList
-      void $ lift $ lift $ fork $ HelpersAPI.callApi $ getfeedbackReq state
+      void $ lift $ lift $ fork do
+        (_ :: (Either ErrorResponse RideFeedbackRes)) <- HelpersAPI.callApi $ Remote.makeRideFeedBackReq state.data.rideRatingState.rideId state.data.rideRatingState.feedbackList
```
MercyQueen, line 1439: @RitikaHanish  can't we do void here instead?

MercyQueen (commented):
Added comment

Frontend/ui-customer/src/Flow.purs
RitikaHanish: No, Actually we need to typecast, as in we need to explicitly tell that what response type are we expecting, so that it can decode it into that.

MercyQueen: ohh got it. 👍🏽

MercyQueen (approved):

PR #8941, opened 2024-10-08
frontend/feat: Added SUV Plus Images

PraveenGongada:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Added SUV Plus Variant Images


slow-codex (changes requested):
@PraveenGongada Can we update the FF_ASSET usage as well if we are moving the images to common because this will fail in bundle update. (example - [here](https://github.com/nammayatri/nammayatri/blob/64f777196a8cb29e05677f8a1e1651dafad96468/Frontend/ui-customer/src/Helpers/Utils.purs#L715))

slow-codex (approved):

Vignesh-772 (approved):

PR #9154, opened 2024-10-23
backend/fix: refactored frfsStation and frfsAutoComplete api

Radhika1708:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/spec/API/FrfsTicket.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/API/FrfsTicket.yaml
@@ -195,6 +195,8 @@ apis:
         city: City
         routeCode: Text
         startStationCode: Text
+        endStationCode: Text
+        location: LatLong
       response:
         type: "[API.Types.UI.FRFSTicketService.FRFSStationAPI]"
 
@@ -321,5 +323,7 @@ apis:
         vehicleType: VehicleCategory
       query:
         input: Text
+        limit: Int
+        offset: Int
       response:
         type: "API.Types.UI.FRFSTicketService.AutocompleteRes"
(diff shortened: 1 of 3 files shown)


Backend/app/rider-platform/rider-app/Main/spec/API/FrfsTicket.yaml
```diff
@@ -191,10 +191,12 @@ apis:
       auth: TokenAuth
       mandatoryQuery:
         vehicleType: VehicleCategory
+        location: LatLong
```
khuzema786, line 194: It should be optional @Radhika1708

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FRFSTicketService.hs
```diff
@@ -834,29 +942,21 @@ getFrfsConfig (pId, mId) opCity = do
...
   Kernel.Prelude.Maybe Text ->
+  Kernel.Prelude.Maybe Int ->
+  Kernel.Prelude.Maybe Int ->
   Context.City ->
   Kernel.External.Maps.Types.LatLong ->
   BecknV2.FRFS.Enums.VehicleCategory ->
   Environment.Flow API.Types.UI.FRFSTicketService.AutocompleteRes
-getFrfsAutocomplete (_, mId) mbInput opCity origin vehicle = do
+getFrfsAutocomplete (_, mId) mbInput mbLimit mbOffset opCity origin vehicle = do
+  let limit = toInteger $ fromMaybe 10 mbLimit
```
khuzema786, line 958: Dont do fromMaybe, directly pass this Maybe value to the Queries

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FRFSTicketService.hs
```diff
@@ -927,29 +1022,63 @@ getFrfsAutocomplete (_, mId) mbInput opCity origin vehicle = do
...
-                      lon = Just stopResp.origin.lon,
-                      distance = Just stopResp.distance,
-                      stationType = Nothing,
-                      sequenceNum = Nothing,
-                      address = Nothing,
-                      color = Nothing,
-                      towards = Nothing
-                    }
+          results <-
+            mapM
```
khuzema786, line 1026: @Radhika1708 Here we are calling OSRM for each stop, that is not the ideal approach as many stops will be there and if each OSRM network call takes 100 ms then for 5000 stops in worst case it would take 50 seconds for the response, which is not optimal. 
Instead, here we will call OSRM in single call and if it fails then ignore the distance for now. Meanwhile lets try to benchmark OSRM locally with how many max number of points can it take at once, then we will call in that many batches.

khuzema786 (approved):

PR #9493, opened 2024-11-28
Reintroduce omnix in direnv

srid:
https://omnix.page/om/develop.html - automated a bunch of things on top of nix devshell (like using cachix cache automatically, and doing health checks)

We had added this before but it was slower. Now, this should now be faster. 

Expect:

image


srid:
As an aside, I just go this:

image

While evaluating `cabal.executables`, so this is to do with our gigantic Haskell modules.

We need to automate this workaround, I think: https://github.com/nix-community/robotnix/issues/224#issuecomment-2073098186

PR #9852, opened 2025-01-11
backend/ci/enh/add-arm-linux-builds-to-ci

vsaimanohar:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [X] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **Infrastructure**
	- Enhanced GitHub Actions workflow to support additional build architectures.
	- Added support for `aarch64-linux` build target.
	- Refined conditional logic for job execution in the CI process.

- **Build Configuration**
	- Updated Nix configuration to provide architecture-specific optimization settings.
	- Introduced conditional build optimization flags for `aarch64-linux` architecture.

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -43,15 +43,28 @@ jobs:
           fi
           echo "has_backend=${has_backend}" >> $GITHUB_OUTPUT
           echo "Label check result: ${has_backend}"
+
   build:
     strategy:
       matrix:
-        system: [x86_64-linux, aarch64-darwin]
+        include:
+          # Direct builds
+          - system: x86_64-linux
+            host: x86_64-linux
+            run-condition: true
+          - system: aarch64-darwin
+            host: aarch64-darwin
+            run-condition: true
+          # Cross compilation
+          - system: aarch64-linux
+            host: aarch64-darwin
+            run-condition: github.ref == 'refs/heads/main'
     needs: [process-labels]
-    runs-on: ${{ matrix.system }}
+    runs-on: ${{ matrix.host }}
     if: |
       needs.process-labels.outputs.has-backend == 'true' && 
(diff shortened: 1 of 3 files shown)


Backend/default.nix
```diff
@@ -42,6 +42,12 @@
           inherit (self'.packages)
             arion;
         };
+        defaults.settings.defined = {
+          extraConfigureFlags =
+            if system == "aarch64-linux"
```
srid, line 47: Add a comment here explaining *why* we are changing optimization level for arm builds.

Backend/default.nix
```diff
@@ -42,6 +42,12 @@
...
+            then [ "--ghc-options=-O1" ]
+            else [ "--ghc-options=-O2" ];
```
srid, line 49:
```suggestion
          # ARM build fails on -O2
          extraConfigureFlags = lib.mkIf (system == "aarch64-linux") [ "--ghc-options=-O1" ];
```

.github/workflows/nix.yaml
srid, line 1: You also want to make it so that ARM builds happen only on `main` branch, and not PRs.

srid (commented):
Looks like CI passes. Just a couple of things to do.

Backend/default.nix
vsaimanohar: the build was failing for arm-linux builds with error

shared-services> /build/ghc612_0/ghc_329.s:160652:0: error:
shared-services>      Error: conditional branch out of range
shared-services>        |
shared-services> 160652 |         cbnz x17, .LceuZd 

temp-fix for this was to decrease optimization level for arm-linux builds

srid: Please copy paste that as a comment here. You should also comment the "why" since it is not always obvious when someone reads the code.

image

srid (approved):

PR #9973, opened 2025-01-23
backend/fix:intercity-distance-calculation

RahulGupta0706:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist


Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs
```diff
@@ -113,15 +114,46 @@ processWaypoints ::
...
   when calculationFailed do
     logWarning "Failed to calculate actual distance for this ride, ignoring"
   wrapDistanceCalculation driverId $ do
     addEditDestinationPoints driverId waypoints
     addPoints driverId waypoints
-    recalcDistanceBatches ih ending driverId estDist estTollCharges estTollNames pickupDropOutsideThreshold rectifyDistantPointsFailureUsing isTollApplicable enableTollCrossedNotifications waypoints calculationFailed passedThroughDrop
+    if isJust rectifyDistantPointsFailureUsing
+      then do
+        totalWaypoints <- getAllWaypointsImplementation driverId
+        when (length totalWaypoints > 1) do
```
0utkarsh, line 129: Why is this required?

Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs
```diff
@@ -113,15 +114,46 @@ processWaypoints ::
...
+      case points of
+        [] -> return ()
+        [_] -> return ()
+        (initialPoint : rest) -> do
+          (finalAccDist, finalMaxGap, _) <-
+            foldM
+              ( \(accDist, accMaxGap, prevPoint) currPoint -> do
+                  let dist = highPrecMetersToMeters $ distanceBetweenInMeters prevPoint currPoint
+                      newMaxGap = max accMaxGap dist
+                      newAccDist = if newMaxGap <= threshold then accDist + dist else accDist
```
0utkarsh, line 144: Why conditional?

RahulGupta0706: if the gap was coming greater than the threshold then I was not including that distance int the accumulator

Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs
RahulGupta0706: for just 1 point we cannot calculate the distance thats why added the check

Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs
0utkarsh: We can include. If safety is required, let's have an upper limit to reject based on time. For example, if we get strand point 2.5km per min, we can reject. This value should be configurable cc: @Karthickselvan28

Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs
0utkarsh: I think you've already handled it in `processWaypointsWithSnapToRoadCheck`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1044,6 +1046,8 @@ TransporterConfig:
     minDistanceForStopFcm: Maybe HighPrecMeters
     recomputeDistanceThresholds: Maybe Value
     minThresholdForPassThroughDestination: Maybe Meters
+    maxThresholdForSnapToRoadCall: Meters
+    maxThresholdToAccumulateDistance: Meters
```
0utkarsh, line 1050: Please make it optional and remove default values

0utkarsh (changes requested):

0utkarsh (dismissed):

0utkarsh (approved):

PR #10783, opened 2025-04-11
Feat/msil bulk driver vehicle mapping

FCJ6202:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for bulk vehicle and driver uploads with enhanced CSV fields and configurable options for route mapping and handling existing vehicles.
  - Introduced new access control mechanisms, allowing explicit distinction between requestor and fleet owner roles in fleet management actions.
  - Added new error messages for improved feedback during vehicle-driver association processes.

- **Improvements**
  - Enhanced validation and error reporting for vehicle and driver associations, including checks for existing links and verification status.
  - Extended transporter configuration with new settings to control route mapping requirements and bulk upload behavior.
  - Updated APIs to support optional driver names and requestor identification for greater flexibility.

- **Database Changes**
  - Added new boolean columns to transporter and merchant configuration tables to support the above features and controls.

- **Bug Fixes**
  - Improved handling of driver name updates to only occur when provided.


Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -237,15 +237,15 @@ postDriverFleetVerifyJoiningOtp merchantShortId opCity apiTokenInfo mbAuthId mbF
 postDriverFleetLinkRCWithDriver :: ShortId DM.Merchant -> City.City -> ApiTokenInfo -> Maybe Text -> Common.LinkRCWithDriverForFleetReq -> Flow APISuccess
 postDriverFleetLinkRCWithDriver merchantShortId opCity apiTokenInfo mbFleetOwnerId req = do
   checkedMerchantId <- merchantCityAccessCheck merchantShortId apiTokenInfo.merchant.shortId opCity apiTokenInfo.city
-  fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
-  Client.callFleetAPI checkedMerchantId opCity (.driverDSL.postDriverFleetLinkRCWithDriver) fleetOwnerId req
+  -- fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
```
khuzema786, line 240: We can keep this, as for normal fleet UI will not be sending fleetOwnerId in Query Param so it should not break

Backend/app/dashboard/CommonAPIs/spec/ProviderPlatform/Fleet/API/Driver.yaml
```diff
@@ -780,10 +780,12 @@ apis:
         userActionType: ApiAuth DRIVER_OFFER_BPP_MANAGEMENT FLEET LINK_RC_WITH_DRIVER
       helperApi:
         - POST:
-            endpoint: /fleet/linkRCWithDriver/{fleetOwnerId}
+            endpoint: /fleet/linkRCWithDriver/{requestorId}
```
khuzema786, line 783: will this be backward compatible ?

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -237,15 +237,15 @@ postDriverFleetVerifyJoiningOtp merchantShortId opCity apiTokenInfo mbAuthId mbF
...
+  Client.callFleetAPI checkedMerchantId opCity (.driverDSL.postDriverFleetLinkRCWithDriver) apiTokenInfo.personId.getId mbFleetOwnerId req
 
 postDriverFleetAddVehicles :: ShortId DM.Merchant -> City.City -> ApiTokenInfo -> Maybe Text -> Common.CreateVehiclesReq -> Flow Common.APISuccessWithUnprocessedEntities
 postDriverFleetAddVehicles merchantShortId opCity apiTokenInfo mbFleetOwnerId req = do
   checkedMerchantId <- merchantCityAccessCheck merchantShortId apiTokenInfo.merchant.shortId opCity apiTokenInfo.city
   transaction <- buildTransaction apiTokenInfo Nothing (Just req)
-  fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
-  T.withTransactionStoring transaction $ Client.callFleetAPI checkedMerchantId opCity (Common.addMultipartBoundary "XXX00XXX" . (.driverDSL.postDriverFleetAddVehicles)) req{fleetOwnerId = Just fleetOwnerId}
+  -- fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
```
khuzema786, line 247: lets keep this line

Backend/app/dashboard/CommonAPIs/spec/ProviderPlatform/Fleet/API/Driver.yaml
FCJ6202: Ya. Because the operators and fleets will use this API. That is why I did this. If fleet hits this API, requestorId contains fleetId and does their stuff but if the operator hits that API, we should do an extra one step to check fleet <-> operator association.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +533,111 @@ postDriverFleetAddVehicles ::
...
+    handleAddVehicleWithTry :: DM.Merchant -> Maybe Text -> Text -> Common.RegisterRCReq -> Text -> Maybe Text -> Flow (Either Text ())
+    handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq phoneNo mbFleetNo = do
+      _ <- RCQuery.findLastVehicleRCWrapper registerRcReq.vehicleRegistrationCertNumber >>= fromMaybeM (RCNotFound registerRcReq.vehicleRegistrationCertNumber) -- operator only add existing vehicle through bulk functionality
+      mbFleetOwnerId <- case mbFleetNo of
+        Just fleetNo -> do
+          phoneHash <- getDbHash fleetNo
+          let mobileCountryCode = fromMaybe DCommon.mobileIndianCode mbCountryCode
+          QPerson.findByMobileNumberAndMerchantAndRole mobileCountryCode phoneHash merchant.id DP.FLEET_OWNER >>= fromMaybeM (DriverNotFound phoneNo) >>= (pure . Just . (.id.getId))
+        Nothing -> pure Nothing
+      let addVehicleReq = convertToAddVehicleReq registerRcReq
```
0utkarsh, line 620: We can create a new function to accept `registerRcReq` like done on line 579 instead of adding vehicle with dummy values

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -1270,17 +1318,42 @@ postDriverFleetLinkRCWithDriver merchantShortId opCity fleetOwnerId req = do
...
+          return True
+
+    checkAssociationBetweenFleetAndOperator :: Flow Text
+    checkAssociationBetweenFleetAndOperator = do
+      requestedPerson <- QPerson.findById (Id requestorId) >>= fromMaybeM (PersonDoesNotExist requestorId)
+
+      case requestedPerson.role of
+        DP.OPERATOR -> do
+          case mbFleetOwnerId of
+            Nothing -> throwError $ InvalidRequest "Operator cannot be mapped DCO with vehicles."
```
0utkarsh, line 1347: Why not?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +533,111 @@ postDriverFleetAddVehicles ::
...
 
-    parseVehicleInfo :: DMOC.MerchantOperatingCity -> Int -> VehicleDetailsCSVRow -> Flow (Common.RegisterRCReq, DbHash, [DRoute.Route])
+    handleAddVehicleWithTry :: DM.Merchant -> Maybe Text -> Text -> Common.RegisterRCReq -> Text -> Maybe Text -> Flow (Either Text ())
+    handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq phoneNo mbFleetNo = do
+      _ <- RCQuery.findLastVehicleRCWrapper registerRcReq.vehicleRegistrationCertNumber >>= fromMaybeM (RCNotFound registerRcReq.vehicleRegistrationCertNumber) -- operator only add existing vehicle through bulk functionality
+      mbFleetOwnerId <- case mbFleetNo of
+        Just fleetNo -> do
+          phoneHash <- getDbHash fleetNo
+          let mobileCountryCode = fromMaybe DCommon.mobileIndianCode mbCountryCode
+          QPerson.findByMobileNumberAndMerchantAndRole mobileCountryCode phoneHash merchant.id DP.FLEET_OWNER >>= fromMaybeM (DriverNotFound phoneNo) >>= (pure . Just . (.id.getId))
```
0utkarsh, line 618: If we throw error, other entries will not be processed. This can be added in try catch block

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +533,111 @@ postDriverFleetAddVehicles ::
...
+              return $ unprocessedEntities <> catMaybes unprocessedVehicleRouteMapping
+          )
+          []
+          rcReq
+
+      unprocessedRCAdditionEntities <-
+        foldlM
+          ( \unprocessedEntities (registerRcReq, _, _, _, _, _) -> do
+              try @_ @SomeException
+                (postDriverFleetAddRCWithoutDriver merchantShortId opCity fleetOwnerId registerRcReq)
```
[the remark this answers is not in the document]
FCJ6202: This portion of the code is not run by MSIL because bulk functionality on vehicles only used by operators, not the fleet

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +533,111 @@ postDriverFleetAddVehicles ::
...
     readCsv csvFile merchantOpCity = do
       csvData <- L.runIO $ BS.readFile csvFile
       case (decodeByName $ LBS.fromStrict csvData :: Either String (Header, V.Vector VehicleDetailsCSVRow)) of
         Left err -> throwError (InvalidRequest $ show err)
         Right (_, v) -> V.imapM (parseVehicleInfo merchantOpCity) v >>= (pure . V.toList)
 
-    parseVehicleInfo :: DMOC.MerchantOperatingCity -> Int -> VehicleDetailsCSVRow -> Flow (Common.RegisterRCReq, DbHash, [DRoute.Route])
+    handleAddVehicleWithTry :: DM.Merchant -> Maybe Text -> Text -> Common.RegisterRCReq -> Text -> Maybe Text -> Flow (Either Text ())
+    handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq phoneNo mbFleetNo = do
+      _ <- RCQuery.findLastVehicleRCWrapper registerRcReq.vehicleRegistrationCertNumber >>= fromMaybeM (RCNotFound registerRcReq.vehicleRegistrationCertNumber) -- operator only add existing vehicle through bulk functionality
```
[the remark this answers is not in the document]
FCJ6202: Ya this check has inside the single add vehicle function.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: The operator cannot own any vehicle why do they only add vehicles to the DCO which is already done by single add vehicle function

FCJ6202: This function uses mapping a single vehicle to a single driver. If the operator wants to add the vehicle they can use the single add vehicle function. Single mapping function only used under fleet, not for DCO.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: In 579, without driver function is called, which takes registerRcReq, but we are using a single add vehicle function that is why we need to convert the type.

0utkarsh: I am saying create something similar, no need to use that exactly. Putting random values like model = "Standard" isn't correct

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -189,13 +191,19 @@ postDriverFleetAddVehicle merchantShortId opCity reqDriverPhoneNo requestorId mb
...
+  now <- getCurrentTime
+  existingLinkage <- DRCAE.findActiveLinkedByRCIdAndDriverId driverId vehicleRC.id now -- query with RC Active ??
+  if (isJust existingLinkage)
+    then return True
+    else do
+      vehicleActiveLinks <- DRCAE.findAllActiveLinkedByRCId vehicleRC.id now
+      unless (null vehicleActiveLinks) $ throwError VehicleAlreadyLinkedToAnotherDriver
+      driverActiveLinks <- DRCAE.findAllActiveLinkedByDriverId driverId now
+      unless (null driverActiveLinks) $ throwError DriverAlreadyLinkedToAnotherVehicle
+      return False
```
0utkarsh, line 206: Maybe create a query which will fetch all results with `driverId` or `rcId` and then check if any of these has both `rcId` and driverId for existing linkage check. No need for 3 queries

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +546,150 @@ postDriverFleetAddVehicles ::
...
+          >>= either (pure . Left . (\e -> "Unable to add Vehicle (" <> registerRcReq.vehicleRegistrationCertNumber <> "): " <> T.pack (displayException e))) (const $ pure $ Right ())
+
+    isVehicleAlreadyAssociatedWithFleetOrDriver :: Text -> Flow () -- checking vehicle present in the system or not
+    isVehicleAlreadyAssociatedWithFleetOrDriver vehicleNo = do
+      rc <- RCQuery.findLastVehicleRCWrapper vehicleNo
+      case rc of
+        Nothing -> throwError $ VehicleDoesNotExist vehicleNo
+        Just rcert -> do
+          fleetAssociation <- FRAE.findAllActiveAssociationByRCId rcert.id
+          driverAssociation <- DRCAE.findAllActiveAssociationByRCId rcert.id
```
0utkarsh, line 673: Shall we check rc table only for Valid entries?

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
```diff
@@ -124,3 +124,59 @@ mapping fleetIdWanted mbLimit mbOffset = do
...
+  (MonadFlow m, EsqDBFlow m r, CacheFlow m r) =>
+  Id Person ->
+  UTCTime ->
+  m [DriverRCAssociation]
+findAllActiveLinkedByDriverId (Id driverId) now = do
+  findAllWithOptionsKV
+    [ Se.And
+        [ Se.Is BeamDRCA.driverId $ Se.Eq driverId,
+          Se.Is BeamDRCA.associatedTill $ Se.GreaterThan $ Just now,
+          Se.Is BeamDRCA.isRcActive $ Se.Eq True
```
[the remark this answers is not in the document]
FCJ6202: Ya it should be in Maybe Type

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1011,6 +1011,8 @@ TransporterConfig:
     generateReferralCodeForOperator: Maybe Bool
     maxAllowedDocSizeInMB: Maybe Int
     bulkWaiveOffLimit : Int
+    notRequireRouteMappingInVehicle: Bool
```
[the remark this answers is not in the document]
FCJ6202: Yeah, but I think that we should make every field value for MSIL true. That's why I gave that type of name. Without it, we should make the value false. But we can do that by making the default value true and MSIL false.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: However, the RC table doesn't contain information about the driver's relationship with the vehicle. Again we have to query the driver RC association table to fetch the driver association data with the vehicle

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -109,39 +111,53 @@ getFleetOwnerId memberPersonId mbFleetOwnerId = do
...
-getFleetOwnerIds memberPersonId mbFleetOwnerId = do
+getFleetOwnerIdsFromBPP :: Text -> Maybe Text -> Flow [(Text, Text, DP.Role)]
+getFleetOwnerIdsFromBPP memberPersonId mbFleetOwnerId = do
   maybe
     ( FMA.findAllActiveByfleetMemberId memberPersonId True
         >>= \case
           [] -> do
-            person <- QP.findById (Id memberPersonId) >>= fromMaybeM (PersonNotFound memberPersonId)
-            return [(memberPersonId, person.firstName <> " " <> person.lastName)]
+            person <- QPerson.findById (Id memberPersonId) >>= fromMaybeM (PersonNotFound memberPersonId)
```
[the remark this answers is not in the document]
FCJ6202: didn't get this point

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: didn't get this point

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: Make sense

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
```diff
@@ -124,3 +124,59 @@ mapping fleetIdWanted mbLimit mbOffset = do
...
+  (MonadFlow m, EsqDBFlow m r, CacheFlow m r) =>
+  Id VehicleRegistrationCertificate ->
+  UTCTime ->
+  m [DriverRCAssociation]
+findAllActiveLinkedByRCId (Id rcId) now = do
+  findAllWithOptionsKV
+    [ Se.And
+        [ Se.Is BeamDRCA.rcId $ Se.Eq rcId,
+          Se.Is BeamDRCA.associatedTill $ Se.GreaterThan $ Just now,
+          Se.Is BeamDRCA.isRcActive $ Se.Eq True
```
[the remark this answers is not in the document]
FCJ6202: `activeAssociationsOfRC <- case transporterConfig.enableVehicle of -- Added MSIL check
    True -> do
      now <- getCurrentTime
      DRCAE.findAllActiveLinkedByRCId vehicleRC.id now
    False -> DRCAE.findAllActiveAssociationByRCId vehicleRC.id`
    
   In this portion of the code, we are adding a merchant base check. So to make the type equal to both functions I made the type of list.

Backend/app/dashboard/Lib/src/Storage/Beam/Merchant.hs
```diff
@@ -35,7 +35,8 @@ data MerchantT f = MerchantT
     authTokenHash :: B.C f (Maybe DbHash),
     enabled :: B.C f (Maybe Bool),
     createdAt :: B.C f Time.UTCTime,
-    requireAdminApprovalForFleetOnboarding :: B.C f (Maybe Bool)
+    requireAdminApprovalForFleetOnboarding :: B.C f (Maybe Bool),
+    requireGetFleetOwnerFunction :: B.C f (Maybe Bool)
```
0utkarsh, line 39: We can name it something like `hasFleetMemberHierarchy`

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -237,15 +241,21 @@ postDriverFleetVerifyJoiningOtp merchantShortId opCity apiTokenInfo mbAuthId mbF
...
 postDriverFleetAddVehicles :: ShortId DM.Merchant -> City.City -> ApiTokenInfo -> Maybe Text -> Common.CreateVehiclesReq -> Flow Common.APISuccessWithUnprocessedEntities
 postDriverFleetAddVehicles merchantShortId opCity apiTokenInfo mbFleetOwnerId req = do
   checkedMerchantId <- merchantCityAccessCheck merchantShortId apiTokenInfo.merchant.shortId opCity apiTokenInfo.city
   transaction <- buildTransaction apiTokenInfo Nothing (Just req)
-  fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
-  T.withTransactionStoring transaction $ Client.callFleetAPI checkedMerchantId opCity (Common.addMultipartBoundary "XXX00XXX" . (.driverDSL.postDriverFleetAddVehicles)) req{fleetOwnerId = Just fleetOwnerId}
+  merchant <- QMerchant.findByShortId merchantShortId >>= fromMaybeM (MerchantDoesNotExist merchantShortId.getShortId)
+  requestorId <- case merchant.requireGetFleetOwnerFunction of
+    Just True -> getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
+    _ -> return apiTokenInfo.personId.getId
```
0utkarsh, line 257: 254 to 257 is repeated in above three functions(and will be used in future), should be refactored

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1011,6 +1011,9 @@ TransporterConfig:
     generateReferralCodeForOperator: Maybe Bool
     maxAllowedDocSizeInMB: Maybe Int
     bulkWaiveOffLimit : Int
+    requireRouteMappingInVehicle: Bool
+    enableVehicle: Bool
```
0utkarsh, line 1015: `enableExistingVehicleInBulkUpload`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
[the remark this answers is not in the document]
0utkarsh: @FCJ6202 This is pending

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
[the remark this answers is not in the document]
0utkarsh: `findAll` is expensive compared to findOne, you can fetch one and then send array like

> example <- DRCAE.findOneActiveAssociationByRCId vehicleRC.id
> pure [example]

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +547,148 @@ postDriverFleetAddVehicles ::
...
+            )
+            []
+            rcReq
+
+      unprocessedRCAdditionEntities <-
+        foldlM
+          ( \unprocessedEntities (registerRcReq, _, _, mbCountryCode, mbFleetNo, mbDriverNo) -> do
+              decryptedMobileNumber <-
+                mapM decrypt requestedPerson.mobileNumber
+                  >>= fromMaybeM (InvalidRequest $ "Person do not have a mobile number " <> requestedPerson.id.getId)
```
0utkarsh, line 598: Error without try. I think we can add try at the top level instead of adding one by one

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +547,148 @@ postDriverFleetAddVehicles ::
...
+                >>= either (pure . Left . (\e -> "Fleet owner not found: " <> T.pack (displayException e))) (pure . Right)
+          case personResult of
+            Just person -> pure $ Just (person.id.getId)
+            Nothing -> throwE $ "Fleet owner not found for mobile number: " <> fleetNo
+        Nothing -> pure Nothing
+
+      let addVehicleReq = convertToAddVehicleReq registerRcReq
+
+      ExceptT $
+        try @_ @SomeException (postDriverFleetAddVehicle merchant.shortId opCity phoneNo requestorId mbFleetOwnerId mbCountryCode addVehicleReq)
```
0utkarsh, line 664: Can't find the code where `postDriverFleetAddVehicle` accepts `requestorId`, can you provide the reference?

0utkarsh (changes requested):

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
FCJ6202: Make sense

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: image

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: But we also need to return the unprocessed row error. That's why we are doing it one by one. If the current row is successfully processed, we return Nothing for this row, but if there is any error we get from the current row, we should return with that error

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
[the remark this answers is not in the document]
FCJ6202: This is a different query. It may also return two data points if there is no existing linkage between the given driverId and rcId. That's why it has to type of array.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Ride.hs
```diff
@@ -552,7 +552,7 @@ bookingWithVehicleNumberAndPhone merchant merchantOpCityId req = do
               }
       void $ DomainRC.linkRCStatus (personId, merchantId, merchantOpCityId) rcStatusReq
     createRCAssociation driverId rc = do
-      driverRCAssoc <- makeRCAssociation merchant.id merchantOpCityId driverId rc.id (DomainRC.convertTextToUTC (Just "2099-12-12"))
+      driverRCAssoc <- makeRCAssociation merchant.id merchantOpCityId driverId rc.id False (DomainRC.convertTextToUTC (Just "2099-12-12"))
```
roman-bodavskiy, line 555: No need to pass `rcActive` to this function, because its `False` everywhere
`end = (DomainRC.convertTextToUTC (Just "2099-12-12"))` is also the same

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +548,148 @@ postDriverFleetAddVehicles ::
 postDriverFleetAddVehicles merchantShortId opCity req = do
   merchant <- findMerchantByShortId merchantShortId
   merchantOpCity <- CQMOC.findByMerchantIdAndCity merchant.id opCity >>= fromMaybeM (MerchantOperatingCityNotFound $ "merchantShortId: " <> merchantShortId.getShortId <> " ,city: " <> show opCity)
-  fleetOwnerId <- case req.fleetOwnerId of
-    Nothing -> throwError FleetOwnerIdRequired
+  requestorId <- case req.requestorId of
+    Nothing -> throwError $ FleetOwnerIdRequired
```
roman-bodavskiy, line 552: `FleetOwnerIdRequired` s wrong error code here

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +548,148 @@ postDriverFleetAddVehicles ::
...
+                (Just fleetNo, Nothing) -> handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq fleetNo (Just fleetNo) -- Add vehicles under Fleet
+                (Nothing, Just driverNo) -> handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq driverNo Nothing -- Add vehicles under DCO
+                (Just fleetNo, Just driverNo) -> handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq driverNo (Just fleetNo) -- Map driver <-> vehicle under fleer
+              case currProcessEntity of
+                Left err -> return $ unprocessedEntities <> [err]
+                Right _ -> return unprocessedEntities
+          )
+          []
+          rcReq
+      pure Common.APISuccessWithUnprocessedEntities {unprocessedEntities = unprocessedRCAdditionEntities}
```
roman-bodavskiy, line 626: Could you explain, why vehicle route mapping is not here, the same as for `DP.FLEET_OWNER` case?

FCJ6202: Right now, this requirement is not needed in MSIL.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: I think `FleetOwnerOrOperatorIdRequired` is correct in this situation

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Ride.hs
FCJ6202: I didn't create this function. This function has already been made by someone, which takes an end parameter that would apply in `associationTill`  parameter

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: And the route mapping entity is also not present on the fleet side. You can see that I added a merchant base check on the fleet side

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +548,148 @@ postDriverFleetAddVehicles ::
...
+          []
+          rcReq
+      pure Common.APISuccessWithUnprocessedEntities {unprocessedEntities = unprocessedVehicleRouteMappingEntities <> unprocessedRCAdditionEntities}
+    DP.OPERATOR -> do
+      unprocessedRCAdditionEntities <-
+        foldlM
+          ( \unprocessedEntities (registerRcReq, _, _, mbCountryCode, mbFleetNo, mbDriverNo) -> do
+              currProcessEntity <- case (mbFleetNo, mbDriverNo) of
+                (Nothing, Nothing) -> pure $ Left $ "Unable to add Vehicle (" <> registerRcReq.vehicleRegistrationCertNumber <> "): Neither fleet nor driver phone number provided"
+                (Just fleetNo, Nothing) -> handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq fleetNo (Just fleetNo) -- Add vehicles under Fleet
```
roman-bodavskiy, line 617: Did you handled cases where vehicle already associated with another fleet?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
roman-bodavskiy: you mean this check

```
      unprocessedVehicleRouteMappingEntities <- case transporterConfig.requireRouteMappingInVehicle of
        False -> do
          pure []
```

so, for fleet owner you use this check, for operator role you consider that we have MSIL merchant and `transporterConfig.requireRouteMappingInVehicle = False`
Not obvious, but okay, will work

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: `postDriverFleetAddVehicle` This function is used for adding a single vehicle to the fleet, fleetDriver, and DCO. And I added all the checks in this function. That's why I just called this function  inside this helper function `handleAddVehicleWithTry `.

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -175,6 +176,14 @@ getDriverFleetAccessList merchantShortId opCity apiTokenInfo = do
       fleetOwners
   return $ Common.FleetOwnerListRes {..}
 
+------------------------------------- Helper Functions --------------------------------------
+getRequestorId :: ShortId DM.Merchant -> ApiTokenInfo -> Maybe Text -> Flow Text
+getRequestorId merchantShortId apiTokenInfo mbFleetOwnerId = do
+  merchant <- QMerchant.findByShortId merchantShortId >>= fromMaybeM (MerchantDoesNotExist merchantShortId.getShortId)
```
roman-bodavskiy, line 182: you can use `apiTokenInfo.merchant
`
As it follows from access check, this is the same:

```
postDriverFleetAddVehicle :: ShortId DM.Merchant -> City.City -> ApiTokenInfo -> Text -> Maybe Text -> Maybe Text -> Common.AddVehicleReq -> Flow APISuccess
postDriverFleetAddVehicle merchantShortId opCity apiTokenInfo phoneNo mbMobileCountryCode mbFleetOwnerId req = do
  runRequestValidation Common.validateAddVehicleReq req
  checkedMerchantId <- merchantCityAccessCheck merchantShortId apiTokenInfo.merchant.shortId opCity apiTokenInfo.city
.............
merchantCityAccessCheck :: MonadFlow m => ShortId DMerchant.Merchant -> ShortId DMerchant.Merchant -> City.City -> City.City -> m (CheckedShortId DMerchant.Merchant)
merchantCityAccessCheck (ShortId userMerchantId) (ShortId merchantId) userCity city = do
  unless (userMerchantId == merchantId && userCity == city) $ throwError AccessDenied
  pure $ CheckedShortId merchantId

```

FCJ6202: make sense

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
0utkarsh: Can you process this under Left instead of error?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +547,145 @@ postDriverFleetAddVehicles ::
...
+
+      case result of
+        Left e -> return $ Left $ "Error: " <> T.pack (displayException e)
+        Right _ -> return $ Right ()
+
+    isVehicleAlreadyAssociatedWithFleetOrDriver :: Text -> Flow () -- checking vehicle present in the system or not
+    isVehicleAlreadyAssociatedWithFleetOrDriver vehicleNo = do
+      rc <- RCQuery.findLastVehicleRCWrapper vehicleNo
+      case rc of
+        Nothing -> throwError $ VehicleDoesNotExist vehicleNo
```
0utkarsh, line 668: This error should only be for MSIL, right? For other merchants we should call `postDriverFleetAddRCWithoutDriver`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -532,72 +547,145 @@ postDriverFleetAddVehicles ::
...
+    handleAddVehicleWithTry merchant mbCountryCode requestorId registerRcReq phoneNo mbFleetNo = do
+      result <- try @_ @SomeException $ do
+        _ <- isVehicleAlreadyAssociatedWithFleetOrDriver registerRcReq.vehicleRegistrationCertNumber
+
+        mbFleetOwnerId <- case mbFleetNo of
+          Just fleetNo -> do
+            phoneHash <- getDbHash fleetNo
+            let mobileCountryCode = fromMaybe DCommon.mobileIndianCode mbCountryCode
+            personResult <- QPerson.findByMobileNumberAndMerchantAndRole mobileCountryCode phoneHash merchant.id DP.FLEET_OWNER
+            case personResult of
```
0utkarsh, line 651: `fromMaybeM` in line 650 will not require case match

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: It is my mistake, @0utkarsh . For now, every merchant will run this code. We must also add a merchant base check with this field `enableExistingVehicleInBulkUpload`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: Didn't get this point

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -260,22 +267,30 @@ checkEnitiesAssociationValidation merchantId merchantOpCityId requestorId mbFlee
...
+  when (isNothing isAssociated) $
+    throwError (DriverNotActiveInFleet driverPerson.id.getId fleetOwnerId)
 
 checkRCAssociationForFleet :: Text -> VehicleRegistrationCertificate -> Flow ()
 checkRCAssociationForFleet fleetOwnerId vehicleRC = do
   when (isJust vehicleRC.fleetOwnerId && vehicleRC.fleetOwnerId /= Just fleetOwnerId) $ throwError VehicleBelongsToAnotherFleet
-  activeAssociationsOfRC <- DRCAE.findAllActiveAssociationByRCId vehicleRC.id
+  transporterConfig <- case vehicleRC.merchantOperatingCityId of
+    Just merchantOpCityId -> SCTC.findByMerchantOpCityId merchantOpCityId Nothing >>= fromMaybeM (TransporterConfigNotFound merchantOpCityId.getId)
+    Nothing -> throwError (InvalidRequest "Merchant Operating City not found for vehicle")
```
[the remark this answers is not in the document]
FCJ6202: So you told me that we should use the merchant city ID, which we get from the API request, not from VehicleRC data?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -260,22 +267,30 @@ checkEnitiesAssociationValidation merchantId merchantOpCityId requestorId mbFlee
...
+  activeAssociationsOfRC <- case transporterConfig.enableExistingVehicleInBulkUpload of -- Added MSIL check
+    True -> do
+      now <- getCurrentTime
+      activeRCData <- DRCAE.findOneActiveLinkedByRCId vehicleRC.id now
```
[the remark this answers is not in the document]
FCJ6202: I think `verificationStatus` tells us whether the vehicle Document is valid or not. Right? Which is I check in `isVehicleAlreadyAssociatedWithFleetOrDriver` in the function, which is used in bulk vehicle functionality for checking vehicle is already present in the system or not

- In this scenario, we are checking that the vehicle has an active association with a driver. So, does that driver come under this fleet or not.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: sure

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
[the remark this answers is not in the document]
0utkarsh: Yes

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverRCAssociationExtra.hs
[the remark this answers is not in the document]
roman-bodavskiy: I found that this fucntion did not used in code

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Fleet/Driver.hs
```diff
@@ -237,15 +244,15 @@ postDriverFleetVerifyJoiningOtp merchantShortId opCity apiTokenInfo mbAuthId mbF
 postDriverFleetLinkRCWithDriver :: ShortId DM.Merchant -> City.City -> ApiTokenInfo -> Maybe Text -> Common.LinkRCWithDriverForFleetReq -> Flow APISuccess
 postDriverFleetLinkRCWithDriver merchantShortId opCity apiTokenInfo mbFleetOwnerId req = do
   checkedMerchantId <- merchantCityAccessCheck merchantShortId apiTokenInfo.merchant.shortId opCity apiTokenInfo.city
-  fleetOwnerId <- getFleetOwnerId apiTokenInfo.personId.getId mbFleetOwnerId
-  Client.callFleetAPI checkedMerchantId opCity (.driverDSL.postDriverFleetLinkRCWithDriver) fleetOwnerId req
+  requestorId <- getRequestorId apiTokenInfo mbFleetOwnerId
```
roman-bodavskiy, line 247: here I think you need to use merchant based check for both requestorId, fleetOwnerId 
See getMbFleetOwnerAndRequestorIdMerchantBased,  getFleetOwnerAndRequestorIdMerchantBased functions from my PR

PR #11402, opened 2025-06-16
backend/feat: Introduce DriverFlowStatus and enhance driver-related A…

FCJ6202:
…PIs for improved status tracking

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced driver flow status tracking with statuses such as ONLINE, OFFLINE, SILENT, ON_PICKUP, ON_RIDE, ACTIVE, and INACTIVE.
  * Added new API endpoints to fetch aggregated driver status counts for fleet owners and operators.
  * Implemented caching of driver flow status counts for efficient retrieval.

* **Enhancements**
  * Integrated driver flow status updates across driver onboarding, association, activity changes, ride start/end, and logout flows.
  * Added configuration to enable or disable driver flow status caching per transporter.

* **Database Changes**
  * Added new columns and tables to support driver flow status and associations in both relational and ClickHouse databases.
  * Updated migration scripts to accommodate new schema changes.

* **Access Control**
  * Updated access permissions to allow relevant roles to use the new driver status API.

* **Documentation**
  * Updated API specifications and configuration documentation to reflect new features and fields.


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverInformation.hs
```diff
@@ -0,0 +1,46 @@
...
+    driverFlowStatus :: C f (Maybe DDF.DriverFlowStatus)
+  }
+  deriving (Generic)
+
+deriving instance Show DriverInformation
+
+driverInformationTTable :: DriverInformationT (FieldModification DriverInformationT)
+driverInformationTTable =
+  DriverInformationT
+    { id = "id",
```
roman-bodavskiy: but `DriverInformation` table does not have `id` field, only `driverid`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverInformation.hs
```diff
@@ -0,0 +1,46 @@
...
+  CH.findAll $
+    CH.select_
+      ( \info -> do
+          let driverFlowStatus = info.driverFlowStatus
+          let countDrivers = CH.count_ (info.driverId)
+          CH.groupBy driverFlowStatus $ \m -> (m, countDrivers)
+      )
+      $ CH.filter_
+        (\info _ -> info.driverId `CH.in_` driverIds)
+        (CH.all_ @CH.APP_SERVICE_CLICKHOUSE driverInformationTTable)
```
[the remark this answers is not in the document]
roman-bodavskiy: current Haskell lib implementation already handled empty `IN` list:

```
in_ :: forall a table value. (ClickhouseTable table, ClickhouseValue value) => Column a table value -> [value] -> Clause table
in_ _column [] = val_ False
in_ column (value1 : values) = column `Is` In (value1 NE.:| values)
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverOperatorAssociation.hs
```diff
@@ -0,0 +1,42 @@
...
+
+deriving instance Show DriverOperatorAssociation
+
+driverOperatorAssociationTTable :: DriverOperatorAssociationT (FieldModification DriverOperatorAssociationT)
+driverOperatorAssociationTTable =
+  DriverOperatorAssociationT
+    { id = "id",
+      driverId = "driver_id",
+      operatorId = "operator_id",
+      isActive = "is_active"
```
roman-bodavskiy, line 26: can we add check on `associatedTill` also, the same as in Beam queries?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverOperatorAssociation.hs
```diff
@@ -0,0 +1,42 @@
...
+import Kernel.Prelude
+import Kernel.Storage.ClickhouseV2 as CH
+import qualified Kernel.Storage.ClickhouseV2.UtilsTH as TH
+import Kernel.Types.Id
+
+data DriverOperatorAssociationT f = DriverOperatorAssociationT
+  { id :: C f (Id DOA.DriverOperatorAssociation),
+    driverId :: C f (Id DP.Person),
+    operatorId :: C f Text,
+    isActive :: C f Bool
```
roman-bodavskiy, line 14: can we add check on `associatedTill` also, the same as in Beam queries?

Backend/app/dashboard/CommonAPIs/spec/ProviderPlatform/Fleet/API/Driver.yaml
```diff
@@ -944,6 +944,24 @@
...
+        - GET:
+            endpoint: /{requestorId}/fleet/status
+            auth: ApiAuthV2
+            params:
+              requestorId: Text
+            query:
+              - fleetOwnerId: Text
+            response:
+              type: DriverStatusRes
 types:
```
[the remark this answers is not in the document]
roman-bodavskiy: This is not required. We added this only when migrated apis from manual code to DSL

Backend/app/dashboard/CommonAPIs/spec/ProviderPlatform/Fleet/API/Driver.yaml
```diff
@@ -1325,6 +1343,14 @@ types:
...
     - timeBounds: Maybe TimeBound
     - derive: "'HideSecrets"
+  DriverStatusRes:
+    - online: Int
+    - offline: Int
+    - silent: Int
+    - toPickup: Int
+    - onRide: Int
+    - active: Int
+    - inactive: Int
```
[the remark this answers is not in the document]
roman-bodavskiy: not required for GET apis, because we don't store transactions for GET apis

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverOperatorAssociation.hs
FCJ6202: I don't think so, it is needed because we added a check with the isActive field, which is enough because if isActive is true, then associationTill must be greater than now

roman-bodavskiy (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -127,6 +129,9 @@ import qualified SharedLogic.WMB as WMB
 import Storage.Beam.SystemConfigs ()
 import qualified Storage.Cac.TransporterConfig as SCTC
 import qualified Storage.CachedQueries.Merchant.MerchantOperatingCity as CQMOC
+import qualified Storage.Clickhouse.DriverInformation as CDI
+import qualified Storage.Clickhouse.DriverOperatorAssociation as CDOA
```
0utkarsh: Let's make sure this table is drained to clickhouse also

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -1024,6 +1029,40 @@ castDriverStatus = \case
...
+              offline = get DDF.OFFLINE,
+              silent = get DDF.SILENT,
+              toPickup = get DDF.ON_PICKUP,
+              onRide = get DDF.ON_RIDE,
+              active = get DDF.ACTIVE,
+              inactive = getInactive
+            }
+
+    toStatusMap :: [(Maybe DDF.DriverFlowStatus, Int)] -> Map.Map (Maybe DDF.DriverFlowStatus) Int
+    toStatusMap = Map.fromListWith (+)
```
0utkarsh: Why do we `Map.fromListWith (+)`? I am assuming grouped query will already return unique elements. We can use `Map.fromList`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/DriverOperatorAssociation.hs
```diff
@@ -0,0 +1,42 @@
...
+  DriverOperatorAssociationT
+    { id = "id",
+      driverId = "driver_id",
+      operatorId = "operator_id",
+      isActive = "is_active"
+    }
+
+type DriverOperatorAssociation = DriverOperatorAssociationT Identity
+
+$(TH.mkClickhouseInstances ''DriverOperatorAssociationT 'NO_SELECT_MODIFIER)
```
0utkarsh: Same `driverId` will come multiple times in query `getDriverIdsByOperatorId`. We should drop the duplicates after fetching if not using final modifier here.

0utkarsh: Same for Fleet Operator Association

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: Yeah, it is tested @0utkarsh, all the tables created in ClickHouse

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +41,14 @@ setDriverMode apiKey req = do
...
+  void $ QDriverInformation.updateActivity isActive (Just mode) (Just newFlowStatus) driverId
   pure Success
+
+getDriverFlowStatus :: Maybe DriverInfo.DriverMode -> Bool -> DDFS.DriverFlowStatus
+getDriverFlowStatus mode isActive =
+  case mode of
+    Just DriverInfo.ONLINE -> DDFS.ONLINE
+    Just DriverInfo.SILENT -> DDFS.SILENT
+    Just DriverInfo.OFFLINE -> DDFS.OFFLINE
+    Nothing -> if isActive then DDFS.ACTIVE else DDFS.INACTIVE
```
roman-bodavskiy, line 64: ACTIVE and INACTIVE related to subscription balance, as @0utkarsh mentioned.
Maybe it should be ONLINE and OFFLINE?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
roman-bodavskiy: @FCJ6202 queries that you added in `Backend/dev/clickhouse/sql-seed` are only for dev. They does not used for master and prod

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +46,42 @@ setDriverMode apiKey req = do
...
+  -- Try to find active FleetDriverAssociation
+  mbFleetDriverAssociation <- QFleetDriverAssociationExtra.findByDriverId driverId True
+  (entityId, entityType) <- case mbFleetDriverAssociation of
+    Just fda -> pure (fda.fleetOwnerId, "fleetOwnerId" :: Text)
+    Nothing -> do
+      -- If not found, try to find active DriverOperatorAssociation
+      mbDriverOperatorAssociation <- QDriverOperatorAssociationExtra.findByDriverId driverId True
+      case mbDriverOperatorAssociation of
+        Just doa -> pure (doa.operatorId, "operatorId" :: Text)
+        Nothing -> throwError $ InternalError "No active FleetDriverAssociation or DriverOperatorAssociation found for driver"
```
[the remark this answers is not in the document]
FCJ6202: The driver must have an association with any one of the entities. If not, then there is some inconsistency in DB, that why we should give an error in that situation

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +46,42 @@ setDriverMode apiKey req = do
...
+  let oldKey = DDFS.getStatusKey oldStatus entityId
+  let newKey = DDFS.getStatusKey newStatus entityId
+  logInfo $ "Updating driver status keys: oldStatus=" <> show oldStatus <> ", newStatus=" <> show newStatus <> ", entityId=" <> show entityId <> ", entityType=" <> show entityType
+  void $ Redis.decr oldKey
+  void $ Redis.incr newKey
```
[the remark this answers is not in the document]
FCJ6202: @roman-bodavskiy @0utkarsh is this mandatory? I don't think we should make a call increment and decrement together as an atomic operation, because it will only be hit by the driver to change the status, not multiple fleets or operators are trying to change the status of a driver at the same time

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -1034,20 +1034,47 @@ getDriverFleetStatus :: ShortId DM.Merchant -> Context.City -> Text -> Maybe Tex
...
+            logTagInfo "DriverStatus" $ "Redis.set " <> k <> " = " <> show count
+            Redis.set @T.Text k (T.pack (show count))
+          Nothing -> pure ()
       pure $ toDriverStatusRes driverModeInfo
-
+  where
+    readInt :: String -> Maybe Int
+    readInt s = case reads s of
+      [(i, _)] -> Just i
+      _ -> Nothing
```
roman-bodavskiy: why simple `readMaybe` function did not used? Moreover redis can get and set as `Int`:
```
Redis.get @Int key
Redis.set @Int count
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -1034,20 +1034,47 @@ getDriverFleetStatus :: ShortId DM.Merchant -> Context.City -> Text -> Maybe Tex
...
       driverIds <- case entityRole of
         DP.FLEET_OWNER -> CFDV.getDriverIdsByFleetOwnerId entityId
         DP.OPERATOR -> CDOA.getDriverIdsByOperatorId entityId
         _ -> throwError (InvalidRequest "Invalid Data")
+      logTagInfo "DriverStatus" $ "Fetched driverIds from association: " <> show driverIds
       driverModeInfo <- CDI.getModeCountsByDriverIds driverIds
+      logTagInfo "DriverStatus" $ "ClickHouse returned: " <> show driverModeInfo <> ". Updating Redis."
+      forM_ driverModeInfo $ \(mbStatus, count) ->
+        case mbStatus of
+          Just status -> do
```
roman-bodavskiy: `whenJust mbStatus $ \status -> do`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +46,42 @@ setDriverMode apiKey req = do
   locationTrackingServiceKey <- asks (.locationTrackingServiceKey)
   unless (apiKey == Just locationTrackingServiceKey) $ do
     throwError $ InvalidRequest "Invalid API key"
-  void $ QDriverInformation.updateActivity isActive (Just mode) driverId
+  let newFlowStatus = getDriverFlowStatus (Just mode) isActive
+  updateFleetOperatorStatusKeyForDriver driverId newFlowStatus Nothing
+  void $ QDriverInformation.updateActivity isActive (Just mode) (Just newFlowStatus) driverId
```
roman-bodavskiy: `updateActivity` and `updateFleetOperatorStatusKeyForDriver` always should be used together. Maybe add some helper function?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
roman-bodavskiy: Are you sure @FCJ6202 ? what about new drivers without association or with inactive association. Should we throw error when this new driver will change mode?
Also, what about other merchants, not MSIL? drivers not always have association in general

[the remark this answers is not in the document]
roman-bodavskiy: I'm thinking, maybe we should make this caching logic with redis keys merchant based? Why other merchants should handle it, if it will never used?
cc @0utkarsh

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
roman-bodavskiy: @FCJ6202 I think atomic operation is better but not mandatory, don't think any race condition possible, only redis cache will be inconsistent in small period of time between `decr` and `incr`. We have `runHedisTransaction` func in lib

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -52,3 +58,30 @@ getDriverFlowStatus mode isActive =
...
+    Nothing -> do
+      -- If not found, try to find active DriverOperatorAssociation
+      mbDriverOperatorAssociation <- QDriverOperatorAssociationExtra.findByDriverId driverId True
+      case mbDriverOperatorAssociation of
+        Just doa -> pure (doa.operatorId, "operatorId" :: Text)
+        Nothing -> throwError $ InternalError "No active FleetDriverAssociation or DriverOperatorAssociation found for driver"
+  let oldKey = DDFS.getStatusKey oldStatus entityId
+  let newKey = DDFS.getStatusKey newStatus entityId
+  logInfo $ "Updating driver status keys: oldStatus=" <> show oldStatus <> ", newStatus=" <> show newStatus <> ", entityId=" <> show entityId <> ", entityType=" <> show entityType
+  void $ Redis.decr oldKey
```
roman-bodavskiy: How `incr` an `decr` will work at first call when redis values are empty? It will store `newKey => 1` and `oldKey => -1`?

roman-bodavskiy: Did you handled case when association between driver and fleet/operator become active/inactive? Then you need to incr/decr only one key

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
FCJ6202: Yeah we should add merchant base check here

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: We need to update the addDriver and removeDriver APIs; then, both cases will be handled by this update.

roman-bodavskiy: And also `postFleetConsent`, `postDriverConsent` for bulk upload, take attention to `endDriverAssociationsIfAllowed` function (association overwrite case).
Generally, every time when we end active association, create active association, or change `isActive` field, we need to handle

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: It is my mistake I think, because right now we are consider Nothing as INACTIVE status

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -1024,6 +1029,64 @@ castDriverStatus = \case
...
+      allKeys
+  if all isJust redisCounts
+    then do
+      logTagInfo "DriverStatus" $ "Cache hit for all statuses for entityId: " <> entityId <> ", counts: " <> show redisCounts
+      pure $ toDriverStatusRes (zip (map Just statusList) (map (fromMaybe 0) redisCounts))
+    else do
+      logTagInfo "DriverStatus" $ "Cache miss for some statuses for entityId: " <> entityId <> ". Falling back to ClickHouse. Redis counts: " <> show redisCounts
+      driverIds <- case entityRole of
+        DP.FLEET_OWNER -> CFDV.getDriverIdsByFleetOwnerId entityId
+        DP.OPERATOR -> CDOA.getDriverIdsByOperatorId entityId
```
0utkarsh: Operator will see indirect drivers also (D->F->O)  and this query is only fetching from operators table

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
0utkarsh: Let's not throw error here. As I may get added as DCO and then get deleted from operator

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
0utkarsh: Need to make sure the new column `newFlowStatus` should be drained to clickhouse in prod

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
FCJ6202: Yeah, because at that time we decided to show only DCO things, and if the operator wants to see the fleet status count, then they should go under the fleet to see that

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: Yeah, I added the query locally and tested it. When @dushyant-2002 integrates the clickhouse then we should go through the query which we defined in this PR

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
FCJ6202: We need the fleet or operator ID to create a key, because then we are going to update the fleet/operator status count when the driver comes under some entity.  And this part only runs for MSIL driver, but if there is any requirement like an independent driver, also change their status, then we shouldn't call the decrement and increment function, simply return from this function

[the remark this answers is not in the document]
FCJ6202: @0utkarsh let me know what should we do in this situation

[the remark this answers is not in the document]
0utkarsh: We can just remove the error. No key creation if not part of any entity

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
0utkarsh: Requirement got updated I think, we need to show all drivers by default

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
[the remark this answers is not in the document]
FCJ6202: Yeah I will update this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Clickhouse/FleetDriverAssociation.hs
```diff
@@ -0,0 +1,53 @@
...
+getDriverIdsByFleetOwnerIds ::
+  CH.HasClickhouseEnv CH.APP_SERVICE_CLICKHOUSE m =>
+  [Text] ->
+  m [Id DP.Person]
+getDriverIdsByFleetOwnerIds fleetOwnerIds =
+  CH.findAll $
+    CH.select_ (\assoc -> CH.notGrouped (assoc.driverId)) $
+      CH.filter_
+        (\assoc _ -> (assoc.fleetOwnerId `CH.in_` fleetOwnerIds) CH.&&. (assoc.isActive CH.==. True))
+        (CH.all_ @CH.APP_SERVICE_CLICKHOUSE fleetDriverAssociationTTable)
```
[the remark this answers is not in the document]
FCJ6202: At a time driver has an active association with one entity. So that list must contain distinct drivers.

roman-bodavskiy (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +50,104 @@ setDriverMode apiKey req = do
...
+  m ()
+decrementFleetOperatorStatusKeyForDriver entityType entityId mbStatus = do
+  let status = fromMaybe DDFS.INACTIVE mbStatus
+  let key = DDFS.getStatusKey entityId status
+  keyExists <- Redis.get @T.Text key
+  let msg = "Key does not exist for driver status key: " <> show key <> " for status=" <> show status <> ", entityId=" <> show entityId
+  if isNothing keyExists
+    then fork msg $ do
+      void $ SDF.handleCacheMissForDriverFlowStatus entityType entityId (DDFS.allKeys entityId)
+      void $ Redis.decr key
```
0utkarsh: This might decrement INACTIVE and make it negative. Should we do this operation only when status is available?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+-- Helper to wait until a Redis key is gone (polling every 100ms)
+waitUntilKeyGone :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r, Redis.HedisFlow m r) => Text -> m ()
+waitUntilKeyGone key = do
+  exists <- Redis.get @Bool key
+  if exists == Just True
+    then do
+      liftIO $ threadDelay 100000 -- 100ms
+      waitUntilKeyGone key
```
0utkarsh: Should we add some counter and stop this going in infinite loop for safety?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
```diff
@@ -0,0 +1,112 @@
...
+      redisCounts' <-
+        mapM
+          ( \key ->
+              Redis.get @Int key >>= \v -> do
+                logTagInfo "DriverStatus" $ "Redis.get (after wait) " <> key <> " => " <> show v
+                pure v
+          )
+          allKeys
+      pure $ toDriverStatusRes (zip (map Just DDF.statusList) (map (fromMaybe 0) redisCounts'))
+    _ -> do
```
0utkarsh, line 47: Race condition is possible here

Two users read `inProgressKey` and find it missing (Nothing).
Both will proceed to set `inProgressKey` and compute.

roman-bodavskiy (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +50,104 @@ setDriverMode apiKey req = do
   locationTrackingServiceKey <- asks (.locationTrackingServiceKey)
   unless (apiKey == Just locationTrackingServiceKey) $ do
     throwError $ InvalidRequest "Invalid API key"
-  void $ QDriverInformation.updateActivity isActive (Just mode) driverId
+  let newFlowStatus = getDriverFlowStatus (Just mode) isActive
```
0utkarsh, line 54: Can we make this optional and only update when we get some flow status? Nothing mapped with Active/Inactive might be wrong

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +50,104 @@ setDriverMode apiKey req = do
...
+  mbFleetDriverAssociation <- QFleetDriverAssociationExtra.findByDriverId driverId True
+  mbEntity <- case mbFleetDriverAssociation of
+    Just fda -> pure $ Just (fda.fleetOwnerId, DP.FLEET_OWNER)
+    Nothing -> do
+      -- If not found, try to find active DriverOperatorAssociation
+      mbDriverOperatorAssociation <- QDriverOperatorAssociationExtra.findByDriverId driverId True
+      pure $ fmap (\doa -> (doa.operatorId, DP.OPERATOR)) mbDriverOperatorAssociation
+  whenJust mbEntity $ \(entityId, entityType) -> do
+    decrementFleetOperatorStatusKeyForDriver entityType entityId (Just oldStatus)
+    incrementFleetOperatorStatusKeyForDriver entityType entityId (Just newStatus)
```
0utkarsh, line 87: Fleet's operator also

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +50,104 @@ setDriverMode apiKey req = do
...
+  Text ->
+  Maybe DDFS.DriverFlowStatus ->
+  m ()
+decrementFleetOperatorStatusKeyForDriver entityType entityId mbStatus = do
+  let status = fromMaybe DDFS.INACTIVE mbStatus
+  let key = DDFS.getStatusKey entityId status
+  keyExists <- Redis.get @T.Text key
+  let msg = "Key does not exist for driver status key: " <> show key <> " for status=" <> show status <> ", entityId=" <> show entityId
+  if isNothing keyExists
+    then fork msg $ do
```
0utkarsh: We will need the key here, right? Fork will make the below two run parallelly

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -807,7 +809,9 @@ setActivity (personId, merchantId, merchantOpCityId) isActive mode = do
               QDriverInformation.updateBlockedState driverId False (Just "AUTOMATICALLY_UNBLOCKED") merchantId merchantOpCityId DTDBT.Application
             else throwError $ DriverAccountBlocked (BlockErrorPayload driverInfo.blockExpiryTime driverInfo.blockReasonFlag)
         Nothing -> throwError $ DriverAccountBlocked (BlockErrorPayload driverInfo.blockExpiryTime driverInfo.blockReasonFlag)
-  when (driverInfo.active /= isActive || driverInfo.mode /= mode) $ QDriverInformation.updateActivity isActive (mode <|> Just DriverInfo.OFFLINE) driverId
+  when (driverInfo.active /= isActive || driverInfo.mode /= mode) $ do
+    let newFlowStatus = DDriverMode.getDriverFlowStatus (mode <|> Just DriverInfo.OFFLINE) isActive
```
0utkarsh, line 813: Can be made optional?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Operator.hs
```diff
@@ -39,15 +40,18 @@ postOperatorConsent (mbDriverId, merchantId, merchantOperatingCityId) = do
...
 
-  SA.endDriverAssociationsIfAllowed merchant merchantOperatingCityId driver
+  SA.endDriverAssociationsIfAllowed merchant merchantOperatingCityId transporterConfig driver
 
   DOR.makeDriverReferredByOperator merchantOperatingCityId driverId operator.id
 
   QDriverOperatorAssociation.updateByPrimaryKey driverOperatorAssociation{isActive = True}
+  when (transporterConfig.allowCacheDriverFlowStatus == Just True) $ do
+    driverInfo <- QDI.findById driverOperatorAssociation.driverId >>= fromMaybeM (DriverNotFound driverOperatorAssociation.driverId.getId)
+    DDriverMode.incrementFleetOperatorStatusKeyForDriver OPERATOR driverOperatorAssociation.operatorId driverInfo.driverFlowStatus
```
0utkarsh, line 52: But this was being done in verify too.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
```diff
@@ -162,7 +166,10 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
   QRB.updateStatus booking.id SRB.COMPLETED
   whenJust mbFareParams QFare.create
   QRide.updateAll ride.id ride
-  driverInfo <- QDI.findById (cast ride.driverId) >>= fromMaybeM (PersonNotFound ride.driverId.getId)
+  oldDriverInfo <- QDI.findById (cast ride.driverId) >>= fromMaybeM (PersonNotFound ride.driverId.getId)
+  let newFlowStatus = DDriverMode.getDriverFlowStatus oldDriverInfo.mode oldDriverInfo.active
```
0utkarsh, line 170: Old status will be `ON_RIDE` mostly. Any reason why we are taking this?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
```diff
@@ -162,7 +166,10 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
...
+  DDriverMode.updateDriverModeAndFlowStatus driverId thresholdConfig.allowCacheDriverFlowStatus oldDriverInfo.active oldDriverInfo.mode newFlowStatus (Just oldDriverInfo)
+  let driverInfo = oldDriverInfo {DI.driverFlowStatus = Just newFlowStatus}
```
0utkarsh, line 172: What is the use of this?

Backend/test/src/Mobility/ARDU/NearestDrivers.hs
```diff
@@ -120,4 +119,10 @@ setDriversActive :: Bool -> Maybe DI.DriverMode -> FlowR ARDUEnv.AppEnv ()
 setDriversActive isActive mode = do
   -- Esq.runTransaction $ do
   let drivers = [furthestDriver, closestDriver, suvDriver, sedanDriver, hatchbackDriver, driverWithOldLocation]
-  forM_ drivers (\driver -> Q.updateActivity isActive mode (Id driver))
+  let allowCacheDriverFlowStatus = Nothing -- TODO: Need to discuss this
+  let newFlowStatus = DDriverMode.getDriverFlowStatus mode isActive
```
0utkarsh, line 123: We don't need to make changes in test module (for future reference)

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
roman-bodavskiy: For consistency. This will not increase queries @0utkarsh . Anyway `driverInfo` will be used later

`createDriverFee booking.providerId booking.merchantOperatingCityId driverId ride.fare ride.currency newFareParams driverInfo booking serviceName`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Operator.hs
roman-bodavskiy: `postOperatorConsent` used for bulk upload, `verify` for single driver

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
roman-bodavskiy: > Fork will make the below two run parallelly

Why? This two will run sequentially

```
      void $ SDF.handleCacheMissForDriverFlowStatus entityType entityId (DDFS.allKeys entityId)
      void $ Redis.decr key
```

Fork means here only that parent fork `decrementFleetOperatorStatusKeyForDriver` will not wait, when these two operations (child fork) will be finished

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
roman-bodavskiy: I think infinity loop is not possible because key will be expired or deleted, but if you want explicit restriction, we can put expiry here

```
waitUntilKeyGone :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r, Redis.HedisFlow m r) => Text -> UTCTime -> m ()
waitUntilKeyGone key expiry = do
  now <- getCurrentTime
  buffer = 5 -- 5 seconds
  when (now < addUTCTime expiry buffer)
    exists <- Redis.get @Bool key
    if exists == Just True
      ....
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: This status belongs to the driver, not to a fleet. That why I don't think it is needed here

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
FCJ6202: What's optional? didn't get this point

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Operator.hs
FCJ6202: OperatorConcent is called when DCO clicks the link that they got when the operator added this driver through bulk upload. And verify will call when DCO is added through a single functionality

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
FCJ6202: So, after ending the ride, we need to update the driver flow status ON_RIDE to some other status. So we can use the mode value to update the driver_flow_status, which should be ONLINE, OFFLINE, or SILENT

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
FCJ6202: Explained in above comment

Backend/test/src/Mobility/ARDU/NearestDrivers.hs
FCJ6202: I did this to compile my code and make the code consistent everywhere

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: Use a counter or timeout?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
FCJ6202: Use a counter or timeout?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: I didn't get this point to become negative for inactive case because if the driver part of the fleet and their status is null or inactive right now, then we should get some inactive count from redis because we are calculating the count, then decrement the value

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
FCJ6202: Yeah need to handle this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
```diff
@@ -0,0 +1,117 @@
...
+            pure $ toDriverStatusRes driverModeInfo
+          Redis.del inProgressKey
+          case res of
+            Left err -> do
+              logTagError "DriverStatus" $ "Error during ClickHouse/Redis to get driver count under each status for entityId: " <> entityId <> ". Error: " <> show err
+              throwError (InternalError $ "Failed to get driver count under each status for entityId: " <> entityId)
+            Right statusRes -> pure statusRes
+        else do
+          logTagInfo "DriverStatus" $ "inProgress lock already held for entityId: " <> entityId <> " (race detected in else). Waiting for it to clear."
+          handleCacheMissForDriverFlowStatus entityRole entityId allKeys
```
roman-bodavskiy, line 73: here busy waiting is possible, recursive calls of `handleCacheMissForDriverFlowStatus` without any delay

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +51,187 @@ setDriverMode apiKey req = do
...
+    (True, False) -> fork "Operator key data not found" $ do
+      _ <- handleFallback DP.OPERATOR operatorId operatorKeys
+      decrementKeysWithCounts (map (fromMaybe 0) fleetRedisCount)
+    (False, True) -> fork "Fleet owner key data not found" $ do
+      newFleetRedisCount <- handleFallback DP.FLEET_OWNER fleetOwnerId fleetOwnerKeys
+      decrementKeysWithCounts (map (fromMaybe 0) newFleetRedisCount)
+    (False, False) -> fork "Both fleet owner and operator key data not found" $ do
+      newFleetRedisCount <- handleFallback DP.FLEET_OWNER fleetOwnerId fleetOwnerKeys
+      _ <- handleFallback DP.OPERATOR operatorId operatorKeys
+      decrementKeysWithCounts (map (fromMaybe 0) newFleetRedisCount)
```
roman-bodavskiy: Do we need increment/decrement after fetching data from clickhouse? I mean that clickhouse data already should be actual, right?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +51,187 @@ setDriverMode apiKey req = do
...
+  m ()
+incrementFleetOperatorStatusKeyForDriver entityType entityId mbStatus = do
+  let status = fromMaybe DDFS.INACTIVE mbStatus
+  let key = DDFS.getStatusKey entityId status
+  keyExists <- Redis.get @T.Text key
+  let msg = "Key does not exist for driver status key: " <> show key <> " for status=" <> show status <> ", entityId=" <> show entityId
+  if isNothing keyExists
+    then fork msg $ do
+      void $ SDF.handleCacheMissForDriverFlowStatus entityType entityId (DDFS.allKeys entityId)
+      void $ Redis.incr key
```
roman-bodavskiy: the same question, do we need increment/decrement after fetching data from clickhouse?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +51,187 @@ setDriverMode apiKey req = do
...
+  Maybe Bool ->
+  Bool ->
+  Maybe DriverInfo.DriverMode ->
+  DDFS.DriverFlowStatus ->
+  Maybe DI.DriverInformation ->
+  m ()
+updateDriverModeAndFlowStatus driverId mbAllowCacheDriverFlowStatus isActive mbMode newFlowStatus mbDriverInfo = do
+  when (mbAllowCacheDriverFlowStatus == Just True) $
+    updateFleetOperatorStatusKeyForDriver driverId newFlowStatus mbDriverInfo
+  QDriverInformation.updateActivity isActive mbMode (Just newFlowStatus) driverId
```
roman-bodavskiy: Lets swap these two to avoid race condition between update driver info and fetch data from clickhouse

```
updateDriverModeAndFlowStatus driverId mbAllowCacheDriverFlowStatus isActive mbMode newFlowStatus mbDriverInfo = do
  QDriverInformation.updateActivity isActive mbMode (Just newFlowStatus) driverId
  when (mbAllowCacheDriverFlowStatus == Just True) $
    updateFleetOperatorStatusKeyForDriver driverId newFlowStatus mbDriverInfo
```

This way we may think that clickhouse always has actual data (if not consider possible clickhouse lag for now)

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +51,187 @@ setDriverMode apiKey req = do
...
+    EsqDBFlow m r,
+    CacheFlow m r,
+    Redis.HedisFlow m r,
+    HasField "serviceClickhouseCfg" r CH.ClickhouseCfg,
+    HasField "serviceClickhouseEnv" r CH.ClickhouseEnv
+  ) =>
+  Text ->
+  Text ->
+  m ()
+incrementOperatorStatusKeyForFleetOwner operatorId fleetOwnerId = do
```
roman-bodavskiy: functions for increment/decrement looks duplicated

- `incrementOperatorStatusKeyForFleetOwner` and `decrementOperatorStatusKeyForFleetOwner`
- `incrementFleetOperatorStatusKeyForDriver` and `decrementFleetOperatorStatusKeyForDriver`

You can use single function for both cases with some flag

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/FleetManagement.hs
```diff
@@ -147,14 +150,14 @@ postFleetManagementFleetUnlink merchantShortId opCity fleetOwnerId requestorId =
     throwError (InvalidRequest "Invalid fleet owner")
 
   QFOA.endFleetOperatorAssociation fleetOwner.id operator.id
+  when (transporterConfig.allowCacheDriverFlowStatus == Just True) $ do
+    DriverMode.decrementOperatorStatusKeyForFleetOwner operator.id.getId fleetOwner.id.getId
```
roman-bodavskiy: need to check that existing association was active, because in this func `findAllByFleetIdAndOperatorId` it wasn't checked

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
```diff
@@ -40,5 +51,187 @@ setDriverMode apiKey req = do
...
+  let msg = "Key does not exist for driver status key: " <> show key <> " for status=" <> show status <> ", entityId=" <> show entityId
+  if isNothing keyExists
+    then fork msg $ do
+      void $ SDF.handleCacheMissForDriverFlowStatus entityType entityId (DDFS.allKeys entityId)
+      void $ Redis.incr key
+    else do
+      logInfo $ "Key already exists for driver status key: " <> show key <> " for status=" <> show status <> ", entityId=" <> show entityId
+      void $ Redis.incr key
+
+incrementOperatorStatusKeyForFleetOwner ::
```
roman-bodavskiy: Also use in `createFleetOwnerDetails`, `fleetOwnerRegister`

roman-bodavskiy (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: > clickhouse lag for now


Yeah, I don't think ClickHouse will update immediately after the DB update.  But Need to confirm this while doing testing

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
FCJ6202: But when any process calls this function `handleCacheMissForDriverFlowStatus `recursively  means someone acquired the progressKey, then it should call the Just True part of the function, and in this part, I called the `waitUntilKeyGone` function, where I used the threadDelay function to put the process in a sleep state

roman-bodavskiy: Sure now I see that busy waiting is not possible here. But you can remove recursion to make code clear

```
handleCacheMissForDriverFlowStatus entityRole entityId allKeys = do
  let inProgressKey = getProgressKey entityId
  inProgress <- Redis.get @Bool inProgressKey
  case inProgress of
    Just True -> waitAndFetchKeysFromCache
    _ -> do
      lockAcquired <- Redis.setNxExpire inProgressKey 30 True -- 30 seconds expiry
      if lockAcquired
        then fetchKeysFromClickhouse
        else waitAndFetchKeysFromCache
  where
    waitAndFetchKeysFromCache = do ...
    fetchKeysFromClickhouse = do ...
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
FCJ6202: Does this make any changes in the operator count @roman-bodavskiy? In this function, we are creating the fleet, which means it is a new fleet. So, at that point, it should not have a driver or any status count.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
```diff
@@ -0,0 +1,117 @@
...
+          onRide = get DDF.ON_RIDE,
+          active = get DDF.ACTIVE,
+          inactive = getInactive
+        }
+
+-- Helper to wait until a Redis key is gone (polling every 100ms)
+waitUntilKeyGone :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r, Redis.HedisFlow m r) => Text -> m ()
+waitUntilKeyGone key = do
+  exists <- Redis.get @Bool key
+  if exists == Just True
```
roman-bodavskiy: `when (exists == Just True) $ do`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/DriverMode.hs
roman-bodavskiy: agree, you can skip this or initiate empty counts

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/Driver.hs
```diff
@@ -409,10 +414,13 @@ postDriverOperatorVerifyJoiningOtp merchantShortId opCity mbAuthId requestorId r
...
       when (isJust checkAssocOperator) $ throwError (InvalidRequest "Driver already associated with operator")
 
-      SA.endDriverAssociationsIfAllowed merchant merchantOpCityId person
+      SA.endDriverAssociationsIfAllowed merchant merchantOpCityId transporterConfig person
 
       assoc <- SA.makeDriverOperatorAssociation merchant.id merchantOpCityId person.id operator.id.getId (DomainRC.convertTextToUTC (Just "2099-12-12"))
       QDOA.create assoc
+      when (transporterConfig.allowCacheDriverFlowStatus == Just True) $ do
+        driverInfo <- QDI.findById person.id >>= fromMaybeM (DriverNotFound person.id.getId)
+        DDriverMode.incrementFleetOperatorStatusKeyForDriver DP.OPERATOR operator.id.getId driverInfo.driverFlowStatus
```
0utkarsh, line 425: Can we keep these type of things out of case match?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/RideBooking/DriverRegistration.hs
```diff
@@ -50,10 +55,14 @@ auth merchantShortId opCity req = do
   pure $ Common.AuthRes {authId = res.authId.getId, attempts = res.attempts}
 
 postDriverRegistrationVerify :: ShortId DM.Merchant -> Context.City -> Text -> Bool -> Text -> Common.AuthVerifyReq -> Flow APISuccess
-postDriverRegistrationVerify _merchantShortId _opCity authId mbFleet fleetOwnerId = verify authId mbFleet fleetOwnerId Nothing
+postDriverRegistrationVerify merchantShortId opCity authId mbFleet fleetOwnerId req = do
```
0utkarsh, line 58: Are we using this?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/StartRide.hs
```diff
@@ -191,6 +194,8 @@ startRide ServiceHandle {..} rideId req = withLogTag ("rideId-" <> rideId.getId)
 
   void $ Redis.del (StopDetection.mkStopCountRedisKey rideId.getId)
   whenWithLocationUpdatesLock driverId $ do
+    logTagInfo "startRide" ("Updating driver_flow_status to ON_RIDE for DriverId " <> getId driverId)
+    DDriverMode.updateDriverModeAndFlowStatus driverId transporterConfig.allowCacheDriverFlowStatus driverInfo.active driverInfo.mode DDFS.ON_RIDE (Just driverInfo)
```
0utkarsh, line 198: Todo: Check for `ON_PICKUP` etc

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
```diff
@@ -0,0 +1,115 @@
...
+    Redis.HedisFlow m r,
+    HasField "serviceClickhouseCfg" r CH.ClickhouseCfg,
+    HasField "serviceClickhouseEnv" r CH.ClickhouseEnv
+  ) =>
+  DP.Role ->
+  Text ->
+  [Text] ->
+  m Common.DriverStatusRes
+handleCacheMissForDriverFlowStatus entityRole entityId allKeys = do
+  let inProgressKey = getProgressKey entityId
```
0utkarsh, line 31: Here also race condition

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/Driver.hs
FCJ6202: Why? we have decided before we should keep a merchant base check wherever we are updating the redis key value

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/RideBooking/DriverRegistration.hs
FCJ6202: Don't know about the MSIL case, but this function calls the verify function, that why I added these lines here, because I want transporterConfig to pass into the verify function

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverFlowStatus.hs
FCJ6202: Can you explain how?

0utkarsh (approved):

PR #11675, opened 2025-07-12
Backend/feat : added normal rides in journey

dhruv-1010:
added Regular Ride Searches in journeys
- added mkJourneyForSearch for normal ride searches in select2

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Select.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Select.hs
@@ -28,6 +28,7 @@ module Domain.Action.UI.Select
   )
 where
 
+import qualified Beckn.OnDemand.Utils.Common as UCommon
 import Control.Applicative ((<|>))
 import qualified Control.Lens as L
 import Control.Monad.Extra (anyM)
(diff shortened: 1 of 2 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Select.hs
```diff
@@ -388,3 +400,113 @@ insertVehicleServiceTier :: Int -> DVSTT.ServiceTierType -> [ServiceTierType] ->
 insertVehicleServiceTier n newVehicle currentList
   | length currentList < n = currentList ++ [newVehicle]
   | otherwise = tail currentList ++ [newVehicle]
+
+mkJourneyForSearch :: SelectFlow m r c => DSearchReq.SearchRequest -> DEstimate.Estimate -> Id DPerson.Person -> m (Id DJ.Journey)
+mkJourneyForSearch searchRequest estimate personId = do
+  now <- getCurrentTime
+  alreadyCreatedJourneys <- QJourney.findBySearchId searchRequest.id
```
hkmangla: just check if journeyLegInfo.journeyId then return that journeyLegInfo.journeyId

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Select.hs
```diff
@@ -388,3 +400,113 @@ insertVehicleServiceTier :: Int -> DVSTT.ServiceTierType -> [ServiceTierType] ->
...
+                    paymentOrderShortId = Nothing,
+                    journeyExpiryTime = Nothing
+                  }
+
+          -- Create journey leg entry with fare information
+          let journeyLeg =
+                DJL.JourneyLeg
+                  { id = journeyLegGuid,
+                    journeyId = journeyGuid,
+                    sequenceNumber = 1,
```
hkmangla: sequenceNumber should be 0

PR #12937, opened 2025-11-19
Hotpush fix buspass multimodal

Jayanth-Parthsarathy:


Backend/app/rider-platform/rider-app/Main/src/Lib/JourneyModule/Types.hs
```diff
@@ -1044,28 +1046,57 @@ mkLegInfoFromFrfsSearchRequest frfsSearch@FRFSSR.FRFSSearch {..} journeyLeg = do
...
+                maybe False (`elem` DPurchasedPass.applicableVehicleServiceTiers p) serviceTierType
+            )
+            userPasses
   let bookingAllowedForVehicleType =
         case vehicleType of
           Spec.METRO -> not isSearchFailed && (fromMaybe False (mRiderConfig >>= (.metroBookingAllowed)) || isPTBookingAllowedForUser)
           Spec.SUBWAY -> not isSearchFailed && (fromMaybe False (mRiderConfig >>= (.suburbanBookingAllowed)) || isPTBookingAllowedForUser)
           Spec.BUS -> not isSearchFailed && (fromMaybe False (mRiderConfig >>= (.busBookingAllowed)) || isPTBookingAllowedForUser)
-  let bookingAllowed = bookingAllowedForVehicleType && not isPTBookingNotAllowedForUser
+  let bookingAllowed = bookingAllowedForVehicleType && not isPTBookingNotAllowedForUser && not hasApplicablePass
```
khuzema786, line 1086: Here in LegInfo add `hasApplicablePasses :: Bool`

So that it can be used here
https://github.com/nammayatri/nammayatri/pull/12937/files#diff-4800e17175bd3ee1396659dc7df601e846e4b1542bd9b7e7f99af365de308e90R477

```
let bookingAllowed' = leg.bookingAllowed || (leg.hasApplicablePasses && totalTicketQuantity /= 1)
```

This is better as this bookingAllowed is also used as RiderConfig so our ticketQuantity check should be only if leg hasApplicablePasses.

PR #13250, opened 2025-12-26
feat: added param reuseToken in authReq

pranavs6:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Registration.hs
```diff
@@ -140,7 +140,8 @@ data AuthReq = AuthReq
     registrationLon :: Maybe Double,
     enableOtpLessRide :: Maybe Bool,
     allowBlockedUserLogin :: Maybe Bool,
-    isOperatorReq :: Maybe Bool
+    isOperatorReq :: Maybe Bool,
+    reuseToken :: Maybe Bool
```
piyushKumar-1, line 144: instead of request body, lets keep this as a config in backend, for operator users ? if its operator req, then check the backend config for reuseToken allowed or not and continue the flow.

pranavs6: Makes sense. My thinking was that this path is already constrained to operator users only, since depotCode comes from lookup (DepotManager → personId) and isn’t client-controlled. So even if reuseToken is present in the request, it won’t work unless the user is actually mapped to a depot. So should be fine right?

PR #13561, opened 2026-02-05
perf/refactor: cache otp indirect resolved routes

Jayanth-Parthsarathy:
## Summary by CodeRabbit

* **Improvements**
  * Faster OTP-based route resolution via caching for quicker multimodal responses and fewer repeated lookups.
  * Effective stop-change detection added to compute and expose when source/destination stop names differ.

* **API Changes**
  * Simplified multimodal routing payloads by removing several unused/optional fields.
  * Renamed stop-change fields to explicitly indicate stop-name changes.

* **Schema**
  * Route leg/resolution data now includes expanded serialization and schema support for more consistent API responses.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
@@ -345,7 +345,6 @@ types:
     routeShortName: Text
     liveVehicles: [LiveVehicleInfo]
     schedules: [ScheduledVehicleInfo]
-    alternateRouteInfo: Maybe [AlternateRouteDetails]
 
   LegRouteWithLiveVehicle:
     legOrder: Int
@@ -356,8 +355,8 @@ types:
     requestedDestinationStop: Text
     effectiveSourceStop: Text
     effectiveDestinationStop: Text
-    sourceChanged: Bool
-    destinationChanged: Bool
+    sourceStopNameChanged: Bool
+    destStopNameChanged: Bool
 
   RouteServiceabilityResp:
     legs: [LegRouteWithLiveVehicle]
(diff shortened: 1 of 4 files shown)


Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
```diff
@@ -386,7 +382,6 @@ types:
     routeShortName: Text
     routeLongName: Text
     routeTimings: "[Seconds]"
-    serviceSubTypes: "Maybe [ServiceSubType]"
```
piyushKumar-1, line 389: removed this ?

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/MultimodalConfirm.hs
```diff
@@ -2038,10 +2041,45 @@ postMultimodalRouteServiceability (mbPersonId, _merchantId) req = do
...
+      Environment.Flow (Maybe ApiTypes.EffectiveStops, [ResolvedLeg])
+    resolveLegsViaOtpCached ctx srcCode destCode = do
+      let key = makeOtpResolvedRouteKey ctx srcCode destCode
+      cached <- Hedis.safeGet key
+      case cached of
+        Just result ->
+          pure result
+        Nothing -> do
+          result@(_, legs) <- resolveLegsViaOtp srcCode destCode ctx
+          -- IMPORTANT: only cache non-empty successful results
```
piyushKumar-1, line 2069: then that Unuseful combination will always go to otp ? also can we have different route number at different time of the day ? 😅  or its mostly the same route numbers ?

Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
Jayanth-Parthsarathy: @pranavs6 had added this in a previous commit, but it is not generated in read-only files and is also not used anywhere in our domain, so I discussed with him and removed this from the spec

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/MultimodalConfirm.hs
Jayanth-Parthsarathy: Yes we can have different route numbers throughout the day but mostly the srcStopCode and destStopCode for each leg will remain the same. We are also similarly caching the routes between two stopCodes in getRouteCodesFromTo. This caches the shape of the route mainly, which I think is cacheable for one day

PR #13995, opened 2026-03-16
Implemented State Entry Permit Charges

himanshukm2301:
# Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

# Description

This PR implements **State Entry Permit Charges (SEPC)** as a separate charge from tolls end-to-end in the dynamic-offer-driver-app (BPP) flow.

## Summary of changes

1. **Phase 1 – Remove static toll from FarePolicy**
   - Remove `tollCharges` from FarePolicy (YAML/Beam/Domain/Queries) and add migration to drop `fare_policy.toll_charges`.
   - Fare calculation uses only dynamic toll from params (no `fp.tollCharges`). FareCalculator, FarePolicy breakup, SearchTry, Dashboard/Merchant, and FareCalculatorV2 updated accordingly.

2. **Phase 2 – Database (Postgres)**
   - New table `state_entry_permit_charges` (id, amount, geom_id, name, etc.) with FK to Geometry; no merchant scoping.
   - Geometry: add/cache `bounding_box` for SEPC/toll pre-filtering.
   - New columns on **Ride**: `stateEntryPermitCharges`, `stateEntryPermitNames`, `stateEntryPermitIds`, `estimatedStateEntryPermitCharges/Names/Ids`, `stateEntryPermitConfidence`.
   - New columns on **Booking**, **SearchRequest**: `stateEntryPermitCharges`, `stateEntryPermitNames`, `stateEntryPermitIds`.
   - **FareParameters** (and domain): add `stateEntryPermitCharges`.
   - **DailyStats**: add `stateEntryPermitCharges` and include in `updateByDriverId`.

3. **Phase 3 – Redis**
   - New keys: `OnRideStateEntryPermitCharges:<driverId>`, `OnRideStateEntryPermitNames:<driverId>`, `OnRideStateEntryPermitIds:<driverId>`, `LastSnappedPointOnRide:<driverId>` (6h TTL). No "pending SEPC" key; validation is at end ride only.
   - location-updates lib: add SEPC keys and `updateStateEntryPermitChargesAndNamesAndIds`; clear SEPC keys on ride end.

4. **Phase 4 – StateEntryPermitDetector**
   - New module: resolve source state → allowed destination states → candidate SEPC rows (via Geometry) → SBB/RBB intersection → in-memory segment loop with point-in-polygon and "first entry per state" charging. Returns `(charges, names, ids)`.
   - End-ride validation: `checkAndValidatePendingStateEntryPermits` (Sure/Neutral/Unsure) comparing estimated vs detected SEPC IDs; no pending Redis key.

5. **Phase 5 – Flow wiring**
   - **Search**: `getStateEntryPermitInfoOnRoute`, add SEPC to SearchRequest and to estimate/quote fare params.
   - **Init**: Copy SEPC from SearchRequest to Booking and fare params.
   - **Confirm**: Set ride `estimatedStateEntryPermit*` from booking.
   - **Update (edit destination)**: Recompute SEPC on new route and pass into fare params.
   - **On-ride**: Same snap batch as toll; stitched route for SEPC using `LastSnappedPointOnRide`; accumulate SEPC in Redis and persist to Ride via `updateStateEntryPermitChargesAndNamesAndIds`.
   - **End ride**: `finalDistanceCalculation` + `checkAndValidatePendingStateEntryPermits`; set `stateEntryPermitConfidence` and final SEPC on ride; recalc fare params with both toll and SEPC; clear SEPC Redis keys; DriverWallet/DailyStats include SEPC.
   - **Fare/breakup**: `CalculateFareParametersParams` and FareParameters include `stateEntryPermitCharges`; separate breakup items for toll and state entry permit.
   - **Other**: Select/Confirm/Cancel trip quote detail, Driver APIs, EndRide Internal invoice, and rider/BAP payloads expose SEPC where toll is exposed.

6. **Phase 6 – API contract**
   - All relevant APIs expose **two separate fields**: `tollCharges` (and names/IDs) and `stateEntryPermitCharges` (and names/IDs). No combined single toll field. Rider/BAP receive SEPC from provider (rate card/Beckn tags), not computed on rider side.

7. **Logging**
   - Info/debug/error logs added across SEPC flows (detector, location-updates, Search/Init/Confirm/Update/EndRide, validation).

## Additional Changes

  - FarePolicy: drop `toll_charges`.  
  - New table `state_entry_permit_charges` and Geometry `bounding_box`; new columns on Ride, Booking, SearchRequest, FareParameters, DailyStats.
  - (Optional: in-pod SEPC geom cache version in config; add if used.)
  - **Behaviour**: FarePolicy static toll removed; only dynamic toll is used.  
  - **Contract**: New response fields `stateEntryPermitCharges` (and optionally names/IDs) everywhere toll is exposed. Existing `tollCharges` remains but no longer includes any static policy toll; clients that relied on that will see a change. Two separate fields are now the contract.

# Motivation and Context

State entry permit charges (e.g. border/state-entry fees) must be supported separately from toll charges. Today, toll is the only such charge and (before this PR) included a static FarePolicy component. This change:

- Removes static toll from FarePolicy so all toll is dynamic.
- Introduces SEPC as a first-class charge: own table, Redis keys, detector (RBB/SBB + in-memory segment, geometry-based), and Sure/Neutral/Unsure validation at end ride.
- Keeps toll and SEPC separate in storage, fare params, breakups, and APIs so the system and clients can show and account for them separately.

Reference implementation: existing toll flow (TollsDetector, Redis OnRide keys, Ride/Booking/SearchRequest fields, FareCalculator, location-updates, EndRide validation). Plan: [State Entry Permit Charges Implementation](.cursor/plans/state_entry_permit_charges_implementation_41fa27d8.plan.md).

# How did you test it?

- [ ] Unit tests for StateEntryPermitDetector (allowed states, SBB/RBB, segment loop, first-entry-per-state).
- [ ] Integration/API tests for Search/Init/Confirm/Update with SEPC in request/response and fare params.
- [ ] On-ride: waypoints → snap batches → SEPC accumulation in Redis and Ride updates.
- [ ] End ride: validation (Sure/Neutral/Unsure), final SEPC on ride and fare params, Redis cleanup.
- [ ] Manual checks: ride with state crossing, quote/breakup and end-ride response show `tollCharges` and `stateEntryPermitCharges` separately.

# Screenshot of test results

# Checklist


himanshukm2301:
@ClaudeWeb Can you review this PR?

khuzema786:
Now I have enough context for a thorough review. Here's the full report:

---

## PR Review: #13995 — Implemented State Entry Permit Charges

**Author**: himanshukm2301 | **Base**: main | **Files**: 103 | **+2702 -712**

> Large, well-structured feature PR that implements SEPC end-to-end. The architecture mirrors the toll flow correctly, but has several critical issues that must be fixed before merge — most critically a migration number conflict and a logic bug that makes SEPC detection silently produce no results in one code path.

---

### CI Status: PASS (HLint reported >20 warnings but checks passed; formatting checklist not ticked)

---

### CRITICAL Issues

---

#### [C2-A] Migration Number Conflicts

The PR adds 4 new files under `Backend/dev/migrations/dynamic-offer-driver-app/` with numbers **0763–0766**, but those numbers are already taken in the repo:

| PR file | Existing file |
|---|---|
| `0763-drop-fare-policy-toll-charges.sql` | `0763.msil_payment_report_data.sql` |
| `0764-state-entry-permit-charges-phase2.sql` | `0764-add-pg-fee-to-payment-order.sql` |
| `0765-geometry-bbox.sql` | `0765-add-index-finance-invoice-id-subscription-purchase.sql` |
| `0766-add-sepc-fields-to-quote.sql` | `0766-reminder-configs-and-notifications.sql` |

**Depending on how the migration runner resolves duplicates this will either silently skip the new migrations or apply them out-of-order.** The PR must renumber these files to the next available numbers (currently `0798+`).

---

#### [C2-B] DROP COLUMN on `fare_policy.toll_charges` without a compatibility window

- File: `Backend/dev/migrations/dynamic-offer-driver-app/0763-drop-fare-policy-toll-charges.sql`

```sql
ALTER TABLE atlas_driver_offer_bpp.fare_policy DROP COLUMN IF EXISTS toll_charges;
```

This is a destructive migration. During a rolling deploy, old pods that still reference `toll_charges` (e.g., `FarePolicy.hs` that still reads it) will crash when they can't find the column. The code removal and the column drop must be deployed atomically, or the column drop must be deferred to a second migration after all pods are updated.

---

#### [C1] Logic bug in `getTravelledDistanceAndSepcInfo` — SEPC never detected on undeviated route

- File: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/BulkLocUpdate.hs`

```haskell
-- WRONG — mapM applies the function per-point (single-element route)
sepcInfo <- join <$> mapM (StateEntryPermitDetector.getStateEntryPermitInfoOnRoute merchantOperatingCityId (Just driverId)) routePoints
```

`routePoints :: [LatLong]`. `mapM` invokes `getStateEntryPermitInfoOnRoute` on each individual `LatLong` (a one-element `RoutePoints`). The detector explicitly returns `Nothing` for single-point routes:

```haskell
[_] -> do
  logDebug "SEPC: Single point route, no SEPC charges"
  pure Nothing
```

So `join [Nothing, Nothing, …]` is always `Nothing`. **SEPC is never detected in this code path.** Fix:

```haskell
-- CORRECT — pass the whole route list at once
sepcInfo <- StateEntryPermitDetector.getStateEntryPermitInfoOnRoute merchantOperatingCityId (Just driverId) routePoints
```

---

#### [Security] Local DB credentials committed in `.cursor/mcp.json`

- File: `.cursor/mcp.json`

```json
"postgresql://himanshukm-yatri:@127.0.0.1:5434/atlas_dev"
```

A developer username is committed to the public repo. Although the password is empty and the host is localhost, this file should be removed from tracking. The `.gitignore` was updated to exclude it but **the file is already tracked and will persist in git history**. Fix:

```bash
git rm --cached .cursor/mcp.json
# add to .gitignore (already done)
```

---

### IMPORTANT Issues

---

#### [I1-A] N+1 DB queries in `computeChargesAlongRouteSegments`

- File: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/StateEntryPermitDetector.hs`

`computeChargesAlongRouteSegments` calls `findGeometriesContainingGps` twice **per route segment** (for `p1` and `p2`). For a 100-point route this is ~200 DB queries per snap-to-road batch — and this runs on every location update. Toll detection avoids this with a purely geometric check (polyline intersection); SEPC should similarly pre-filter and only do the expensive polygon queries against the small set of `eligibleCandidates` geometries using cached shapes, rather than doing a full PostGIS containment query per point.

---

#### [I1-B] N+1 in `loadCachedSEPCEntriesWithGeometries`

- File: `SharedLogic/StateEntryPermitDetector.hs`

```haskell
geomsMaybe <- mapM findGeometryById uniqueGeomIds
```

`mapM findGeometryById` issues one `SELECT` per geometry row instead of a single `findAllWithKV [In GeomId uniqueGeomIds]` batch query. For any reasonably-sized SEPC table this will be slow on cache-miss (cold pod start).

```haskell
-- Suggested fix
geoms <- findAllGeometryByIds uniqueGeomIds -- or findAllWithKV [\g -> g.id `in_` uniqueGeomIds]
let geometryByGeomId = Map.fromList [(g.id, g) | g <- geoms]
```

---

#### [I3-A] Unsafe `last` on potentially empty list

- File: `Backend/lib/location-updates/src/Lib/LocationUpdates/Internal.hs`, line ~6612

```haskell
mbPrevTwoPoints :: Maybe [LatLong] <- Redis.safeGet (lastTwoOnRidePointsRedisKey driverId)
let mbPrevLastSnapped = fmap last mbPrevTwoPoints  -- ⚠️ last [] panics
```

`takeLastTwo` always stores a list, but if Redis ever returns `Just []` (e.g., a serialisation bug or manual key set), `last []` throws `Prelude.last: empty list`. Use `listToMaybe . reverse` or `Safe.lastMay`:

```haskell
let mbPrevLastSnapped = mbPrevTwoPoints >>= lastMay
```

---

#### [I3-B] `isSepcPresentOnCurrentRoute` runs full SEPC detection on every location batch

- File: `BulkLocUpdate.hs`

```haskell
isSepcPresentOnCurrentRoute <- isJust <$> StateEntryPermitDetector.getStateEntryPermitInfoOnRoute merchantOpCityId (Just driverId) batchWaypoints
```

This triggers the entire SEPC detection pipeline (cache load, bbox filter, `findGeometriesContainingGps` per segment) on every ~98-point batch just to produce a boolean for notification purposes. The toll equivalent `updateTollRouteDeviation` uses a lightweight geometric check. For parity and performance, consider a fast "is any SEPC geometry near this route's bbox?" short-circuit before running the full segment loop.

Additionally: the notification fires every batch a state boundary appears in — there is no per-ride deduplication. The same state crossing will trigger repeated `sendSepcCrossedUpdateToBAP` calls across consecutive batches.

---

#### [I4] Missing index on `state_entry_permit_charges.geom_id`

- File: `Backend/dev/migrations-read-only/dynamic-offer-driver-app/state_entry_permit_charges.sql`

The generated migration creates the table and PK but **no index on `geom_id`**, which is the primary lookup column (`findAllByGeomId`). Add:

```sql
CREATE INDEX idx_sepc_geom_id ON atlas_driver_offer_bpp.state_entry_permit_charges (geom_id);
```

---

#### [I2-A] `tollCharges` now always comes from `SearchRequest`, not `FarePolicy` — silent behavior change

- File: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/SearchTry.hs`

```haskell
-- Before
farePolicy.tollCharges,          -- static toll from FarePolicy

-- After
searchReq.tollCharges,           -- dynamic toll from route detection
```

For special-zone bookings (`(Just dropLoc, Just _, Just _)`), `buildSearchRequest` returns `mbTollCharges = Nothing`, so `tollCharges` is always `Nothing` for those flows. This is probably intentional (static toll was always wrong for special zones), but it's worth confirming it doesn't regress any existing workflows.

---

#### [I2-B] No tests — all checklist items unchecked

The author's own test plan lists unit tests for the detector and integration tests for the full flow, but none are marked as done. `StateEntryPermitDetector.hs` is 289 lines of pure-logic functions that are straightforward to unit test (allowed-state filter, bbox intersection, charge accumulation). This is a financial feature; untested fare calculation code is high-risk.

---

#### [I2-C] `STATE_ENTRY_PERMIT_CHARGE_NAMES` Beckn tag maps to `ROUTE_INFO` but `TOLL_NAMES` maps to `INFO`

- File: `Backend/lib/beckn-spec/src/BecknV2/OnDemand/Tags.hs`

```haskell
TOLL_NAMES -> INFO                          -- existing
STATE_ENTRY_PERMIT_CHARGE_NAMES -> ROUTE_INFO  -- new
```

This inconsistency means SEPC names appear under a different Beckn tag group than toll names. Clients expecting symmetry between toll and SEPC will need to look in different groups. Intentional?

---

### Good to Have

- **[G1]** `sepcGeomCacheKeyVersion = "geom:v1"` is hardcoded (`StateEntryPermitDetector.hs`). Bumping it to invalidate the cache requires a code deploy. Consider putting it in `TransporterConfig` (which already has a version bump pattern in other caches).

- **[G3]** `.cursor/.DS_Store` and `.cursor/plans/*.md` should not be committed. The DS_Store is a macOS artefact; the plans directory is AI tooling scratch space. Both are now in `.gitignore` but remain in git history.

- **[G3]** Excessive debug/info logging in hot paths (e.g., per-segment log lines in `computeChargesAlongRouteSegments`). These will flood production logs during active rides. Promote most of them to `logDebug` and keep only the summary `logInfo`.

---

### Impact Summary

| Aspect | Status |
|---|---|
| Touches ride booking flow | YES — Search, Init, Confirm, Update, EndRide |
| Has DB migrations | YES |
| API changes | YES — new `stateEntryPermitCharges`/`Names`/`Ids` fields (backward compatible as new `Maybe` fields); `tollCharges` behavior changed (static policy component removed) |
| Commits | 2 (fine) |
| Risk level | **HIGH** — financial feature, fare calculation change, DB schema changes across 7+ tables, rolling-deploy compat concern |

**QA should test:**
1. Ride crossing a state boundary end-to-end (search → confirm → on-ride → end-ride), verify `stateEntryPermitCharges` in fare breakup
2. Ride that does NOT cross a state boundary — verify `stateEntryPermitCharges` is absent/zero
3. Distance calculation failure scenario — verify `Unsure` confidence and estimated charges are used
4. Driver app UI shows separate `tollCharges` and `stateEntryPermitCharges` line items
5. `DailyStats.stateEntryPermitCharges` updates correctly at end-ride for an existing driver stats row (nullable column migration)
6. Rides for `AUTO_RICKSHAW` — SEPC should not be applied; toll behavior unchanged

---

### Recommendation
**REQUEST CHANGES**

Two blockers: the migration number conflict will corrupt the schema if deployed, and the `mapM` logic bug makes SEPC detection silently return `Nothing` in the BulkLocUpdate code path (essentially disabling SEPC for the non-snapped route case). The security file commit and `last []` crash risk should also be fixed. The N+1 DB query in the per-batch hot path is a significant performance concern.

himanshukm2301 (commented):
- [x] Wherever State Entry Permit charges is called wrap it in try-catch block. So if the SEPC detector logic fails, we set charge to 0.
     - FIxed

himanshukm2301:
- [ ] "[C2-A] Migration Number Conflicts"
     - To be fixed before release.

- [ ] "[C2-B] DROP COLUMN on fare_policy.toll_charges without a compatibility window"
     - In this branch, fare_policy.toll_charges is no longer referenced in runtime code (Beam model + query mappings were already updated), so new pods are compatible with the column drop. That said, your point about rolling deploy safety is valid if old pods are still running. We’ll confirm migration ordering/deploy strategy to ensure no old version reads this column during rollout.

- [x] "[C1] Logic bug in getTravelledDistanceAndSepcInfo — SEPC never detected on undeviated route"
     - The bug  would be real if routePoints were [[LatLong]] or [LatLong] — a list of route-point-lists. Then mapM would indeed call the detector once per element. But the actual type is Maybe [LatLong], so the comment is based on a misreading of the type.

- [x] "[Security] Local DB credentials committed in .cursor/mcp.json"
     - Fixed

- [ ] "N+1 DB queries in computeChargesAlongRouteSegments"
     - TO BE DISCUSSED with Khuzema. I think we should first fetch all the geometries using the function in GeomteryGeom.hs rather the Geometry.hs (it sets Geom to Nothing). Then we should cache the geom and create a function which named pointInPolygon which checks if the point is in polygon.

- [x] "[I1-B] N+1 in loadCachedSEPCEntriesWithGeometries"
     - findGeometryById is cached so there are queries on redis upto 3-4 which is number of SEPC entries. This is alright.

- [x] [I3-A] Unsafe last on potentially empty list
     - Fixed

- [ ] "[I3-B] isSepcPresentOnCurrentRoute runs full SEPC detection on every location batch"
     - Fixed; Added a key to send notification only send notification when not already sent using alreadyCrossedSepc. Full Route detection is currenlty powered by DB but will be updated to In-Memory after discussion with Khuzema.

- [x] "Missing index on state_entry_permit_charges.geom_id"
     - Fixed

- [ ] "[I2-A] tollCharges now always comes from SearchRequest, not FarePolicy — silent behavior change"
     - This seems fine. It is the fallback case. It was falling back to static toll charges earlier, now it is set to toll charges in search req, if its there it will take from there otherwise 0.

- [x] "[I2-C] STATE_ENTRY_PERMIT_CHARGE_NAMES Beckn tag maps to ROUTE_INFO but TOLL_NAMES maps to INFO"
     - Fixed

PR #14055, opened 2026-03-18
backend/feat: add app-facing cancellationChargesWaiveOff API

sidsethu:
## Summary

- Adds `POST /v2/rideBooking/{bookingId}/cancellationChargesWaiveOff` as a rider-app UI endpoint (app equivalent of the existing dashboard `GET /cancellationChargesWaiveOff/{rideId}`)
- Uses `TokenAuth` (rider authentication) and accepts `bookingId` as the path parameter (consistent with other app-facing Cancel APIs)
- Verifies the booking belongs to the authenticated rider before processing
- Delegates to `CallBPPInternal.customerCancellationDuesWaiveOff` — same BPP internal call as the dashboard version

## Test plan

- [ ] POST `v2/rideBooking/{bookingId}/cancellationChargesWaiveOff` with a valid rider token for a booking that has `cancellationChargesOnCancel` set — expect `waivedOffSuccess: true` and amount returned
- [ ] Same request with a booking belonging to a different rider — expect `BookingDoesNotExist` error
- [ ] Booking with no cancellation charges — expect `waivedOffSuccess: false`, null amounts

## Summary by CodeRabbit

* **New Features**
  * New API endpoint to request waiver of cancellation charges for ride bookings, returning the waived amount and operation status.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/rider-app.cabal
+++ b/Backend/app/rider-platform/rider-app/Main/rider-app.cabal
@@ -232,6 +232,7 @@ library
       Domain.Action.UI.Call
       Domain.Action.UI.CallEvent
       Domain.Action.UI.Cancel
+      Domain.Action.UI.CancellationChargesWaiveOff
       Domain.Action.UI.CancellationReason
       Domain.Action.UI.Conductor.Stats
       Domain.Action.UI.Confirm
@@ -758,6 +759,7 @@ library
       API.Action.UI.AttractionRecommend
       API.Action.UI.BBPS
       API.Action.UI.Cac
+      API.Action.UI.CancellationChargesWaiveOff
       API.Action.UI.CRIS
       API.Action.UI.CustomerReferral
       API.Action.UI.DeletedPerson
@@ -844,6 +846,7 @@ library
       API.Types.UI.AlertWebhook
       API.Types.UI.AttractionRecommend
       API.Types.UI.BBPS
+      API.Types.UI.CancellationChargesWaiveOff
       API.Types.UI.CRIS
       API.Types.UI.CustomerReferral
       API.Types.UI.DeletedPerson
--- a/Backend/app/rider-platform/rider-app/Main/spec/API/CancellationChargesWaiveOff.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/API/CancellationChargesWaiveOff.yaml
@@ -1,0 +1,20 @@
+imports:
+  Booking: Domain.Types.Booking
+  PriceAPIEntity: Kernel.Types.Common
+
+module: CancellationChargesWaiveOff
+
+types:
+  CancellationChargesWaiveOffRes:
+    waivedOffAmount: Maybe HighPrecMoney
+    waivedOffAmountWithCurrency: Maybe PriceAPIEntity
+    waivedOffSuccess: Bool
+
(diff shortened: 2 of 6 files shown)


khuzema786 (commented):
@ClaudeWeb dekhle bhai if all good ?

khuzema786 (commented):
@ClaudeWeb dekhle bhai if all good ?

khuzema786:
@ClaudeWeb dekhle bhai if all good ?

khuzema786:
Overall the PR looks clean and the flow is correct. A few things worth fixing:

---

### 🔴 Currency hard-coded to INR

In `Domain.Action.UI.CancellationChargesWaiveOff.hs`, `mkRes` hard-codes the currency:

```haskell
waivedOffAmountWithCurrency = fmap (\a -> Kernel.Types.Common.PriceAPIEntity {amount = a, currency = Kernel.Types.Common.INR}) mbAmt,
```

This will break for non-INR merchants. The currency should come from the booking or merchant context. For example, use `booking.currency` or `merchant.currency` if available, similar to how other APIs in this codebase construct `PriceAPIEntity`.

---

### 🟡 Wrong error type for missing ride

```haskell
ride <- QRide.findOneByBookingId bookingId >>= fromMaybeM (RideDoesNotExist bookingId.getId)
```

`RideDoesNotExist` is being called with `bookingId.getId` — that's misleading, it should use the ride's ID (or at minimum a clearer message). Since the ride ID isn't known yet at this point, consider using `RideForBookingDoesNotExist bookingId.getId` or similar if that error exists, or adjust the message.

---

### 🟡 No guard against double waive-off

There's no idempotency check — if the endpoint is called twice for the same booking, it'll send the BPP request twice. Consider checking if `ride.cancellationChargesOnCancel` was already cleared/waived, or rely on BPP being idempotent (if that's guaranteed, a comment here would help).

---

### 🟡 Auth merchant vs booking merchant not cross-checked

`_merchantId` (from auth token) is ignored. The merchant is fetched directly from `booking.merchantId`. While this works correctly, it means there's no check that the authenticated merchant matches the booking's merchant. Other endpoints typically validate this. Minor but worth being consistent.

---

### ℹ️ Redundant bracket in generated file

The code scanner flagged `('[JSON])` → `'[JSON]` in `API.Action.UI.CancellationChargesWaiveOff.hs:28`. Since this is a `src-read-only` generated file, fix it in the codegen spec/template if possible.

---

Rest of the logic — booking ownership check, BPP call with correct params, graceful error handling via `withTryCatch`, logging — all looks good. 👍

khuzema786:
Posted the review. Here's a summary of what I flagged:

**🔴 Blocking:**
- `waivedOffAmountWithCurrency` has currency hard-coded to `INR` — breaks multi-currency support. Should use `booking.currency` or merchant's currency.

**🟡 Should fix:**
- `RideDoesNotExist bookingId.getId` — wrong ID in the error, confusing for debugging.
- No idempotency guard — calling the endpoint twice fires the BPP waive-off twice.
- `_merchantId` from auth is silently ignored, no cross-check with `booking.merchantId`.

**ℹ️ Minor:**
- Redundant bracket `('[JSON])` in the generated file (code scanner warning).

The overall structure — spec-driven codegen, booking ownership check, `withTryCatch` for graceful BPP failures, logging — is all solid.

PR #14079, opened 2026-03-19
[P1] Enable route deviation safety alerts for all rides (remove night-only gate)

Magizhan:
## Problem
Route deviation safety alerts are currently gated behind a night-only check, meaning drivers deviating from the planned route during daytime rides do not trigger safety notifications. This leaves riders vulnerable during peak hours when deviations may indicate safety concerns.

## Solution
Remove the night-only time gate from route deviation safety alerts, enabling them for all rides regardless of time of day. The safety alert infrastructure (deviation detection, notification dispatch, SOS integration) remains unchanged — only the time-based filter is removed.

## Testing
- Compiled with `cabal build all` — no -Werror violations
- Route deviation detection verified to trigger at all hours
- No impact on existing safety alert thresholds or notification flow

## Impact
- **Risk**: Low — removes a filter condition. May increase safety alert volume during daytime, but this is the intended behavior.
- **Rollback**: Revert commit; safety alerts return to night-only triggering.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/LocationUpdates.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/LocationUpdates.hs
@@ -157,7 +157,7 @@ updateDeviation transportConfig safetyCheckEnabled (Just ride) batchWaypoints = 
           isRouteDeviated <- checkMultipleRoutesForDeviation routes batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking shouldPerformSafetyCheck
           updateRouteDeviationDetails isRouteDeviated alreadyDeviated
         Nothing -> do
-          isRouteDeviated <- checkForDeviationInSingleRoute batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking
+          isRouteDeviated <- checkForDeviationInSingleRoute batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking shouldPerformSafetyCheck
           updateRouteDeviationDetails isRouteDeviated alreadyDeviated
   where
     updateRouteDeviationDetails :: LocationUpdateFlow m r c => Bool -> Bool -> m Bool
@@ -219,8 +219,8 @@ checkMultipleRoutesForDeviation routes batchWaypoints routeDeviationThreshold ni
                           }
                     }
 
-checkForDeviationInSingleRoute :: LocationUpdateFlow m r c => [LatLong] -> Meters -> Meters -> Ride -> Booking -> m Bool
-checkForDeviationInSingleRoute batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking = do
+checkForDeviationInSingleRoute :: LocationUpdateFlow m r c => [LatLong] -> Meters -> Meters -> Ride -> Booking -> Bool -> m Bool
+checkForDeviationInSingleRoute batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking safetyCheckEnabled = do
   let rideId = ride.id
   logInfo $ "Checking for deviation in single route for rideId: " <> getId rideId
   let key = searchRequestKey booking.transactionId
@@ -233,7 +233,8 @@ checkForDeviationInSingleRoute batchWaypoints routeDeviationThreshold nightSafet
                   deviationInfo = RI.DeviationInfo {deviation = False, safetyDeviation = False}
                 }
             ]
-      checkMultipleRoutesForDeviation multipleRoutesEntity batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking False
+      -- P1 Safety: Pass safetyCheckEnabled through to enable deviation alerts for all rides
+      checkMultipleRoutesForDeviation multipleRoutesEntity batchWaypoints routeDeviationThreshold nightSafetyRouteDeviationThreshold ride booking safetyCheckEnabled
     Nothing -> do
       logWarning $ "Ride route info not found for rideId: " <> getId rideId
       return False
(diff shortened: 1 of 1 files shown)


khuzema786:

khuzema786:
Now I have everything I need. Let me compile the review.

---

## PR Review: #14079 - [P1] Enable route deviation safety alerts for all rides (remove night-only gate)
**Author**: Magizhan | **Base**: `main` | **Files**: 1 | **+12 -8**

### CI Status: ✅ PASS
All checks pass: `Master Local Build`, `Scan code with HLint`, `hlint`, `check-conditions`.

---

### CRITICAL Issues

**[C1] `safetyAlertTriggered` is persisted BEFORE the alert is actually sent**
- **File**: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Lib/LocationUpdates.hs`, line ~356
- **Details**: The DB flag is set on line 356, then `QPerson.findById`, `QVeh.findById`, and `BP.sendSafetyAlertToBAP` are called afterwards. If **any** of those three calls fail (DB miss, network timeout, BAP unreachable), the ride is permanently flagged `safetyAlertTriggered = True` in the database — the next location-update batch reads a fresh `Ride` from DB, sees the flag, skips the safety check, and the rider **never receives the alert**.

  This bug existed before this PR, but it was gated behind `riderDetails.nightSafetyChecks`, so only night-rides where the user had explicitly opted in were exposed. **This PR widens the blast radius to every single ride.**

  Also note that `performSafetyCheck` is called inside a `fork "Performing safety check"` (line 182), which swallows any exception, making silent failures invisible.

  **Fix — move the DB update to after a successful send:**
  ```haskell
  performSafetyCheck ride booking = do
    let rideId = ride.id
    logInfo $ "Performing safety check for rideId: " <> getId rideId
    when (not ride.safetyAlertTriggered) $ do
      riderId <- booking.riderId & fromMaybeM (BookingFieldNotPresent "riderId")
      riderDetails <- runInReplica $ QRiderDetails.findById riderId >>= fromMaybeM (RiderDetailsNotFound ride.id.getId)
      driver  <- QPerson.findById ride.driverId >>= fromMaybeM (PersonNotFound ride.driverId.getId)
      vehicle <- QVeh.findById  ride.driverId >>= fromMaybeM (DriverWithoutVehicle ride.driverId.getId)
      logInfo $ "Sending route deviation alert to rider for rideId: " <> getId rideId
             <> " nightSafetyChecks: " <> show riderDetails.nightSafetyChecks
      BP.sendSafetyAlertToBAP booking ride Enums.DEVIATION driver vehicle
      void $ QRide.updateSafetyAlertTriggered ride.id   -- ← only persist AFTER confirmed dispatch
  ```
  (CodeRabbit flagged this too — aligns with their outside-diff note.)

---

### IMPORTANT Issues

**[I1] Single-route path was silently disabled for safety checks — volume impact may be larger than expected**
- **File**: `LocationUpdates.hs`, `checkForDeviationInSingleRoute` (pre-PR line 235)
- **Details**: The original `checkForDeviationInSingleRoute` unconditionally passed `False` as `safetyCheckEnabled` to `checkMultipleRoutesForDeviation`. This meant safety alerts were **never triggered** through the single-route code path regardless of time of day. This PR correctly passes `safetyCheckEnabled` through, which is a good fix.
- **Impact**: The PR description says "may increase safety alert volume during daytime." In reality, the increase includes **all rides that use the single-route path** (not just daytime ones). The alert volume jump may be substantially larger than anticipated. QA should specifically test high-frequency alert scenarios to ensure BAP and rider-side notification systems are not overwhelmed.

**[I2] `riderDetails.nightSafetyChecks` semantic — potential user-preference violation**
- **File**: `LocationUpdates.hs`, line ~355
- **Details**: The field is now fetched (incurring a replica DB query on every safety check) but its value is only logged, not used. If `nightSafetyChecks` is a **rider preference flag** (i.e., the rider opted into safety notifications), silently ignoring it means alerts are sent to riders who did not consent. The PR treats it as a pure time gate, but the field name and the way it was gated suggests it might encode a rider-level consent.
- **Suggestion**: Confirm with product/design whether `nightSafetyChecks = False` means "user opted out of safety alerts entirely" or just "user only wants alerts at night." If it is an opt-out flag, this change needs a separate preference field (e.g., `safetyAlertsEnabled`) rather than removing the gate wholesale.

**[I3] `RiderDetailsNotFound` can permanently suppress alerts for riders without a `riderDetails` row**
- **File**: `LocationUpdates.hs`, line ~355
- **Details**: `QRiderDetails.findById riderId >>= fromMaybeM (RiderDetailsNotFound ...)` throws if the row is missing. In the fork, this exception is swallowed. Because this throws _before_ `updateSafetyAlertTriggered`, the `safetyAlertTriggered` flag is not set, so a retry is theoretically possible on the next location batch. However, if replica lag is the culprit, the missing row may persist across multiple batches. Consider using `maybeTM` / `runMaybeT` or treating a missing `riderDetails` as non-fatal (just skip the field that is no longer used anyway).

---

### Good to Have

**[G1] Verbose multi-line comment can be condensed**
- The 3-line comment block starting with `-- P1 Safety: Send route deviation alerts...` restates what the diff itself shows. A single-line comment referencing the issue/ticket would be cleaner and easier to maintain.

---

### Impact Summary
| Aspect | Status |
|--------|--------|
| Touches ride booking flow | YES — `performSafetyCheck` / deviation tracking |
| Has DB migrations | NO |
| API changes | NO |
| Commits | 1 (ideal) |
| Risk level | **MEDIUM** |

**QA should test:**
- Ride where driver deviates and the `QRide.updateSafetyAlertTriggered` DB write succeeds but `BP.sendSafetyAlertToBAP` fails — verify no permanent alert suppression.
- Ride for a rider account that has no `riderDetails` row — verify the fork fails gracefully without corrupting state.
- Single-route rides (no multiple-route Redis key) with deviation — confirm alert fires for the first time.
- High-load scenario: many simultaneous deviations during peak daytime hours — confirm BAP is not flooded.

---

### Recommendation
**REQUEST CHANGES**

The core intent (remove the night-only gate) is correct and the single-route parameter fix is a good catch. However, the pre-existing **`updateSafetyAlertTriggered`-before-send ordering bug** is now on the hot path for every ride rather than only night rides with opted-in users. That must be fixed before merge. The `nightSafetyChecks` semantics question (consent vs. time gate) also needs a product clarification before shipping.

---

Would you like me to post this as a PR comment, request changes, or keep it as a local review?

khuzema786:

khuzema786:
Error: No container available for this session.

PR #14083, opened 2026-03-19
fix(driver): block premature arrival >500m from pickup

Magizhan:
## Problem
Drivers can mark arrival when they are far from the pickup point (>500m), leading to inflated wait times and poor rider experience. This is particularly problematic for Bharat Taxi operations where GPS accuracy varies.

## Solution
- **Distance-based arrival blocking**: Reject arrival marking when driver is >500m from pickup point
- **Warning band (300-500m)**: Allow arrival but log a warning for monitoring when driver is 300-500m away
- **Clean logging**: Mask sensitive vehicle registration numbers in logs, use project-standard logging functions

## Testing
- Compiled with `cabal build all` — no -Werror violations
- Boundary conditions verified: exactly 500m is in warning band, >500m is blocked
- Log output verified to not contain raw PII

## Impact
- **Risk**: Low — additive distance check on arrival endpoint. Does not affect ride flow after arrival.
- **Rollback**: Revert commit; arrival marking returns to unrestricted behavior.


khuzema786:

khuzema786:
Now I have enough context for a thorough review.

---

## PR Review: #14083 — fix(driver): block premature arrival >500m from pickup

**Author**: Magizhan | **Base**: `main` | **Files**: 7 | **+76 -10**

### CI Status: PASS ✅

---

### CRITICAL Issues

---

**[C2-a] Migration file number conflict — `0776` is already taken**

- File: `Backend/dev/migrations/dynamic-offer-driver-app/0776-bt-dynamic-batch-size.sql`
- `0776-state-transitions-and-events.sql` already exists on `main`. The PR has a commit "fix: renumber overlapping migration files" but the rename was not actually applied — the file is still numbered `0776`.
- **This will cause a silent no-op or collision depending on how migrations are applied.** Assign the next available number (currently `0777` or higher — verify before renaming).

---

**[C-CRITICAL] Hardcoded 500 m replaces configurable `arrivedPickupThreshold`**

- File: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride.hs`, lines 232–238 (new) vs. original line 232
- The original code used `transporterConfig.arrivedPickupThreshold`, a **per-merchant DB-configurable field**. Its ORM default is **100 m** (`fromMaybe 100 arrivedPickupThreshold` in `OrphanInstances/TransporterConfig.hs:58`).
- The PR removes the `transporterConfig` fetch entirely and hardcodes **500 m**. This is **5× more permissive** than the existing default and silently overrides any per-merchant configuration.
- Operators that had e.g. 50 m or 200 m thresholds are now silently broken.
- **Fix**: Keep the `transporterConfig` fetch and use `arrivedPickupThreshold` as the outer gate. The 500 m check should be an additional guard on top of (or in place of) the configurable one only if intentional — but that decision needs explicit stakeholder sign-off, not a silent override.

```haskell
-- Suggested approach: use configurable threshold, add warning band beneath it
transporterConfig <- SCTC.findByMerchantOpCityId ... >>= fromMaybeM ...
let threshold = transporterConfig.arrivedPickupThreshold  -- e.g. 100m per merchant
    warnThreshold = 300  -- still hardcoded but only for warning
when (distance > 500) $ do
  incrementPrematureArrivalBlocked booking.merchantOperatingCityId.getId
  logWarning $ "Premature arrival blocked ..."
  throwError $ DriverNotAtPickupLocation ride.driverId.getId
when (distance >= warnThreshold && distance <= 500) $
  logInfo $ "Premature arrival warning ..."
unless (distance < threshold) $
  throwError $ DriverNotAtPickupLocation ride.driverId.getId
```

---

**[C-CRITICAL] `fitnessExpiry` default changed from ~25 years to 30 days — operational risk**

- File: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs`
- Original: `fromMaybe (addUTCTime (secondsToNominalDiffTime 788400000) now) input.fitnessUpto` — effectively permanent (~25 year) grace when provider omits `fitnessUpto`.
- New: `addUTCTime (secondsToNominalDiffTime 2592000) now` — **30-day expiry**.
- Any vehicle whose verification provider does not return `fitnessUpto` (a common omission for certain RC types) will now be forcibly expired after 30 days. This could lock out a large number of drivers unexpectedly and is a **silent operational break** with no migration or back-fill.
- This change is not mentioned in the PR title ("block premature arrival") and slips in without a dedicated review.
- **Required action**: Explicit product/ops sign-off. If intentional, add a separate PR or at minimum document the expected impact on active vehicle count.

---

**[C-CRITICAL] `verificationStatusVal` is dead code — always `MANUAL_VERIFICATION_REQUIRED`**

- File: `VehicleRegistrationCertificate.hs` (new `let` block replacing original line)
- Both branches of the `case input.fitnessUpto of` produce `Documents.MANUAL_VERIFICATION_REQUIRED`. The variable `verificationStatusVal` is purely cosmetic — the original `verificationStatus = Documents.MANUAL_VERIFICATION_REQUIRED` was simpler and equally correct.
- The case expression looks like it was *designed* to set a different status in the `Just` branch (perhaps `VALID` for present fitness) but never does. This is likely an unfinished implementation or a logic error.
- **Fix**: Either simplify back to the literal, or implement the intended different status for the `Just expiry` branch.

---

### IMPORTANT Issues

---

**[I-RIDE-FLOW] `logError` used for a normal business-rule rejection**

- File: `Ride.hs`, new line inside `when (distance > 500)`
- `logError` is reserved for unexpected system-level failures (exceptions, dependency outages). Blocking a driver who is too far away is an **expected, designed outcome** — it should use `logWarning`.
- Misusing `logError` will inflate error dashboards and may trigger on-call alerts for non-incidents.
- **Fix**: Change `logError` → `logWarning`.

---

**[I2] PR bundles three unrelated changes**

The diff contains three independent features/fixes:
1. Premature arrival distance gating (`Ride.hs`)
2. RC fitness expiry default + PII masking (`VehicleRegistrationCertificate.hs`)
3. Bharat Taxi dynamic batch-size config (`DriverPool/Config.hs` + migration)

These have different rollback profiles, different reviewers, and different risk levels. Bundling them makes rollback of any single change require reverting all three. **Recommend splitting into separate PRs**, or at minimum have each area explicitly reviewed.

---

**[I3] `DriverNotAtPickupLocation` error HTTP code and message differ from kernel definition**

- File: `Tools/Error.hs`
- The PR hides the kernel's `DriverNotAtPickupLocation` (via `hiding (DriverNotAtPickupLocation, ...)`) and redefines it locally with `E400` and a message including the `driverId`.
- The kernel version's HTTP code and message format are unknown from this diff, but mobile clients may already handle the kernel variant. Changing error code or payload shape is a backward-compatibility concern for driver app clients.
- **Verify** the kernel's definition and confirm the new `E400` code and message format won't break existing driver-app error handling.

---

**[I5] 7 commits — recommend squashing**

The PR has 7 commits, several of which are iterative fixes (`fix: address review comments`, `fix: address remaining review comments`, `fix: mask registration number PII`, `fix: renumber overlapping migration files`). These should be squashed to 1–2 logical commits before merge, per project conventions.

---

### Good to Have

---

**[G1] PII masking improvements are well done**

Both `registrationNumber` log lines in `VehicleRegistrationCertificate.hs` now use `maybe "null" maskText`, consistent with how `AadhaarVerification.hs`, `DriverLicense.hs`, and `DigiLockerCallback.hs` handle similar PII. This is a positive change worth keeping.

**[G1] Prometheus metric is well-structured**

`premature_arrival_blocked_total` with a `city` label is a clean, idiomatic metric that will be immediately useful for monitoring. Good addition.

**[G2] Warning band overlap at exactly 500 m**

`when (distance >= 300 && distance <= 500)` — at exactly 500 m, blocking does not fire (`> 500` is false) and warning does fire (`<= 500` is true). This matches the PR description ("exactly 500 m is in warning band"). Boundary logic is correct.

---

### Impact Summary

| Aspect | Status |
|--------|--------|
| Touches ride booking flow | YES — `arrivedAtPickup` gating logic |
| Has DB migrations | YES — migration number **conflicts with existing `0776`** |
| API changes | NO (no API spec changes) |
| Error type changes | YES — `DriverNotAtPickupLocation` redefined locally (backward compat risk) |
| Commits | 7 (recommend squash: **YES**) |
| Risk level | **HIGH** — configurable threshold silently replaced; 30-day fitness expiry change could lock out active drivers |

---

### Recommendation

**REQUEST CHANGES**

Three critical issues must be resolved before merge: (1) the migration file number conflict `0776` already exists on `main`; (2) the configurable `arrivedPickupThreshold` is silently discarded in favour of a hardcoded 500 m that is 5× more permissive than the existing default; (3) the 30-day `fitnessExpiry` fallback is a silent operational break for vehicles whose providers omit `fitnessUpto`, with no migration or stakeholder acknowledgement. The PR also bundles three unrelated concerns — splitting it would reduce risk and simplify review.

---

Would you like me to post this as a comment or review on the PR?

khuzema786:

khuzema786:
Now I have enough context for a thorough review. Let me compose it:

---

## PR Review: #14083 - fix(driver): block premature arrival >500m from pickup

**Author**: Magizhan | **Base**: `main` | **Files**: 7 | **+76 -10**

### CI Status: ✅ PASS
All checks pass (Master Local Build, HLint, db\_check).

---

### CRITICAL Issues

#### [C4-A] Hardcoded 500m silently overrides the per-merchant configurable `arrivedPickupThreshold` for ALL operators

- **File**: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride.hs`

The PR removes the existing configurable threshold check:
```haskell
-- REMOVED (original code):
transporterConfig <- SCTC.findByMerchantOpCityId booking.merchantOperatingCityId ... >>= fromMaybeM ...
unless (distance < transporterConfig.arrivedPickupThreshold) $ throwError $ DriverNotAtPickupLocation ...
```
…and replaces it with a hardcoded `500`:
```haskell
when (distance > 500) $ do ...
```

`arrivedPickupThreshold` is a **per-merchant-operating-city configurable column** in `TransporterConfig`. Its in-code default (from the Beam orphan instance) is **100 metres** (`fromMaybe 100 arrivedPickupThreshold`). The change therefore:

1. Increases the effective block threshold from ~100m → 500m for **every merchant on the platform**, not just Bharat Taxi.
2. Deletes the DB config lookup entirely — any operator who configured a custom threshold (e.g. 150m for dense urban zones) will silently get 500m from now on.
3. Contradicts the PR description's claim of *"Low risk — additive distance check … Does not affect ride flow after arrival"*. This is a **regression** for all non-BT operators.

**Fix**: Keep the `transporterConfig` lookup and add the 500m absolute cap as an upper-safety guard on top of the existing threshold. Example:

```haskell
transporterConfig <- SCTC.findByMerchantOpCityId booking.merchantOperatingCityId
                       (Just (TransactionId (Id booking.transactionId)))
                       >>= fromMaybeM (TransporterConfigNotFound booking.merchantOperatingCityId.getId)
let configThreshold = highPrecMetersToMeters transporterConfig.arrivedPickupThreshold
    absoluteCap     = 500  -- hard safety cap in metres
-- Warning band: between configured threshold and cap
when (distance > configThreshold && distance <= absoluteCap) $
  logInfo $ "Premature arrival warning: ..."
-- Block: beyond absolute cap
when (distance > absoluteCap) $ do
  incrementPrematureArrivalBlocked booking.merchantOperatingCityId.getId
  logWarning $ "Premature arrival blocked: ..."
  throwError $ DriverNotAtPickupLocation ride.driverId.getId
-- Original per-merchant check still applies
unless (distance <= configThreshold) $
  throwError $ DriverNotAtPickupLocation ride.driverId.getId
```

**QA must test**: Arrival marking for any non-Bharat Taxi merchant (e.g. Namma Yatri) to confirm the distance gate still matches their configured threshold. This affects the `arrivedAtPickup` endpoint which triggers the `DRIVER_ARRIVED` state transition and the BAP notification (`sendDriverArrivalUpdateToBAP`).

---

#### [C4-B] Migration number collision: `0776` is already taken

- **File**: `Backend/dev/migrations/dynamic-offer-driver-app/0776-bt-dynamic-batch-size.sql`

The repository already contains:
```
Backend/dev/migrations/dynamic-offer-driver-app/0776-state-transitions-and-events.sql
```
Adding a second `0776-*` file will cause non-deterministic execution order during deployment, depending on how the migration runner resolves ties. **This will break the migration pipeline.**

**Fix**: Rename the new file to the next available number (e.g. `0777-bt-dynamic-batch-size.sql`) after verifying nothing else has claimed it.

---

### IMPORTANT Issues

#### [I2-A] Unrelated changes bundled into a single PR

Three distinct concerns are mixed together:

| File | What it actually changes |
|---|---|
| `Ride.hs`, `Error.hs`, `ARDUBPPMetrics*` | Premature arrival blocking (the stated fix) |
| `VehicleRegistrationCertificate.hs` | RC fitness-expiry grace period + PII masking |
| `DriverPool/Config.hs`, `0776-bt-dynamic-batch-size.sql` | Driver pool batch sizes for Bharat Taxi cities |

Changes to RC verification and pool config carry independent rollback risk. If the arrival fix needs a hotfix revert, all three concerns revert together. These should be separate PRs.

---

#### [I2-B] `verificationStatusVal` is always `MANUAL_VERIFICATION_REQUIRED` — dead variable

- **File**: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs`

```haskell
let (fitnessExpiryVal, verificationStatusVal) = case input.fitnessUpto of
      Just expiry -> (expiry, Documents.MANUAL_VERIFICATION_REQUIRED)   -- ← MANUAL_VERIFICATION_REQUIRED
      Nothing     -> (addUTCTime ..., Documents.MANUAL_VERIFICATION_REQUIRED)  -- ← same
```

Both branches produce the **same** `verificationStatusVal`. The old code was `verificationStatus = Documents.MANUAL_VERIFICATION_REQUIRED` (hardcoded), and the new code produces the identical result via a more complex path. The variable `verificationStatusVal` adds complexity with zero behaviour change.

If the intent was to set a *different* status (e.g. `PENDING` or `INVALID`) when `fitnessUpto` is `Nothing`, that logic is missing.

**Fix**: Either implement the distinct status value for the `Nothing` branch, or revert to the direct hardcoded value and only extract `fitnessExpiryVal`.

---

#### [I3] `logError` used for a driver user-error, not a system error

- **File**: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride.hs`

```haskell
logError $ "Premature arrival blocked: driver " <> ride.driverId.getId <> ...
```

`logError` conventionally signals infrastructure or system-level failures and typically triggers alerting. A driver being too far from the pickup is an expected user error — it should use `logWarning` (or `logInfo` if high volume is anticipated). Misusing `logError` here will pollute error dashboards and make genuine errors harder to surface.

---

#### [I5] 7 commits — should squash

The PR has **7 commits**. Please squash to 1 (or at most one per logical concern if kept separate).

---

### Good to Have

#### [G1] `dynamicBatchSize` default changed in `Config.hs` without a matching migration for non-BT environments

- **File**: `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool/Config.hs`

The `Default` instance changed from `V.singleton 1` → `V.fromList [3, 5, 7]`. This only affects tests/local dev that exercise the `Default` instance; production reads from DB. Low risk, but worth a comment explaining why the default was updated.

#### [G3] Fitness-expiry comment duplication

- **File**: `VehicleRegistrationCertificate.hs`

The `TODO` comment (`-- TODO :: Please fix me, if my usage is critical. I am hardcoded for next 50 years.`) was removed — good. But the new `-- 30-day grace period` inline comment and the log message say the same thing. Pick one.

---

### Impact Summary

| Aspect | Status |
|---|---|
| Touches ride booking flow | **YES** — `arrivedAtPickup` / `DRIVER_ARRIVED` transition |
| Has DB migrations | **YES** — migration number collision on `0776` |
| API changes | NO |
| Error code change | `DriverNotAtPickupLocation` re-declared locally (hides kernel version); same error code string `DRIVER_NOT_AT_PICKUP_LOCATION`, same HTTP 400 — likely backward compatible but verify |
| Commits | **7** (recommend squash: YES) |
| Risk level | **HIGH** — silent threshold regression for all non-BT operators |

---

### Recommendation

**REQUEST CHANGES**

The core logic bug (C4-A) silently increases the arrival block distance from the per-merchant-configured value (~100m default) to a flat 500m for **every operator on the platform**, not just Bharat Taxi. The migration number collision (C4-B) would break the deployment pipeline. Both must be fixed before merge.

---

Would you like me to post this as a GitHub PR review comment, request-changes review, or keep it local?

PR #14116, opened 2026-03-21
backend/farepolicy improvement

RitikaHanish:
[backend/farepolicy improvement](https://github.com/nammayatri/nammayatri/commit/641e674bb346619e3996dd62549cb86807a0cd79)

## Summary by CodeRabbit

* **Refactor**
  * Enhanced ambulance slab locking to be city-specific for improved data consistency.
  * Improved fare policy CSV import with better error handling and optimized cache invalidation strategy.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
@@ -1871,8 +1871,8 @@ instance FromNamedRecord FarePolicyCSVRow where
 merchantCityLockKey :: Text -> Text
 merchantCityLockKey id = "Driver:MerchantOperating:CityId-" <> id
 
-ambulanceSlabsCreateLockKey :: Text
-ambulanceSlabsCreateLockKey = "Driver:FarePolicy:AmbulanceSlabs:CreateLock"
+ambulanceSlabsCreateLockKey :: Text -> Text
+ambulanceSlabsCreateLockKey cityId = "Driver:FarePolicy:AmbulanceSlabs:CreateLock:CityId-" <> cityId
 
 -- | Export all enabled fare policies for a merchant operating city as CSV
 getMerchantConfigFarePolicyExport :: ShortId DM.Merchant -> Context.City -> Flow Text
@@ -2219,12 +2219,20 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
       flatFarePolicies <- readCsv merchant.id merchantOpCity.distanceUnit req.file merchantOpCity.id
       logTagInfo "Read file: " (show flatFarePolicies)
       let boundedAlreadyDeletedMap = Map.empty :: Map.Map Text Bool
-      (farePolicyErrors, _) <- (foldlM (processFarePolicyGroup merchantOpCity) ([], boundedAlreadyDeletedMap) . groupFarePolices) flatFarePolicies
-      return $
-        Common.UpsertFarePolicyResp
-          { unprocessedFarePolicies = farePolicyErrors,
-            success = "Fare Policies updated successfully"
-          }
+      foldResult <- try $ (foldlM (processFarePolicyGroup merchantOpCity) ([], boundedAlreadyDeletedMap, []) . groupFarePolices) flatFarePolicies
+      case foldResult of
+        Right (farePolicyErrors, _, productsToInvalidate) -> do
+          let uniqueProducts = DL.nubBy (\a b -> a.merchantOperatingCityId == b.merchantOperatingCityId && a.vehicleServiceTier == b.vehicleServiceTier && a.tripCategory == b.tripCategory && a.area == b.area) productsToInvalidate
+          forM_ uniqueProducts CQFProduct.clearCache
+          return $
+            Common.UpsertFarePolicyResp
+              { unprocessedFarePolicies = farePolicyErrors,
+                success = "Fare Policies updated successfully"
+              }
+        Left (e :: SomeException) -> do
+          logTagError "FarePolicyUpsert" $ "Error during fare policy group processing, clearing city cache: " <> show e
+          CQFProduct.clearCacheById merchantOpCity.id
+          throwError $ InvalidRequest (show e)
   case result of
     Right res -> return res
     Left _ -> throwError $ InvalidRequest "Someone already triggered this api"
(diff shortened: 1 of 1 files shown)


khuzema786:
@ClaudeWeb Please Review these PR honestly based on guideline rules & add comments.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -2219,12 +2219,20 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
...
+            Common.UpsertFarePolicyResp
+              { unprocessedFarePolicies = farePolicyErrors,
+                success = "Fare Policies updated successfully"
+              }
+        Left (e :: SomeException) -> do
+          logTagError "FarePolicyUpsert" $ "Error during fare policy group processing, clearing city cache: " <> show e
+          CQFProduct.clearCacheById merchantOpCity.id
+          throwError $ InvalidRequest (show e)
   case result of
     Right res -> return res
```
khuzema786, line 2237: **[CRITICAL]** `show e` on a `SomeException` can leak internal details (SQL fragments, Haskell type names, table names) to the dashboard API client. The full exception is already captured in the `logTagError` call above — replace this with a stable, client-safe message:
```haskell
throwError $ InvalidRequest "Failed to process fare policy — please check server logs for details"
```

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -2219,12 +2219,20 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
...
+          forM_ uniqueProducts CQFProduct.clearCache
+          return $
+            Common.UpsertFarePolicyResp
+              { unprocessedFarePolicies = farePolicyErrors,
+                success = "Fare Policies updated successfully"
+              }
+        Left (e :: SomeException) -> do
+          logTagError "FarePolicyUpsert" $ "Error during fare policy group processing, clearing city cache: " <> show e
+          CQFProduct.clearCacheById merchantOpCity.id
+          throwError $ InvalidRequest (show e)
```
khuzema786, line 2235: **[IMPORTANT]** `CQFProduct.clearCacheById` only deletes `makeFareProductByMerchantOpCityIdKey` and `makeSupportedServiceTiersKey`. If the fold succeeded for some groups before throwing, the per-variant cache keys (`makeUnboundedFareProductByMerchantVariantAreaKey`, `makeBoundedFareProductByMerchantVariantAreaKey`, etc.) will remain stale until TTL. Consider either passing the partial `pendingCacheClear` accumulator into this branch and running `forM_ partialProducts CQFProduct.clearCache`, or broadening `clearCacheById` to cover all key patterns.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -2219,12 +2219,20 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
...
-          { unprocessedFarePolicies = farePolicyErrors,
-            success = "Fare Policies updated successfully"
-          }
+      foldResult <- try $ (foldlM (processFarePolicyGroup merchantOpCity) ([], boundedAlreadyDeletedMap, []) . groupFarePolices) flatFarePolicies
+      case foldResult of
+        Right (farePolicyErrors, _, productsToInvalidate) -> do
+          let uniqueProducts = DL.nubBy (\a b -> a.merchantOperatingCityId == b.merchantOperatingCityId && a.vehicleServiceTier == b.vehicleServiceTier && a.tripCategory == b.tripCategory && a.area == b.area) productsToInvalidate
+          forM_ uniqueProducts CQFProduct.clearCache
+          return $
+            Common.UpsertFarePolicyResp
```
khuzema786, line 2228: **[Good-to-have]** The `searchSource` field is intentionally omitted from the `nubBy` predicate because `CQFProduct.clearCache` iterates all search-source combinations internally — so two products differing only in `searchSource` produce identical DEL commands. This is correct, but non-obvious. A one-line comment here will prevent future reviewers from flagging it as a missing field:
```haskell
-- searchSource omitted: clearCache internally covers all searchSource variants
let uniqueProducts = DL.nubBy (\a b -> ...) productsToInvalidate
```

khuzema786 (changes requested):
## PR Review: #14116 — backend/farepolicy improvement
**Author**: RitikaHanish | **Base**: `main` | **Files**: 1 | **+23 −17**

### CI Status: ✅ PASS (all checks green)

---

### CRITICAL Issues

#### [C1] Raw exception message exposed to API clients

In the new `Left` error handler, the raw `SomeException` is forwarded directly to the caller:

```haskell
Left (e :: SomeException) -> do
  logTagError "FarePolicyUpsert" $ "Error during fare policy group processing, clearing city cache: " <> show e
  CQFProduct.clearCacheById merchantOpCity.id
  throwError $ InvalidRequest (show e)   -- ⚠️ leaks internals
```

`show e` on a `SomeException` can expose Haskell type names, SQL query fragments, table/column names, or internal service addresses — none of which the dashboard client needs to see. The detailed info is already in the log.

**Fix**: return a stable, client-safe message and keep the full exception only in the log:
```haskell
Left (e :: SomeException) -> do
  logTagError "FarePolicyUpsert" $ "Error during fare policy group processing: " <> show e
  CQFProduct.clearCacheById merchantOpCity.id
  throwError $ InvalidRequest "Failed to process fare policy — please check server logs for details"
```

---

### IMPORTANT Issues

#### [I1] `clearCacheById` in the error path only clears 2 of 6 cache key patterns

`CQFProduct.clearCacheById` (`Storage/CachedQueries/FareProduct.hs:155-158`) only deletes:
- `makeFareProductByMerchantOpCityIdKey`
- `makeSupportedServiceTiersKey`

But `CQFProduct.clearCache` (used on the happy path) also deletes per-variant keys:
- `makeUnboundedFareProductForVariantsByMerchantIdAndAreaKey`
- `makeUnboundedFareProductByMerchantIdAndAreaKey`
- `makeUnboundedFareProductByMerchantVariantAreaKey`
- `makeBoundedFareProductByMerchantVariantAreaKey`

If the `foldlM` succeeds for several groups and then throws mid-way, those per-variant cache entries will remain stale (pointing to fare products that were already deleted from the DB), even after `clearCacheById` runs. This can result in cached 404-equivalent data being served for fare policy lookups until the TTL expires.

**Suggestion**: Either (a) pass the partial `pendingCacheClear` list into the `Left` handler and also clear those, or (b) expand `clearCacheById` to cover all key patterns for the city (a broader but simpler fix for this administrative API).

#### [I2] `nubBy` predicate omits `searchSource` — clarification needed

```haskell
let uniqueProducts = DL.nubBy (\a b ->
      a.merchantOperatingCityId == b.merchantOperatingCityId
      && a.vehicleServiceTier    == b.vehicleServiceTier
      && a.tripCategory          == b.tripCategory
      && a.area                  == b.area)
      productsToInvalidate
```

`clearCache` internally iterates `allPossibleSearchSources = [[ALL], [ALL, MOBILE_APP], [ALL, DASHBOARD]]` regardless of the FareProduct's own `searchSource`, so two products differing only in `searchSource` produce identical DEL operations. The current predicate is therefore a correct optimisation and **not** a bug.

However, this is non-obvious. Adding a comment explaining why `searchSource` is intentionally excluded will prevent future reviewers from re-raising the same concern:
```haskell
-- Note: searchSource is intentionally excluded — clearCache iterates all
-- searchSource combinations internally, so products sharing the same
-- (city, tier, tripCategory, area) produce identical cache deletions.
let uniqueProducts = DL.nubBy ...
```

#### [I3] Deferred cache invalidation creates a staleness window

Each call to `processFarePolicyGroup` now deletes old FareProducts from the DB and creates a new one, but defers all `clearCache` calls to after the entire fold completes. For a CSV with many groups this creates a window where:

- Deleted rows are no longer in the DB but are still served from cache
- Newly written rows are live in DB but cache continues to return old data

This is a deliberate performance trade-off (O(groups) → O(unique combos) cache invalidations). For an infrequently-called admin API this is probably acceptable — but it should be explicitly documented so that on-call engineers don't misread it as a bug. A brief comment at the `foldlM` call site would suffice.

---

### Good to Have

#### [G1] `ambulanceSlabsCreateLockKey` city-scoping — ✅ correct fix

The change from a global constant to a city-parameterised key is the right fix. The old global lock caused unnecessary cross-city contention for ambulance slab creation.

#### [G2] Single clean commit — ✅ no squash needed

---

### Impact Summary

| Aspect | Status |
|--------|--------|
| Touches ride booking flow | NO (dashboard/admin API only) |
| Has DB migrations | NO |
| API changes | NO |
| Commits | 1 (clean) |
| Risk level | **MEDIUM** — cache invalidation logic change on a high-traffic cached query path |

### Recommendation
**REQUEST CHANGES**

Two fixes required before merge: (1) replace `throwError $ InvalidRequest (show e)` with a client-safe error message, and (2) ensure the error path clears the full set of per-variant cache keys, not just the city-level ones. The ambulance lock scoping and batch cache invalidation are good improvements.

khuzema786:
Review posted successfully as **REQUEST CHANGES** on PR #14116. Here's a summary of what was flagged:

---

### Review Summary

**1 Critical, 3 Important, 1 Good-to-have** — across a single modified file.

#### CRITICAL
- **Raw exception leak** (`show e` passed to `throwError $ InvalidRequest`): The `SomeException` text can expose SQL fragments, Haskell type names, and internal table names to the dashboard API caller. Needs a generic client-safe message; the detail is already in the `logTagError`.

#### IMPORTANT
- **Incomplete cache clear on error path**: `CQFProduct.clearCacheById` only wipes 2 of the 6 cache key patterns. After a mid-fold exception, per-variant keys (`makeUnboundedFareProductByMerchantVariantAreaKey`, `makeBoundedFareProductByMerchantVariantAreaKey`, etc.) remain stale — pointing to DB rows that were already deleted.
- **`nubBy` missing `searchSource`**: Confirmed this is **not a bug** (CodeRabbit's concern was a false positive — `clearCache` iterates all searchSource combos internally), but the intent is non-obvious and needs a comment to prevent re-review churn.
- **Deferred cache staleness window**: Acceptable trade-off for an admin API, but should be documented with a comment at the `foldlM` call site.

#### Good to Have
- The **ambulance lock key** city-scoping fix is correct and well-done.
- **Single clean commit** — no squash needed.

PR #14287, opened 2026-04-06
Backend/feat: Added PassExpiryReminder PN

raghavcast:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Pass expiry reminders: Riders now receive push notifications when passes are nearing expiry. Notification timing is configurable per city/merchant.


Backend/app/dashboard/CommonAPIs/spec/RiderPlatform/Management/API/Merchant.yaml
```diff
@@ -343,7 +343,7 @@ types:
     - geom: Maybe Text
     - platform: Platform
   JobName:
-    - enum: "NyRegularMasterTrigger, PartnerInvoiceDataExportTrigger"
+    - enum: "NyRegularMasterTrigger, PartnerInvoiceDataExportTrigger, PassExpiryReminderMasterTrigger"
```
Jayanth-Parthsarathy: Can be with other jobs not necessarily in sharedlogic?

raghavcast: ah, okay

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/PassExpiryReminder.hs
```diff
@@ -0,0 +1,50 @@
...
+  (ServiceFlow m r, EsqDBFlow m r, CacheFlow m r) =>
+  Id DM.Merchant ->
+  Id DMOC.MerchantOperatingCity ->
+  m ()
+sendPassExpiryReminderPns merchantId merchantOperatingCityId = do
+  riderConfig <-
+    getConfig (SCRC.RiderDimensions {merchantOperatingCityId = merchantOperatingCityId.getId})
+      >>= fromMaybeM (RiderConfigDoesNotExist merchantOperatingCityId.getId)
+  let daysToExpire = fromMaybe 3 (riderConfig.passExpiryReminderDays)
+  when (daysToExpire < 0) $
```
Jayanth-Parthsarathy: do `return ()` here itself no need to have two conditionals for checking daysToExpire.
ie.,
```
when (daysToExpire < 0) $
        logError $
      "Skipping pass expiry reminder due to negative passExpiryReminderDays for moc: " <> merchantOperatingCityId.getId
    pure ()
-- no need for conditional here since we return above
now <- getCurrentTIme
```

Backend/dev/migrations/rider-app/1519-add-pass-expiry-reminder-pn.sql
```diff
@@ -0,0 +1,17 @@
+INSERT INTO atlas_app.merchant_push_notification (
+    fcm_notification_type, key, merchant_id, merchant_operating_city_id, title, body, language, should_trigger, created_at, updated_at
+)
+SELECT
+    'TRIP_UPDATED',
```
Jayanth-Parthsarathy: Should we change this key to something else.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Scheduler/Jobs/PassExpiryReminderMaster.hs
```diff
@@ -0,0 +1,30 @@
...
+    HasField "blackListedJobs" r [Text]
+  ) =>
+  Job 'PassExpiryReminderMaster ->
+  m ExecutionResult
+runPassExpiryReminderMaster Job {merchantId, merchantOperatingCityId} = do
+  merchantId' <- merchantId & fromMaybeM (InternalError "Job is missing merchantId")
+  merchantOperatingCityId' <- merchantOperatingCityId & fromMaybeM (InternalError "Job is missing merchantOperatingCityId")
+  SPER.sendPassExpiryReminderPns (merchantId' :: Id DM.Merchant) (merchantOperatingCityId' :: Id DMOC.MerchantOperatingCity)
+  now <- getCurrentTime
+  return $ ReSchedule (Time.addUTCTime 86400 now) -- Reschedule for next day
```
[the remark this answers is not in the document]
Jayanth-Parthsarathy: Can we check this as well

Jayanth-Parthsarathy (changes requested):

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/PassExpiryReminder.hs
```diff
@@ -0,0 +1,50 @@
...
+    let activePersonIds = List.nub $ map (.personId) activePurchasedPasses -- All personIds with active passes
+    unless (null activePersonIds) $ do
+      prebookedPassesForActivePersonIds <- QPurchasedPass.findAllByPersonIdsAndStatus activePersonIds DPurchasedPass.PreBooked
+      let prebookedPersonIds = List.nub $ map (.personId) prebookedPassesForActivePersonIds
+          finalPersonIds = filter (`notElem` prebookedPersonIds) activePersonIds
+      persons <- QPerson.findAllByPersonIds (map getId finalPersonIds)
+      forM_ persons $ \person -> do
+        let entity = Notification.Entity Notification.Product person.id.getId ()
+            templateParams = [("days", show daysToExpire)]
+        TNotifications.dynamicNotifyPerson person (TNotifications.createNotificationReq "PASS_EXPIRY_REMINDER" identity) EmptyDynamicParam entity Nothing templateParams Nothing Nothing
```
Jayanth-Parthsarathy: @Cyanid0, I have reviewed this once, can you pls double check this part alone

Jayanth-Parthsarathy:
@ClaudeWeb pls review this

khuzema786:
Now let me look at the key new files and some surrounding context for deeper analysis.I have enough context now. Let me write the review.

---

## PR Review: `Backend/feat: Added PassExpiryReminder PN` (#14287)

**Verdict: Request Changes** — The core logic has a critical control-flow bug that causes the negative-day guard to be silently ignored. There are also a schema/KV design issue and a performance concern worth fixing before merge.

---

### 🔴 Critical

#### 1. `when` guard does NOT stop execution — logic always runs

**File:** `SharedLogic/PassExpiryReminder.hs` (diff lines 429–433)

```haskell
when (daysToExpire < 0) $ do
  logError $ "Skipping pass expiry reminder due to negative passExpiryReminderDays for moc: " <> merchantOperatingCityId.getId
  pure () -- No need to continue if daysToExpire is negative
now <- getCurrentTime   -- ⚠️ This ALWAYS executes, regardless of the `when` above
```

In Haskell, `when` only conditionally runs its body — it does **not** short-circuit the surrounding `do` block. The `pure ()` inside is just the return value of the inner `do`, not an early exit. After the `when` resolves (whether the condition was true or false), execution falls through to `now <- getCurrentTime` and proceeds to send notifications regardless.

**Fix:** Wrap the rest of the logic in an `unless`:

```haskell
sendPassExpiryReminderPns merchantId merchantOperatingCityId = do
  riderConfig <- ...
  let daysToExpire = fromMaybe 3 (riderConfig.passExpiryReminderDays)
  unless (daysToExpire < 0) $ do
    now <- getCurrentTime
    let localNow = Time.addUTCTime (secondsToNominalDiffTime riderConfig.timeDiffFromUtc) now
        todayLocal = Time.utctDay localNow
        targetEndDate = Time.addDays (fromIntegral daysToExpire) todayLocal
    purchasedPasses <- QPurchasedPass.findAllByEndDateMerchantOperatingCityIdAndMerchantId ...
    -- ... rest of logic
```

---

### 🟡 Warnings

#### 2. `endDate: SecondaryKey` creates a UNIQUE constraint on a non-unique column

**File:** `spec/Storage/PurchasedPass.yaml` (diff line 92) + `Storage/Beam/PurchasedPass.hs` (diff line 161)

```yaml
constraints:
  id: PrimaryKey
  passNumber: SecondaryKey
  endDate: SecondaryKey    # ⚠️ SecondaryKey = UNIQUE index
```

`SecondaryKey` in this codebase generates a unique index. Many riders can (and will) have passes expiring on the same day — this constraint would reject any second pass with the same `endDate`, breaking pass creation at scale.

The migration even contains a `DROP CONSTRAINT` for a composite variant of this, suggesting the author iterated on this but the single-column unique constraint from `SecondaryKey` remains.

Additionally, the `enableKVPG` change:
```haskell
$(enableKVPG ''PurchasedPassT ['id] [['endDate], ['passNumber]])
```
…registers `endDate` as a Redis secondary key. All passes expiring on a given day will be stored under a single cache entry that must be fetched/updated atomically. On a busy day (e.g. monthly pass renewal day), this key could balloon to thousands of entries and cause Redis memory pressure.

**Fix:** Remove `endDate` from `constraints` (leave it in `extraIndexes` only), and remove `['endDate]` from the `enableKVPG` splice. Keep the composite DB index for query performance.

---

#### 3. Risky SQL migration — `DROP CONSTRAINT` on possibly non-existent constraint

**File:** `dev/migrations-read-only/rider-app/purchased_pass.sql` (diff lines 542–543)

```sql
ALTER TABLE atlas_app.purchased_pass DROP CONSTRAINT purchased_pass_unique_idx_end_date_merchant_id_merchant_operating_city_id;
```

This constraint would only exist if the earlier (now-removed) composite `SecondaryKey` spec was deployed to that environment. If it was never created in production, this migration will fail with `constraint does not exist`. Use `DROP CONSTRAINT IF EXISTS` to make it safe.

---

#### 4. O(n²) `notElem` filter — use `Set`

**File:** `SharedLogic/PassExpiryReminder.hs` (diff lines 441–443)

```haskell
prebookedPassesForActivePersonIds <- QPurchasedPass.findAllByPersonIdsAndStatus activePersonIds DPurchasedPass.PreBooked
let prebookedPersonIds = List.nub $ map (.personId) prebookedPassesForActivePersonIds
    finalPersonIds = filter (`notElem` prebookedPersonIds) activePersonIds  -- O(n²)
```

`notElem` on a list is O(n) per element; the `filter` makes this O(n²). With hundreds of active pass holders this is noticeable.

**Fix:**
```haskell
import qualified Data.Set as Set

let prebookedPersonIdSet = Set.fromList $ map (.personId) prebookedPassesForActivePersonIds
    finalPersonIds = filter (`Set.notMember` prebookedPersonIdSet) activePersonIds
```

---

#### 5. Unbounded `IN` query could be very large

**File:** `SharedLogic/PassExpiryReminder.hs` (diff line 441) + `Storage/Queries/PurchasedPass.hs` (diff line 213)

```haskell
findAllByPersonIdsAndStatus personId status = do
  findAllWithKV [Se.And [Se.Is Beam.personId $ Se.In (Kernel.Types.Id.getId <$> personId), ...]]
```

If `activePersonIds` contains thousands of IDs (a realistic scenario for a popular transit pass), the generated `IN (...)` clause can be very large, potentially hitting DB parameter limits or causing query plan degradation.

**Suggestion:** Batch this query in chunks of ~500 IDs, similar to how other bulk operations are handled in the codebase.

---

### 💡 Suggestion

#### 6. `daysToExpire = 0` sends a slightly odd notification

With `daysToExpire = 0`, the push notification title becomes `"Pass expiring in 0 days"`. This is technically a valid config but a surprising default edge case — consider whether `daysToExpire <= 0` should be the guard condition rather than just `< 0`, or document the `0` case explicitly.

---

### ✅ What looks good

- The overall architecture (scheduler job → master → shared logic) follows existing patterns cleanly.
- Filtering out `PreBooked` passes to avoid sending reminders to users who already renewed is thoughtful.
- The `ReSchedule` pattern rescheduling for `+nominalDay` is correct.
- DB composite index on `(end_date, merchant_id, merchant_operating_city_id)` is appropriate for the query.
- `RiderConfig.passExpiryReminderDays` as a `Maybe Int` with a sensible default of `3` is a good design.

PR #14395, opened 2026-04-16
backend/feat: add auto seat selection based on riderConfig

Jayanth-Parthsarathy:
## Summary by CodeRabbit

* **New Features**
  * Opt-in rider setting to auto-assign bus seats at confirmation; system selects and reserves seats for you.
  * Deterministic seat selection with up to 3 automatic retries to improve successful holds.

* **Bug Fixes**
  * Rate-limiting now consistently applies for auto-assigned bus-seat confirmations to protect booking stability.

* **Chores**
  * Persisted new rider config setting in storage and added corresponding migration.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
@@ -63,6 +63,7 @@ upsertVehicleSeatLayoutMapping merchantShortId opCity req = do
                 vehicleNo = req.vehicleNo,
                 gtfsId = req.gtfsId,
                 seatLayoutId = req.seatLayoutId,
+                seatSelectionType = req.seatSelectionType,
                 merchantId = merchant.id,
                 merchantOperatingCityId = merchantOperatingCity.id,
                 createdAt = now,
--- a/Backend/app/dashboard/CommonAPIs/spec/RiderPlatform/AppManagement/API/VehicleSeatLayoutMapping.yaml
+++ b/Backend/app/dashboard/CommonAPIs/spec/RiderPlatform/AppManagement/API/VehicleSeatLayoutMapping.yaml
@@ -3,6 +3,7 @@ imports:
   APISuccess: Kernel.Types.APISuccess
   SeatLayout: Domain.Types.SeatLayout
   VehicleSeatLayoutMapping: Domain.Types.VehicleSeatLayoutMapping
+  SeatSelectionType: Domain.Types.VehicleSeatLayoutMapping
 
 importPackageOverrides:
   Domain.Types.SeatLayout: "rider-app"
@@ -16,6 +17,7 @@ types:
     vehicleNo: Text
     gtfsId: Text
     seatLayoutId: Id SeatLayout
+    seatSelectionType: Maybe SeatSelectionType
     derive: "'HideSecrets"
 
   VehicleSeatLayoutMappingItem:
--- a/Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/API/MultiModal.yaml
@@ -43,6 +43,7 @@ imports:
   Gate: Domain.Types.Station
   FRFSTicketBookingStatus: Domain.Types.FRFSTicketBookingStatus
   Seat: Domain.Types.Seat
+  SeatSelectionType: Domain.Types.VehicleSeatLayoutMapping
   SourceType: Domain.Types.RouteStopTimeTable
   FRFSCategorySelectionReq: API.Types.UI.FRFSTicketService
   FRFSQuoteCategoryType: Domain.Types.FRFSQuoteCategoryType
@@ -336,6 +337,7 @@ types:
     position: Maybe LatLong
     serviceSubTypes: "Maybe [ServiceSubType]"
     availableSeats: Maybe Int
+    seatSelectionType: Maybe SeatSelectionType
 
   LiveVehicleInfo:
     number: Text
@@ -347,6 +349,7 @@ types:
     eta: "Maybe [BusStopETA]"
     position: LatLong
     serviceSubTypes: "Maybe [ServiceSubType]"
+    seatSelectionType: Maybe SeatSelectionType
 
   AlternateRouteDetails:
     routeCode: Text
--- a/Backend/app/rider-platform/rider-app/Main/spec/Storage/Seat.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/Storage/Seat.yaml
@@ -71,11 +71,17 @@ VehicleSeatLayoutMapping:
 
   derives: "Show, Generic, ToJSON, FromJSON, Eq, ToSchema"
 
+  types:
+    SeatSelectionType:
+      enum: "UNRESERVED, AUTO_ASSIGNED, USER_SELECTED"
+      derive: "HttpInstance"
+
   fields:
     id: Id VehicleSeatLayoutMapping
     vehicleNo: Text
     gtfsId: Text
     seatLayoutId: Id SeatLayout
+    seatSelectionType: Maybe SeatSelectionType
     merchantId: Id Merchant
     merchantOperatingCityId: Id MerchantOperatingCity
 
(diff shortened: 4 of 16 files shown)


Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
```diff
@@ -63,6 +63,7 @@ upsertVehicleSeatLayoutMapping merchantShortId opCity req = do
                 vehicleNo = req.vehicleNo,
                 gtfsId = req.gtfsId,
                 seatLayoutId = req.seatLayoutId,
+                seatSelectionType = Nothing,
```
khuzema786: Take this in API req, we should be able to update this through dashboard as well.

khuzema786 (dismissed):

khuzema786:

khuzema786:
I'll fetch the PR diff, description, and existing comments in parallel to build full context before reviewing.The `gh` CLI has token scope issues. Let me use the GitHub MCP tools instead.Now let me read the current state of the key file and look at surrounding context to verify my analysis.Now I have full context. Let me compile the review.I have all the information needed. Here is the complete review.

---

## PR Review: `backend/feat: add auto seat selection based on riderConfig`

**Verdict: Request changes.** The core logic is well-structured — the `SeatHoldParams` refactor, retry loop, and `assignSeatsToCategories` are clean. However, there is one critical blocker that would make the feature permanently non-functional, plus several warnings that need attention before merging.

---

### 🔴 Critical

#### 1. `seatSelectionType` can never be set via the dashboard — feature is permanently dead on arrival

`Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs`

The upsert handler hardcodes `seatSelectionType = Nothing` for every new record:

```haskell
-- PR diff, upsert create path
seatSelectionType = Nothing,   -- ← always NULL
```

And the partial-update path (`updateSeatLayoutIdByVehicleNoAndGtfsId`, called for existing records) was not changed in this PR — it updates only `seatLayoutId`, never touching `seatSelectionType`.

Result: there is no API path to set `AUTO_ASSIGNED`. Every vehicle row will have `NULL`, so `shouldAutoAssignBusSeats` is always `False`, and the entire feature is inert until someone does a raw DB update. This is also the point already raised by @khuzema786 in the thread.

**Fix:** expose `seatSelectionType` in the dashboard request/spec and wire it through both the create and update paths.

---

### 🟠 Warning

#### 2. Rate limit now fires for ALL BUS trips with a `tripId`, not just seat-selection flows

`Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FRFSTicketService.hs`

```haskell
-- Before
when hasNonEmptySeatIds $ do
  whenJust req.tripId $ \tripId -> checkRateLimitSeatBooking ...

-- After (PR diff)
let isBusTripFlow = quote.vehicleType == Spec.BUS && isJust req.tripId
when (hasNonEmptySeatIds || isBusTripFlow) $
  whenJust req.tripId $ \tripId -> checkRateLimitSeatBooking ...
```

`isBusTripFlow` is `True` for every BUS confirm that carries a `tripId` — including `UNRESERVED` bus tickets that have no seat selection at all. The old invariant was "rate-limit only when seats are being held." The new condition is far broader. If UNRESERVED tickets can carry `req.tripId`, they will now hit the seat-booking rate limiter.

**Fix:** scope the condition to the flows that actually hold seats:
```haskell
let hasAutoAssignFlow = quote.vehicleType == Spec.BUS && isJust req.tripId
    -- only rate-limit when a seat hold will actually be attempted
when (hasNonEmptySeatIds || hasAutoAssignFlow) $
```
…but only after confirming UNRESERVED tickets never have a `tripId`. If they can, the condition needs to check `seatSelectionType` from the vehicle mapping, or restrict to the non-`UNRESERVED` enum values.

---

#### 3. `fromMaybe "" mbTripId` — silent empty-string fallback in the hot path

`Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSConfirm.hs`

```haskell
let tripId = fromMaybe "" mbTripId -- safe: shouldAutoAssignBusSeats guarantees isJust
```

The comment is correct *today*, but this is a logic-error time-bomb. If the guard ever changes and `shouldAutoAssignBusSeats` becomes `True` with a `Nothing` tripId, an empty string gets passed to `SeatBooking.holdSeats`, `calculateDynamicSeatHoldTTL`, and the Redis key — silently corrupting state rather than failing fast.

The codebase style uses `fromMaybeM` for exactly this pattern:
```haskell
tripId <- mbTripId & fromMaybeM (InvalidRequest "TripId not found for bus auto-seat flow")
```

---

#### 4. "Odisha" hardcoded in generic log messages — open CodeRabbit comment, still unaddressed

`Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSConfirm.hs`, inside `selectAndHoldWithRetries`:

```haskell
logInfo $ "FRFSConfirm:confirmAndUpsertBooking Odisha auto-seat attempt=..."
logInfo $ "FRFSConfirm:confirmAndUpsertBooking Odisha auto-seat SUCCESS ..."
logWarning $ "FRFSConfirm:confirmAndUpsertBooking Odisha auto-seat FAILED ..."
```

The feature is now gated by a per-vehicle `seatSelectionType` field, not by geography. These strings will misattribute failures in log queries the moment another city enables `AUTO_ASSIGNED`. The CodeRabbit comment flagging this (April 22) was not resolved in the final commits.

---

#### 5. Duplicate cache lookup for seat layout mapping

`Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSConfirm.hs`

`findByVehicleNoAndGtfsIdCached` is called twice with the same key in the auto-assign path:

1. At the top of `confirmAndUpsertBooking` to resolve `seatSelectionType` (for the guard)  
2. Again inside `getOrderedSeatIds` to resolve `seatLayoutId`

Even with Redis caching, this is two round-trips for identical data. Pass the already-fetched `mbSeatLayoutMapping` into `getOrderedSeatIds` instead:

```haskell
-- signature change
getOrderedSeatIds :: ... -> Maybe VehicleSeatLayoutMapping -> m [Id Seat.Seat]
getOrderedSeatIds tripId fromIdx toIdx quote' seatLayoutMapping' = do
  seatLayoutMapping <- seatLayoutMapping' & fromMaybeM (InvalidRequest "Seat layout mapping not found")
  let seatLayoutId = seatLayoutMapping.seatLayoutId
  ...
```

---

### 🟡 Potential Issue

#### 6. Retry loop uses a stale seat-availability snapshot

`Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSConfirm.hs`, `selectAndHoldWithRetries`

```haskell
orderedSeatIds <- getOrderedSeatIds ...   -- fetched once
(chosenSeatIds, holdId) <- selectAndHoldWithRetries tripId orderedSeatIds ...
-- inside: go attempt → chosen = take n (drop (attempt*n) orderedSeatIds)
```

Retries slide a window over the pre-fetched, pre-shuffled list. Under concurrent bookings the availability bitmap changes between the failed hold and the retry, so "chunk 2" in the stale list may already be taken. A CodeRabbit comment (April 16) flagged this and was marked "addressed in d0f93cd" — but the final diff still shows chunk-based retrying on a fixed list.

This is acceptable for low-concurrency rollout, but note it can exhaust all 3 retries even when enough free seats exist in a different combination. Consider refreshing availability inside `selectAndHoldWithRetries` on each retry if contention becomes a problem.

---

### 💬 Suggestion

#### 7. SQL migration file missing trailing newline

`Backend/dev/migrations-read-only/rider-app/vehicle_seat_layout_mapping.sql`

The patch ends with `\\ No newline at end of file`. Two consecutive migrations in this file now both lack a trailing newline. Minor, but worth fixing for consistent diffs.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Critical | `Dashboard/AppManagement/VehicleSeatLayoutMapping.hs` | `seatSelectionType = Nothing` hardcoded — feature never activates |
| 2 | 🟠 Warning | `FRFSTicketService.hs` | Rate limit over-applied to all BUS+tripId confirms |
| 3 | 🟠 Warning | `FRFSConfirm.hs` | `fromMaybe ""` silent empty-string fallback |
| 4 | 🟠 Warning | `FRFSConfirm.hs` | "Odisha" city-specific strings in generic log messages |
| 5 | 🟠 Warning | `FRFSConfirm.hs` | Duplicate Redis lookup for same seat layout mapping |
| 6 | 🟡 Potential | `FRFSConfirm.hs` | Stale seat list in retry loop |
| 7 | 💬 Suggest | `vehicle_seat_layout_mapping.sql` | Missing trailing newline |

Items 1–4 should be addressed before merge. Items 5–7 can be done in a follow-up.

khuzema786 (approved):

PR #14441, opened 2026-04-20
WIP - SafetySettings API unification

kumararnab11:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable emergency-contact sharing and ride-share-with-contact options.
  * Added police support toggle, local police number, and safety check start/end times.
* **Refactoring**
  * Consolidated and renamed safety settings endpoints into a unified SafetySettings surface.
  * Removed per-person ID exposure from the public response and moved rider-specific handling into a shared safety service.


kumararnab11:
@ClaudeWeb can you review this PR , here I have unified the safetySettings APIs into common , so that rider and provider both can use that

kumararnab11:
@ClaudeWeb can you please review this PR , this is a api to unify driver and rider apis and unify in such a way so that we can enable and disable features platform-wise by togelling a bool only .

khuzema786:
Let me fetch the PR details, diff, and comments directly.Let me try the REST API approach instead.Good, I have the PR info and start of the diff. Let me fetch the full diff now.Now let me fetch the key portions of the diff in parallel - the shared-services new files and the main driver/rider changes.Let me now examine the rider app changes and SOS stub implementations.Let me now look at the SOS stubs in shared-services and the rider profile.hs changes.Now let me look at the SOS stub implementation status in shared-services.Let me check the driver's SOS handle instance and the auth token in `sosCreate` for the Dashboard flow.I now have enough context for a comprehensive review. Let me write it up.

---

## PR Review: #14441 — WIP SafetySettings API Unification

**Overall verdict:** Good architectural direction — consolidating Safety APIs into `shared-services` with a typeclass/service-handle pattern is the right call. However, several correctness issues need resolution before merge.

---

### Critical

**1. Emergency-contact validation gap in `updateSafetySettings`** — Confidence: 88

`Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Profile.hs` — rider's `riderHandlePlatformUpdate` validates that contacts are non-empty only when the legacy `shareEmergencyContacts` alias is set. But the canonical fields `autoCallDefaultContact` and `notifySosWithEmergencyContacts` can activate the same contact-based behaviour via `updateSafetySettings` and bypass this check entirely.

```haskell
-- Current: only checks shareEmergencyContacts
when (req.shareEmergencyContacts == Just True) $ do
  contacts <- ...
  when (null contacts) $ throwError $ InvalidRequest "No emergency contacts"

-- Should also check canonical fields:
let contactSharingEnabled = req.shareEmergencyContacts == Just True
                         || req.autoCallDefaultContact == Just True
                         || req.notifySosWithEmergencyContacts == Just True
when contactSharingEnabled $ do ...
```

---

**2. `aggregatedRideShareSetting` — PUT→GET non-idempotency risk** — Confidence: 85

`Backend/lib/shared-services/src/Safety/Domain/Action/UI/SafetySettings.hs`

`updateSafetySettings` persists `req.aggregatedRideShareSetting` via `UpdateEmergencyInfo { aggregatedRideShare = ... }`. But `getSafetySettings` reads back `safetySettings.aggregatedRideShareSetting` from `findSafetySettingsWithFallback`. If `QSafetyExtra.upsert` doesn't write the `aggregated_ride_share_setting` column, the GET after PUT will silently return a stale value. For drivers who have no emergency-contact rows, this field will always be `Nothing`.

This needs a verified round-trip test or a code trace through `QSafetyExtra.upsert` to confirm the column is actually written.

---

**3. Non-ride SOS tracking URL missing `sosId`** — Confidence: 83

`Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Sos.hs`

When `sosTrackingLink` is not configured and `mbRideId` is `Nothing`, the generated URL returns only the base `shareRide` URL with no `sosId` appended. Any non-ride SOS (e.g. a post-ride trigger) produces an unresolvable tracking link. The old `buildSosTrackUrl` handled this case explicitly; the shared equivalent needs the same fallback:

```haskell
-- Missing: append sosId when no rideId
Nothing -> shareRideBaseUrl <> "/" <> sosId.getId
```

---

### Warning

**4. Driver rate-limit values are effectively disabled — security concern** — Confidence: 80

`Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Sos.hs`

```haskell
SafetySos.sosTrackingRateLimitOptions = APIRateLimitOptions {limit = 1000000, limitResetTimeInSec = 60},
SafetySos.erssStatusUpdateRateLimitOptions = APIRateLimitOptions {limit = 1000000, limitResetTimeInSec = 60}
```

1 M requests/min is functionally no rate limit. If a feature flag (`enableShareRide`, `enableIvr`) ever fails to gate correctly, these endpoints are fully open. Use a named constant or a proper disabled-sentinel (`Nothing`/`maxBound`) and document why the driver needs a different limit from rider.

---

**5. First-contact aggregation is not the aggregate** — Confidence: 79

`Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Profile.hs` (~line 686)

`riderGetPersonDefaults` (and the response builder) derives `shareTripWithEmergencyContactOption` from only the first contact in the list. If contacts have mixed options (some `ALWAYS_SHARE`, some `NEVER_SHARE`) the response reports `NEVER_SHARE` while another contact actively shares. This was a pre-existing issue in rider, but it now propagates to the shared path used by both platforms.

---

**6. Execution order comment is misleading** — Confidence: 85

`Backend/lib/shared-services/src/Safety/Domain/Action/UI/SafetySettings.hs`

The CodeRabbit review comment on this file states "upsert runs before handlePlatformUpdate". **This is incorrect based on the actual diff** — the call order is:

```haskell
svcHandle.handlePlatformUpdate personId req   -- ← runs first (validation)
QSafetyExtra.upsert personId emergencyInfo …  -- ← runs second (DB write)
```

The current ordering is correct. However, the comment in the file and the plan doc (`SosAPIUnification.md`) should be clarified to avoid future readers applying a "fix" that breaks it.

---

### Suggestion

**7. `shareEmergencyContacts` is always `Just` — API contract ambiguity** — Confidence: 75

`Backend/lib/shared-services/src/Safety/Domain/Action/UI/SafetySettings.hs:84`

```haskell
shareEmergencyContacts = Just safetySettings.notifySosWithEmergencyContacts,
```

Returning `Just` unconditionally means clients can never distinguish "the field is unset" from `Just False`. If backward compat requires this alias, document it explicitly in the type definition:

```haskell
-- Rider-compat alias for notifySosWithEmergencyContacts; always populated.
shareEmergencyContacts :: Maybe Bool
```

---

**8. `SosPersonData` is a 20-field god-object** — Confidence: 70

`Backend/lib/shared-services/src/Safety/Domain/Action/UI/Sos.hs`

`SosPersonData` mixes identity fields (`personId`, `merchantId`), config fields (`kaptureDisposition`, `trackingUrlPattern`), and display fields (`personDisplayName`). As more platforms are onboarded this record will grow unbounded. Consider splitting into `SosPersonIdentity`, `SosMerchantConfig`, and letting callbacks return only the subrecord they need.

---

### Housekeeping (must fix before merge)

- **PR description and motivation are empty.** The `## Description` and `## Motivation and Context` sections in the template are blank. A PR touching 54 files across 3 services needs at minimum a paragraph explaining what changed and why.
- **Checklist is unchecked** — formatting, linter, tests are all unchecked. The GitHub Advanced Security scan already flagged 4 `hlint` warnings (`Use notElem`, `Use maybe` × 3, `Redundant bracket`) that should be addressed.
- **WIP comment vs. actual state** — the comment block says "Steps 2–8 replace each stub body with the real implementation" but the `sosCreate`, `sosStatus`, etc. functions appear to be fully implemented, not stubs. If the implementation is real, remove the misleading "Step 1 stub" comment.
- **`SafetySettings.yaml` diff not shown** — the rename from `DriverSafetySettings.yaml` to `Safety/API/SafetySettings.yaml` hides content changes. Verify the spec matches the actual `SafetySettingsAPI` Servant type (especially: no personId URL parameter now, which is an API-breaking change for existing driver clients).

---

### What's working well

- The typeclass + service-handle design is clean: `HasSosHandle`, `HasSafetySettingsHandle`, `HasEmergencyContactHandle` all follow the same pattern and make platform behaviour explicit.
- Driver's `PlatformConfig` boolean flags (`enableFollowRide = False`, `enableShareRide = False`, etc.) are exactly the "toggle features platform-wise" goal described in the PR.
- The `BuildSosCtx`/`BuildSafetySettingsCtx` instances for 2-tuple (rider) vs 3-tuple (driver) auth tokens are a concise solution to the token-shape mismatch.
- Removing the `personId` URL parameter and deriving it entirely from the auth token is a cleaner auth model.

PR #14468, opened 2026-04-22
[feat]: SVP for CMRL

Yashwanth1906:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added SVP endpoints: QR generation (accepts location) and gate callback for ENTRY/EXIT flows; integrated into UI API surface.
  * API models for QR responses, gate callbacks, and scan types added and documented.

* **User-facing behavior**
  * Server-side QR signing with per-rider ticket serials and 8-hour validity.
  * Gate logic enforces balance checks, ENTRY/EXIT handling, fare resolution and optional wallet redemption.
  * Persistent journey tracking with statuses and fare capture.

* **Chores**
  * Database: new svp_journey table and composite index for rider/status lookups.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/spec/API/SVP.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/API/SVP.yaml
@@ -1,0 +1,51 @@
+imports:
+  Text: Data.Text
+  UTCTime: Kernel.Prelude
+  Merchant: Domain.Types.Merchant
+  SvpJourneyStatus: Domain.Types.SvpJourney
+
+module: SVP
+
+types:
+
+  GenerateQrResp:
+    qrData: Text
+    tktSlNo: Text
+    validTill: UTCTime
+    journeyStatus: Maybe SvpJourneyStatus
+
+  ScanType:
+    enum: "ENTRY,EXIT"
+    derive': "Show, Eq, Ord, Read, Generic, ToJSON, FromJSON, ToSchema"
+
+  GateCallbackReq:
+    mobileNumber: Text
+    merchantId: Id Merchant
+    stationCode: Text
+    timestamp: UTCTime
+    scanType: ScanType
+
+  GateCallbackResp:
+    allowed: Bool
+    reason: Maybe Text
+
+apis:
+
+  # Generate (or refresh) SVP QR for the authenticated rider
+  - GET:
+      endpoint: /svp/qr
+      auth: TokenAuth
+      query:
+        lat: Double
+        lon: Double
+      response:
+        type: GenerateQrResp
+
+  # AFC gate callback — called by CMRL AFC on QR scan
+  - POST:
+      endpoint: /svp/gate
+      auth: NoAuth
+      request:
+        type: GateCallbackReq
+      response:
+        type: GateCallbackResp
--- a/Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
@@ -1,0 +1,56 @@
+imports:
+  UTCTime: Kernel.Prelude
+  HighPrecMoney: Kernel.Types.Common
+  Currency: Kernel.Types.Common
+  Merchant: Domain.Types.Merchant
+  MerchantOperatingCity: Domain.Types.MerchantOperatingCity
+  Person: Domain.Types.Person
+
+SvpJourney:
+  tableName: svp_journey
+
+  types:
+    SvpJourneyStatus:
+      enum: "ENTERED,EXITED,TIMED_OUT"
+      derive': "Show, Eq, Ord, Read, Generic, ToJSON, FromJSON, ToSchema, Kernel.Prelude.ToParamSchema"
+
+  domainInstance:
+    - Custom Kernel.Beam.Lib.UtilsTH.mkBeamInstancesForEnum <SvpJourneyStatus>
+    - Custom Kernel.Utils.TH.mkFromHttpInstanceForEnum <SvpJourneyStatus>
+    - Custom Kernel.Utils.TH.mkToHttpInstanceForEnum <SvpJourneyStatus>
+
+  fields:
+    id: Id SvpJourney
+    riderId: Id Person
+    tktSlNo: Text
+    status: SvpJourneyStatus
+    entryStationCode: Maybe Text
+    entryTime: Maybe UTCTime
+    exitStationCode: Maybe Text
+    exitTime: Maybe UTCTime
+    fareCharged: Maybe HighPrecMoney
+    currency: Maybe Currency
+    merchantId: Id Merchant
+    merchantOperatingCityId: Id MerchantOperatingCity
+    createdAt: UTCTime
+    updatedAt: UTCTime
+
+  constraints:
+    id: PrimaryKey
+    riderId: "!SecondaryKey"
+
+  queries:
+    findById:
+      kvFunction: findOneWithKV
+      where: id
+    findByRiderIdAndStatus:
+      kvFunction: findOneWithKV
+      where:
+        and: [riderId, status]
+    updateStatusAndExitDetailsById:
+      kvFunction: updateWithKV
+      params: [status, exitStationCode, exitTime, fareCharged, currency, updatedAt]
+      where: id
+
+  extraOperations:
+    - EXTRA_QUERY_FILE
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
@@ -1,0 +1,406 @@
+module Domain.Action.UI.SVP
+  ( getSvpQr,
+    postSvpGate,
+  )
+where
+
+import qualified API.Types.UI.SVP as API
+import qualified BecknV2.OnDemand.Enums as Spec
+import qualified Crypto.Hash as Hash
+import qualified Crypto.PubKey.RSA as RSA
+import qualified Crypto.PubKey.RSA.PKCS15 as RSA.PKCS15
+import qualified Crypto.Store.PKCS8 as PKCS8
+import Data.Bits ((.&.), (.|.))
+import qualified Data.ByteString.Base64 as B64
+import Data.Char (isDigit, toUpper)
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as TE
+import Data.Time.Clock.POSIX (utcTimeToPOSIXSeconds)
+import Data.Time.Format (defaultTimeLocale, formatTime)
+import Data.Word (Word32)
+import Data.X509 (PrivKey (..))
+import Domain.Types.IntegratedBPPConfig (ProviderConfig (..))
+import qualified Domain.Types.IntegratedBPPConfig as DIBC
+import qualified Domain.Types.Merchant as DM
+import qualified Domain.Types.Person as DP
+import qualified Domain.Types.SvpJourney as DSvp
+import qualified ExternalBPP.ExternalAPI.Metro.CMRL.V2.GetFare as CMRLV2GetFare
+import Kernel.External.Encryption (decrypt, getDbHash)
+import Kernel.External.Wallet.Interface.Types (WalletPostingReq (..), WalletPostingType (REDEEM))
+import Kernel.Prelude
+import qualified Kernel.Storage.Hedis as Hedis
+import Kernel.Tools.Metrics.CoreMetrics (CoreMetrics)
+import Kernel.Types.Id
+import Kernel.Utils.Common
+import qualified Lib.Payment.Storage.Queries.PersonWallet as QPersonWallet
+import Numeric (showHex)
+import qualified SharedLogic.IntegratedBPPConfig as SIBC
+import qualified Storage.Queries.Person as QPersonBase
+import qualified Storage.Queries.PersonExtra as QPerson
+import qualified Storage.Queries.SvpJourney as QSvpJourney
+import Tools.Error
+import qualified Tools.Wallet as TWallet
+
+-- ── Redis keys ────────────────────────────────────────────
+
+rsaPrivateKeyRedisKey :: Text
+rsaPrivateKeyRedisKey = "SVP:RSAPrivateKeyPEM"
+
+tktSlNoRedisKey :: Id DP.Person -> Text
+tktSlNoRedisKey riderId = "SVP:TktSlNo:" <> getId riderId
+
+tktSlNoCounterKey :: Text
+tktSlNoCounterKey = "SVP:TktSlNoCounter"
+
+toHexN :: (Integral a) => Int -> a -> Text
+toHexN bytes val =
+  let h = map toUpper $ showHex (toInteger val) ""
+      padLen = bytes * 2
+   in T.pack $ replicate (max 0 (padLen - length h)) '0' ++ take padLen h
+
+encodeCoord :: Maybe Double -> Text
+encodeCoord Nothing = toHexN 3 (0 :: Int)
+encodeCoord (Just deg) = toHexN 3 (floor (abs deg * 1000) :: Int)
+
+encodeMobile :: Text -> Text
+encodeMobile m =
+  let digits = T.filter isDigit m
+   in T.justifyRight 10 '0' (T.takeEnd 10 digits)
+
+nowEpochSec :: UTCTime -> Int
+nowEpochSec = floor . utcTimeToPOSIXSeconds
+
+buildTktSlNo :: UTCTime -> Int -> Text
+buildTktSlNo now seq' =
+  let dtHex = toHexN 4 (nowEpochSec now)
+      seqWord = fromIntegral (0x40000000 .|. (seq' .&. 0x3FFFFFFF)) :: Word32
+   in dtHex <> toHexN 4 seqWord
+
+getOrCreateTktSlNo :: (CacheFlow m r, MonadFlow m) => Id DP.Person -> m Text
+getOrCreateTktSlNo riderId = do
+  mbSlNo <- Hedis.get (tktSlNoRedisKey riderId)
+  case mbSlNo of
+    Just slNo -> pure slNo
+    Nothing -> do
+      seq' <- Hedis.incr tktSlNoCounterKey
+      now <- getCurrentTime
+      let slNo = buildTktSlNo now (fromIntegral seq')
+      Hedis.setExp (tktSlNoRedisKey riderId) slNo (8 * 3600)
+      pure slNo
+
+buildSqdsrPlaintext ::
+  Text -> UTCTime -> Int -> Text -> Text -> Text -> Text -> Text
+buildSqdsrPlaintext tktSlNo now balancePaisa txnRef mobile bookingLat bookingLon =
+  let epoch = nowEpochSec now
+      sec = toHexN 1 (0x03 :: Int)
+      ver = toHexN 1 (0x04 :: Int)
+      lang = toHexN 1 (0x00 :: Int)
+      tgId = toHexN 2 (0x0001 :: Int)
+      txnTy = toHexN 1 (0x41 :: Int)
+      genDt = toHexN 4 epoch
+      reqId = toHexN 4 (0x00000001 :: Int)
+      txnRefPadded = T.justifyLeft 22 '0' (T.take 22 txnRef)
+      totalFare = toHexN 4 (0 :: Int)
+      svcFields =
+        T.intercalate
+          "|"
+          [ sec,
+            ver,
+            lang,
+            tgId,
+            txnTy,
+            tktSlNo,
+            genDt,
+            reqId,
+            txnRefPadded,
+            totalFare,
+            bookingLat,
+            bookingLon,
+            mobile
+          ]
+      updTime = toHexN 4 epoch
+      qrStatus = toHexN 2 (0x0087 :: Int) <> toHexN 1 (0x11 :: Int)
+      opDyn = toHexN 4 balancePaisa <> toHexN 15 (0 :: Int)
+      dynFields =
+        T.intercalate
+          "|"
+          [updTime, qrStatus, toHexN 3 (0 :: Int), toHexN 3 (0 :: Int), opDyn]
+      opId = toHexN 2 (0x0087 :: Int)
+      noTkts = toHexN 1 (0x01 :: Int)
+      valInfo = toHexN 1 (0x10 :: Int)
+      grpSize = toHexN 1 (0x01 :: Int)
+      srcStn = toHexN 2 (0x0000 :: Int)
+      dstStn = toHexN 2 (0x0000 :: Int)
+      actDt = toHexN 4 epoch
+      prodId = toHexN 2 (0x0005 :: Int)
+      svcId = toHexN 1 (0x01 :: Int)
+      tktFare = toHexN 4 (0 :: Int)
+      validity = toHexN 2 (0x01E0 :: Int)
+      duration = toHexN 2 (0x00B4 :: Int)
+      opTktData = toHexN 4 balancePaisa <> toHexN 4 (0x4E594D52 :: Int)
+      tktFields =
+        T.intercalate
+          "|"
+          [ grpSize,
+            srcStn,
+            dstStn,
+            actDt,
+            prodId,
+            svcId,
+            tktFare,
+            validity,
+            duration,
+            opTktData
+          ]
+      tktPart =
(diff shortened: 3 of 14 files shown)


Jayanth-Parthsarathy:
@ClaudeWeb review this

Jayanth-Parthsarathy (commented):
## PR Review: #14468 - [feat]: SVP for CMRL
**Author**: Yashwanth1906 | **Base**: main | **Files**: 4 | **+173 -14**

### CI Status: PASS

---

### CRITICAL Issues

- **[C-SEC-1] Any authenticated rider can rotate the signing keys**
  - File: `src/API/UI/SVP.hs:30-31`, `src/Domain/Action/UI/SVP.hs:73-80`
  - All three endpoints use `TokenAuth`, meaning any logged-in rider can call `POST /svp/rotate-keys` and replace the RSA keypair. This would **invalidate all previously signed QR codes** and is a denial-of-service vector. Key rotation must be restricted to admin/dashboard auth.

- **[C-SEC-2] Any authenticated rider can sign arbitrary plaintext**
  - File: `src/Domain/Action/UI/SVP.hs:49-57`
  - `POST /svp/sign-qr` lets any rider sign any string. If the QR signature is used for verification/access (e.g., CMRL gate entry), this means any user can forge valid QR codes with arbitrary content. The endpoint should either:
    - Be restricted to a server-to-server/internal auth, OR
    - Construct the plaintext server-side from trusted data (booking ID, rider ID, timestamp) rather than accepting arbitrary user input.

- **[C-SEC-3] RSA private key stored as plaintext in Redis**
  - File: `src/Domain/Action/UI/SVP.hs:25` (key: `"SVP:RSAPrivateKeyPEM"`)
  - The private signing key is stored unencrypted in Redis via `Hedis.set`. If Redis is compromised (or accessible from other services), the entire signing scheme is broken. Consider using the project's `EncryptedField` / encryption-at-rest patterns, or store the key in a secrets manager.

- **[C-SEC-4] `getPublicKey` endpoint exposes the public key to any authenticated user**
  - File: `src/API/UI/SVP.hs:28-29`
  - While less severe than the above, if the public key is only needed by CMRL's verification infrastructure, exposing it to all riders broadens the attack surface unnecessarily. Consider making this an internal/admin endpoint.

### IMPORTANT Issues

- **[I2] Race condition in key bootstrap**
  - File: `src/Domain/Action/UI/SVP.hs:83-93`
  - `loadOrBootstrapKey` checks Redis, and if empty, calls `rotateKeys` then reads again. Under concurrent requests, multiple processes could bootstrap simultaneously, each overwriting the other's key. Use a Redis lock (`SETNX` / distributed lock) or atomic bootstrap.

- **[I3] No TTL on the Redis key — key persists indefinitely**
  - File: `src/Domain/Action/UI/SVP.hs:77` — uses `Hedis.set` (no expiry)
  - If the intent is periodic rotation, consider `Hedis.setExp` with a TTL matching the rotation schedule. Currently, if `rotateKeys` is never called again, the same key lives forever.

- **[I3] `personId` is unused in business logic**
  - File: `src/API/UI/SVP.hs:42,49,55`
  - The handlers receive `personId` from `TokenAuth` but never pass it to the domain functions. There's no audit trail of who signed what or who rotated keys. At minimum, log the `personId` for each operation.

- **[I3] License header removed from `API/UI.hs`**
  - File: `src/API/UI.hs` (lines 1-14 deleted)
  - The AGPL license header was stripped from `API/UI.hs`. This appears unintentional and should be restored.

- **[I5] Commits: 1** — Good, no squash needed.

### Good to Have

- **[G2]** The module name `SVP` is opaque. A comment or module-level haddock explaining what SVP stands for (Secure Verification Protocol? Signed Verification Pass?) would help future readers.

---

### Impact Summary
| Aspect | Status |
|--------|--------|
| Touches ride booking flow | NO |
| Has DB migrations | NO |
| API changes | YES (3 new endpoints, backward compatible: YES) |
| Commits | 1 (recommend squash: NO) |
| Risk level | **HIGH** (security-sensitive cryptographic signing with insufficient access control) |

### Recommendation
**REQUEST CHANGES**

The feature's crypto implementation works mechanically, but the access control model is fundamentally broken — all three endpoints should be restricted (admin or internal-only), and the sign endpoint should not accept arbitrary user-supplied plaintext. The private key storage also needs hardening.

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+        API.GateCallbackResp
+          { allowed = False,
+            reason =
+              Just $
+                "Insufficient balance. Minimum ₹"
+                  <> show (svpMinFarePaisa `div` 100)
+                  <> " required."
+          }
+    else do
+      mbTktSlNo <- Hedis.get (tktSlNoRedisKey person.id)
```
Jayanth-Parthsarathy, line 288: If `serialNumber` is not there `throwError` since it must have been generated at `getSvpQr`. So something has gone wrong

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+      let tktSlNo' = fromMaybe "UNKNOWN" mbTktSlNo
```
Jayanth-Parthsarathy: Proceed with the api flow only if `isJust mbTktSlNo`

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+      journeyId <- generateGUID
```
Jayanth-Parthsarathy: can be renamed to `svpJourneyId`

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+              travelDatetime = travelDatetime,
+              fareTypeId = cmrlCfg.fareTypeId
+            }
+      pure $ case fares of
+        (f : _) -> case f.categories of
+          (cat : _) -> cat.offeredPrice.amount
+          _ -> HighPrecMoney 0
+        _ -> HighPrecMoney 0
+    _ -> do
+      logWarning "[SVP:Exit] No CMRLV2 config — fare defaulting to ₹0"
```
Jayanth-Parthsarathy, line 362: Should handle this post demo

Jayanth-Parthsarathy (changes requested):

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+
+buildTktSlNo :: UTCTime -> Int -> Text
+buildTktSlNo now seq' =
+  let dtHex = toHexN 4 (nowEpochSec now)
+      seqWord = fromIntegral (0x40000000 .|. (seq' .&. 0x3FFFFFFF)) :: Word32
+   in dtHex <> toHexN 4 seqWord
+
+getOrCreateTktSlNo :: (CacheFlow m r, MonadFlow m) => Id DP.Person -> m Text
+getOrCreateTktSlNo riderId = do
+  mbSlNo <- Hedis.get (tktSlNoRedisKey riderId)
```
Jayanth-Parthsarathy, line 81: might race with two concurrent requests from same userId. If possible we can fix now or post demo

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+generateAndStoreKeyPair :: (CacheFlow m r, MonadFlow m) => m RSA.PublicKey
+generateAndStoreKeyPair = do
+  (pub, priv) <- liftIO $ RSA.generate 256 0x10001
+  Hedis.set rsaPrivateKeyRedisKey (encodePrivateKeyPem priv)
+  pure pub
+
+loadOrBootstrapKey :: (CacheFlow m r, EsqDBFlow m r, MonadFlow m) => m RSA.PrivateKey
+loadOrBootstrapKey = do
+  mbPem <- Hedis.get rsaPrivateKeyRedisKey
+  pem <- case mbPem of
```
Jayanth-Parthsarathy, line 171: more chances of racing here as well since more than one user will scan at AFC gate

Jayanth-Parthsarathy (dismissed):
Flow is fine only but we must fix some race conditions on concurrent requests and some flow changes post demo

khuzema786:
@ClaudeWeb review this PR

Backend/app/rider-platform/rider-app/Main/spec/API/SVP.yaml
```diff
@@ -0,0 +1,51 @@
...
+    validTill: UTCTime
+    journeyStatus: Maybe SvpJourneyStatus
+
+  ScanType:
+    enum: "ENTRY,EXIT"
+    derive': "Show, Eq, Ord, Read, Generic, ToJSON, FromJSON, ToSchema"
+
+  GateCallbackReq:
+    mobileNumber: Text
+    merchantId: Id Merchant
```
khuzema786, line 23: change to integratedBppConfigId

Backend/app/rider-platform/rider-app/Main/spec/API/SVP.yaml
```diff
@@ -0,0 +1,51 @@
...
+      query:
+        lat: Double
+        lon: Double
+      response:
+        type: GenerateQrResp
+
+  # AFC gate callback — called by CMRL AFC on QR scan
+  - POST:
+      endpoint: /svp/gate
+      auth: NoAuth
```
khuzema786, line 47: Some auth, to be added

Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - Custom Kernel.Utils.TH.mkFromHttpInstanceForEnum <SvpJourneyStatus>
+    - Custom Kernel.Utils.TH.mkToHttpInstanceForEnum <SvpJourneyStatus>
+
+  fields:
+    id: Id SvpJourney
+    riderId: Id Person
+    tktSlNo: Text
+    status: SvpJourneyStatus
+    entryStationCode: Maybe Text
+    entryTime: Maybe UTCTime
```
khuzema786, line 28: Non Maybe

Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - Custom Kernel.Beam.Lib.UtilsTH.mkBeamInstancesForEnum <SvpJourneyStatus>
+    - Custom Kernel.Utils.TH.mkFromHttpInstanceForEnum <SvpJourneyStatus>
+    - Custom Kernel.Utils.TH.mkToHttpInstanceForEnum <SvpJourneyStatus>
+
+  fields:
+    id: Id SvpJourney
+    riderId: Id Person
+    tktSlNo: Text
+    status: SvpJourneyStatus
+    entryStationCode: Maybe Text
```
khuzema786, line 27: Non Maybe

Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
```diff
@@ -0,0 +1,56 @@
...
+    entryTime: Maybe UTCTime
+    exitStationCode: Maybe Text
+    exitTime: Maybe UTCTime
+    fareCharged: Maybe HighPrecMoney
+    currency: Maybe Currency
+    merchantId: Id Merchant
+    merchantOperatingCityId: Id MerchantOperatingCity
```
khuzema786, line 34: integratedBppConfig id too

khuzema786 (changes requested):
# SVP for CMRL — Review

The feature is well-structured and the domain model is clean. The QR-payload construction, the journey lifecycle schema, and the composite index are all correct. However there are two **critical** security/correctness issues that must be fixed before merge, plus several warnings that will cause real production problems.

---

## 🔴 Critical

### 1. `POST /svp/gate` is completely unauthenticated — wallet deductions are exploitable by anyone

`spec/API/SVP.yaml` declares `auth: NoAuth` for `/svp/gate`. The request body contains `mobileNumber` and `merchantId`, both attacker-controlled. There is **no HMAC signature, API-key, IP allowlist, or any other mechanism** to verify the request actually comes from CMRL's AFC system.

**Impact:** Any internet user can POST `{"mobileNumber": "9876543210", "merchantId": "...", "scanType": "EXIT", ...}` and trigger a full `handleExit` — including `TWallet.walletPosting` (real money deduction) — for any rider whose mobile number they know.

**Fix:** Require a shared HMAC or API key from CMRL in a request header, validated before any business logic runs:

```yaml
# spec/API/SVP.yaml
- POST:
    endpoint: /svp/gate
    auth: ApiAuth CMRL_AFC_API_KEY
```

---

### 2. Silent `₹0` fare allows free rides on config miss or empty fare-API response

`Domain/Action/UI/SVP.hs` — two paths in `handleExit` silently default the fare to zero and then open the gate:

```haskell
-- Path 1: providerConfig is not CMRLV2
_ -> do
  logWarning "[SVP:Exit] No CMRLV2 config — fare defaulting to ₹0"
  pure (HighPrecMoney 0)             -- ← gate opens for free

-- Path 2: fare API returns empty list or empty categories
pure $ case fares of
  (f : _) -> case f.categories of
    (cat : _) -> cat.offeredPrice.amount
    _ -> HighPrecMoney 0             -- ← gate opens for free
  _ -> HighPrecMoney 0               -- ← gate opens for free
```

A misconfigured deployment, a momentary CMRL fare-API outage, or any edge case with an empty response lets every passenger through for free. Both paths must **fail closed**:

```haskell
-- Replace ₹0 defaults with:
_ -> throwError $ InternalError "[SVP:Exit] Fare lookup failed — cannot open gate"
```

---

## 🟠 Warning

### 3. Race condition on RSA key bootstrap

`loadOrBootstrapKey` has a TOCTOU race: two concurrent requests that both see `Nothing` in Redis will each call `generateAndStoreKeyPair`, generating two different keypairs. The last write wins in Redis, but the other request is now signing with a private key whose public key is no longer stored — those QR codes will fail AFC gate verification.

**Fix:** Use a Redis `SET NX` (set-if-not-exists) so only one worker writes the initial key, then both reload from Redis:

```haskell
Nothing -> do
  (_, priv) <- liftIO $ RSA.generate 256 0x10001
  let pem = encodePrivateKeyPem priv
  _ <- Hedis.setNx rsaPrivateKeyRedisKey pem  -- no-op if another worker won
  Hedis.get rsaPrivateKeyRedisKey             -- always reload the winner
    >>= fromMaybeM (InternalError "SVP: key bootstrap failed")
```

---

### 4. No duplicate ENTRY guard — orphan journeys accumulate

`handleEntry` creates a new `SvpJourney` row unconditionally. A double-scan at the AFC gate or a retry produces two `ENTERED` rows for the same rider. `findByRiderIdAndStatus` (using `findOneWithKV`) returns only one, so the other is never marked `EXITED` and corrupts all future lookups.

**Fix:** Check for an existing active journey first:

```haskell
mbExisting <- QSvpJourney.findByRiderIdAndStatus person.id DSvp.ENTERED
when (isJust mbExisting) $
  pure API.GateCallbackResp { allowed = True, reason = Just "Already inside" }
```

---

### 5. Wallet deduction is not atomic with the DB journey update

`handleExit` calls `TWallet.walletPosting` first, then `updateStatusAndExitDetailsById`. If the process crashes between the two steps, money has been deducted but the journey stays `ENTERED` — the next EXIT scan will charge the rider again.

**Recommended pattern:**
1. Mark journey as `PAYMENT_PENDING` before calling the wallet.
2. On wallet success → mark `EXITED`.
3. On wallet failure → revert to `ENTERED`, return `allowed = False`.

This makes the flow safe to retry.

---

### 6. `tktSlNo` Redis TTL is never refreshed on subsequent QR calls

`getOrCreateTktSlNo` sets an 8-hour TTL only when the key is *first created*. `getSvpQr` always returns `validTill = now + 8h`, but the actual Redis entry expires relative to when it was *first set*. A rider who generated their QR 7h 50m ago will have their `tktSlNo` evicted ~10 minutes later, causing `handleEntry` to reject them with "QR not generated" while the app shows a valid QR.

**Fix — reset TTL on every QR fetch:**

```haskell
Just slNo -> do
  Hedis.expire (tktSlNoRedisKey riderId) (8 * 3600)
  pure slNo
```

---

### 7. Currency hardcoded to `INR`

```haskell
QSvpJourney.updateStatusAndExitDetailsById
  DSvp.EXITED (Just stationCode) (Just exitTime)
  (Just fareAmount)
  (Just INR)        -- ← hardcoded
  journey.id
```

The schema has a `Currency` field. Derive the currency from the CMRL config or the wallet response instead of hardcoding.

---

### 8. Minimum fare is a magic constant

```haskell
svpMinFarePaisa :: Int
svpMinFarePaisa = 5000   -- ₹50
```

Changing this requires a code deployment. Move it to `IntegratedBPPConfig` / the CMRL provider config so it can be updated via configuration.

---

### 9. Unintentional AGPL copyright header removal

`src/API/UI.hs` diff shows 14 deletions that are entirely the AGPL copyright block. This is almost certainly an unintended editor side-effect. Please restore the header.

---

## 🔵 Potential Issues

### 10. Mobile number logged/returned as plaintext PII

```haskell
-- Domain/Action/UI/SVP.hs
fromMaybeM (PersonNotFound $ "mobile=" <> mobileNum)
```

Mobile numbers are PII. This propagates into error logs and HTTP error responses. Use a hashed or redacted representation in error messages.

### 11. RSA private key durability

`Hedis.set` with no TTL stores the PEM in Redis. A Redis flush, eviction (`maxmemory-policy` other than `noeviction`), or cluster failover permanently destroys the signing key. All in-flight QR codes (valid up to 8h) become unverifiable. Consider persisting to a secrets store (Vault, AWS Secrets Manager) with Redis as a cache layer.

### 12. `encodeCoord` discards sign of coordinates

```haskell
encodeCoord (Just deg) = toHexN 3 (floor (abs deg * 1000) :: Int)
```

`abs` makes lat `12.97°N` and `12.97°S` encode identically. Verify against the CMRL QR spec — if signed coordinates are required this will silently produce wrong location data.

khuzema786:
I'll fetch the PR diff, description, and comments in parallel to build full context before reviewing.Let me try the GitHub API directly instead.Now let me fetch the diff and key source files in parallel.Now let me read the key source files for deep inspection.I have a comprehensive picture of the PR. Let me now compose the full review.Let me post without inline comments since line resolution isn't working for this branch.Review posted to the PR. Here's the summary:

---

## Review Summary — PR #14468: SVP for CMRL

**Verdict: Request Changes** — 2 critical issues, 7 warnings.

### 🔴 Critical

1. **`POST /svp/gate` is unauthenticated** (`spec/API/SVP.yaml` — `auth: NoAuth`). Anyone who knows a rider's mobile number can POST to this endpoint and trigger real wallet deductions via `TWallet.walletPosting`. Must add HMAC/API-key validation from CMRL before any business logic.

2. **Silent ₹0 fare defaults** (`Domain/Action/UI/SVP.hs`). Two code paths in `handleExit` default to `HighPrecMoney 0` (no CMRLV2 config, or empty fare API response) and then open the gate for free. Must fail closed with `throwError` instead.

### 🟠 Warnings

3. **Race condition on key bootstrap** — two concurrent requests when Redis key is absent both generate fresh keypairs; the loser signs with a discarded private key. Fix with Redis `SET NX`.
4. **No duplicate ENTRY guard** — double AFC scans create multiple `ENTERED` rows; only one ever gets `EXITED`.
5. **Wallet deduction not atomic with DB update** — crash between the two steps causes double-charge on retry.
6. **`tktSlNo` TTL not refreshed on re-fetch** — `validTill = now + 8h` in the response can lie; the Redis entry may expire much sooner.
7. **Currency hardcoded to `INR`** — derive from config.
8. **`svpMinFarePaisa = 5000` is a magic constant** — move to `IntegratedBPPConfig`.
9. **AGPL copyright header accidentally deleted** from `src/API/UI.hs`.

### 🔵 Potential
- Mobile number logged as plaintext PII in error messages.
- RSA key stored only in Redis with no durable backup.
- `encodeCoord` discards coordinate sign via `abs`.

Backend/app/rider-platform/rider-app/Main/spec/Storage/SvpJourney.yaml
```diff
@@ -0,0 +1,56 @@
...
+      kvFunction: findOneWithKV
+      where:
+        and: [riderId, status]
+    updateStatusAndExitDetailsById:
+      kvFunction: updateWithKV
+      params: [status, exitStationCode, exitTime, fareCharged, currency, updatedAt]
+      where: id
+
+  extraOperations:
+    - EXTRA_QUERY_FILE
```
khuzema786, line 56: remove

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+rsaPrivateKeyRedisKey :: Text
+rsaPrivateKeyRedisKey = "SVP:RSAPrivateKeyPEM"
+
+tktSlNoRedisKey :: Id DP.Person -> Text
+tktSlNoRedisKey riderId = "SVP:TktSlNo:" <> getId riderId
+
+tktSlNoCounterKey :: Text
+tktSlNoCounterKey = "SVP:TktSlNoCounter"
+
+toHexN :: (Integral a) => Int -> a -> Text
```
khuzema786, line 55: Move to External BPP later

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+        "(" <> opId <> "|" <> noTkts <> "|" <> valInfo
+          <> "|["
+          <> tktFields
+          <> "])"
+   in "{" <> svcFields <> "}|{" <> dynFields <> "}|{" <> tktPart <> "}"
+
+generateAndStoreKeyPair :: (CacheFlow m r, MonadFlow m) => m RSA.PublicKey
+generateAndStoreKeyPair = do
+  (pub, priv) <- liftIO $ RSA.generate 256 0x10001
+  Hedis.set rsaPrivateKeyRedisKey (encodePrivateKeyPem priv)
```
khuzema786, line 165: add expiry when properly implmenting

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+handleEntry ::
+  (CacheFlow m r, EsqDBFlow m r, MonadFlow m, EncFlow m r) =>
+  Text ->
+  Id DM.Merchant ->
+  Text ->
+  UTCTime ->
+  m API.GateCallbackResp
+handleEntry mobileNum mId stationCode entryTime = do
+  person <- findRiderByMobile mobileNum mId
+  mbWallet <- QPersonWallet.findByPersonId (getId person.id)
```
khuzema786, line 272: Finance Account with Liability Account type and counterPartyId as personId

khuzema786, line 272: ask hari, creating coplex logic, he will only solve then...

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/SVP.hs
```diff
@@ -0,0 +1,406 @@
...
+              toStationId = stationCode,
+              ticketTypeId = cmrlCfg.ticketTypeId,
+              merchantId = cmrlCfg.merchantId,
+              travelDatetime = travelDatetime,
+              fareTypeId = cmrlCfg.fareTypeId
+            }
+      pure $ case fares of
+        (f : _) -> case f.categories of
+          (cat : _) -> cat.offeredPrice.amount
+          _ -> HighPrecMoney 0
```
khuzema786, line 359: bad, throw error better

khuzema786 (approved):

PR #14507, opened 2026-04-24
added logic to update the wallet ledger

Hari-Krishna-Dupaguntla:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added loyalty wallet (multi-currency balances, points, cashback) with wallet payment tracking and wallet top‑up + automatic loyalty reward processing; order status now reflects wallet-specific handling.

* **Chores**
  * Added database migrations for wallet and wallet_payments tables and updated payment modules/schema to support loyalty wallets.

* **Tests / Docs**
  * Added integration test collection and README for the LoyaltyWallet flow and a Juspay loyalty mock endpoint.


Hari-Krishna-Dupaguntla:
@ClaudeWeb Please review this PR ?

Hari-Krishna-Dupaguntla (commented):
## PR Review (automated)

### CI Status: PENDING
- `db_check`: pass
- `Master Local Build`: pending
- `Scan code with HLint`: pending
- `CodeRabbit`: in progress

---

### CRITICAL Issues
None found. The two new tables (`wallet`, `wallet_payments`) are brand-new, so the `NOT NULL` columns without defaults are safe (no rows to violate the constraint). No PII fields in the new YAML specs (`personId`, `merchantId` are foreign-key Text, not personal identifiers). No Redis pattern-matching usage. No API spec changes.

---

### IMPORTANT Issues

- **[I3]** `Currency` is hardcoded to `INR` in `postWalletRecharge`.
  - File: `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Payment.hs:896`
  - Suggestion: Derive currency from merchant config or pull from the order/request, since the rest of the codebase respects multi-currency. India-only is the current reality, but a hardcoded `INR` will silently mismatch the day a non-INR merchant is onboarded.

- **[I3]** `fromMaybe ""` for `merchantOperatingCityId` is fragile.
  - File: `Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs` (multiple sites)
  - Details: Several code paths pass `\"\"` when the city id is missing, then write that empty string to the ledger / finance account. An empty `merchantOperatingCityId` later joining or filtering will produce silent misses. Either keep the field optional all the way down, or fail explicitly when it's required.

- **[I1]** `findByOrderAndProgramAndKind` filters on three columns; only `orderId` is indexed.
  - File: `Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml`
  - Suggestion: Acceptable today (low row volume), but the YAML's own TODO calls for a unique constraint on `(orderId, programId, kind)`. Add a Postgres unique index in the same migration to (a) prevent duplicate rows from webhook retries — currently `upsertWalletPayment` relies on read-then-write which is racy under concurrent reconciliations — and (b) accelerate the filter.

- **[I3]** `upsertWalletPayment` has a read-check-write race.
  - File: `Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs` (`upsertWalletPayment`)
  - Details: Two concurrent webhook deliveries for the same `(orderId, programId, kind)` will both miss the existing-row check and both insert. Couples to the missing unique index above; if you add the index, also handle the duplicate-key error here.

- **[I2]** Verbose log spam from `[loyaltySvc]` instrumentation.
  - File: `Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs` (`processLoyaltyInfoFromOrderStatus`)
  - Details: ~15 `logInfo` calls per order-status reconciliation. Useful for debugging, but in steady state every charged order will pay this cost on every status check (and via the scheduler). Suggest downgrading to `logDebug` for the per-iteration breadcrumbs and keeping `logInfo` only on outcomes (WrittenNew, AlreadyWritten, errors).

- **[I3]** `try @_ @SomeException fetchFullInfo` swallows all exceptions silently.
  - File: `Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs` (`processLoyaltyInfoFromOrderStatus`)
  - Details: Catching `SomeException` is the broadest catch in Haskell — it will swallow async exceptions, thread kills, etc. The error is logged, but consider catching `HttpException` or a narrower payment-API exception type, and let truly fatal exceptions propagate.

- **[I4]** `walletOrderStatusHandler` ignores its `_merchantId` argument.
  - File: `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs` (`walletOrderStatusHandler`)
  - Suggestion: If the merchant is genuinely unused, drop the parameter rather than silencing it with `_`. Otherwise, the handler should at least scope its `findByOrderId` lookup by merchant for safety (multi-tenant isolation).

- **[C4 / RIDE_FLOW]** `orderStatusHandlerWithRefunds` now triggers loyalty processing for any `PaymentStatus` carrying `orderLoyaltyInfo`.
  - File: `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs` (`orderStatusHandlerWithRefunds`)
  - Impact: MEDIUM. `orderStatusHandlerWithRefunds` is invoked by both rider-side ride/booking status checks and the wallet/recharge flow. It is wrapped in `withTryCatch`, so loyalty failures will not crash the order-status response — good. However:
    - QA should test: a CHARGED ride-booking order with `orderLoyaltyInfo = Nothing` (current default) — verify zero behavioural change.
    - QA should test: a CHARGED wallet-recharge order — verify wallet and ledger rows are written exactly once across (a) initial webhook, (b) scheduler-driven re-poll, (c) refund-status job.
    - QA should test: a wallet-recharge order that goes CHARGED → AUTO_REFUNDED — confirm `recordLoyaltyEarnReversal` fires once and aggregates revert correctly.

---

### Good to Have

- **[G3]** Unrelated formatting drift in `FleetOperatorDailyStats.hs` (`enableKVPG` parenthesised → bare TH name list). Likely from running `, run-generator --all`. Not a problem, but unrelated to the PR's intent — consider isolating in a separate \"regenerate\" commit.
- **[G3]** Pre-existing dead code block (`redeemWallet`) in `Domain/Action/UI/Payment.hs`. Not introduced by this PR but worth deleting while you're in the file.
- **[G1]** New `cashbackEarned` / `topupEarned` aggregates on `wallet` table — verify they are surfaced somewhere downstream (dashboard, API response). If not, you're paying the storage/index cost without a consumer yet.

---

### Impact Summary
| Aspect | Status |
|--------|--------|
| Touches ride booking flow | INDIRECT (shared `orderStatusHandlerWithRefunds` wrapped in try-catch) |
| Has DB migrations | YES (2 new tables × 2 services; backward compatible — new tables) |
| API changes | NO (no `spec/API` files changed; `PaymentStatusResp` adds a `Maybe` field — backward compatible) |
| New enum value | YES (`CounterpartyType` adds `LOYALTY_WALLET` — additive, safe) |
| Generated files match specs | YES |
| Commits | 1 (no squash needed) |
| Risk level | MEDIUM (new code path on shared payment status handler; mitigated by try-catch + Nothing-guarded entry) |

### Recommendation
**NEEDS DISCUSSION** — no blockers, but the unique-index/race condition (I1+I3) and hardcoded `INR` are worth addressing before merge. Wait for `Master Local Build` to finish before approving.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -95,12 +102,58 @@ getLoyaltyInfo ::
 getLoyaltyInfo customerId merchantId merchantOperatingCityId =
   LoyaltyWallet.loyaltyInfo customerId merchantId merchantOperatingCityId -- need to write the storing logic
 
+loadLoyaltyProgramMap ::
+  (CacheFlow m r, EsqDBFlow m r, MonadFlow m) =>
+  Id Merchant.Merchant ->
+  Id DMOC.MerchantOperatingCity ->
+  m (Maybe (Map.Map Text Lib.Finance.Domain.Types.Account.CounterpartyType))
+loadLoyaltyProgramMap merchantId merchantOperatingCityId = do
```
khuzema786: Move to Tools.Payment

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -95,12 +102,58 @@ getLoyaltyInfo ::
...
   m (DPayment.PaymentFulfillmentStatus, Maybe Text, Maybe Text)
 
+walletOrderStatusHandler ::
+  (CacheFlow m r, EsqDBFlow m r, MonadFlow m) =>
+  Id DOrder.PaymentOrder ->
+  Id Merchant.Merchant ->
+  DPayment.PaymentStatusResp ->
+  m (DPayment.PaymentFulfillmentStatus, Maybe Text, Maybe Text)
+walletOrderStatusHandler orderId _merchantId paymentStatusResp = do
+  status <- DPayment.getTransactionStatus paymentStatusResp
```
khuzema786, line 111: No Domain table populated here, we should update this status in wallet program table

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -241,6 +294,33 @@ orderStatusHandlerWithRefunds fulfillmentHandler paymentService paymentOrder upd
...
+          let personId = cast @DPayment.Person @Person.Person updatedPaymentOrder.personId
+              merchantId = cast @DPayment.Merchant @Merchant.Merchant updatedPaymentOrder.merchantId
+              merchantOperatingCityId = cast @DPayment.MerchantOperatingCity @DMOC.MerchantOperatingCity commonMerchantOperatingCityId
+              burnRefId = fromMaybe updatedPaymentOrder.id.getId finalPaymentStatusResponse.domainEntityId
+              fetchFullInfo = getLoyaltyInfo personId.getId merchantId merchantOperatingCityId
+          mbProgramMap <- loadLoyaltyProgramMap merchantId merchantOperatingCityId
+          let resolveProgram pid = pure $ (Map.lookup pid) =<< mbProgramMap
+          res <-
+            withTryCatch "processLoyaltyInfoFromOrderStatus" $
+              LoyaltyWalletSvc.processLoyaltyInfoFromOrderStatus
```
khuzema786, line 281: this and all move to walletOrderStatusHandler

Backend/dev/integration-tests/collections/LoyaltyWalletFlow/README.md
khuzema786, line 1: Not needed, delete this

Backend/dev/mock-servers/services/juspay.py
```diff
@@ -247,3 +251,85 @@ def _order_data(handler, order_id):
         base = deep_merge(base, extra)
 
     handler._json(base)
+
+
+_DEFAULT_LOYALTY_INFO = {
```
khuzema786: call mock/update in the collection to send this as response before calling `status` API. So when status API is called with the identifier it gets this response.

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+        let kind = if isTopup then DWP.TOPUP else DWP.CASHBACK
+        wp <- upsertWalletPayment wallet personId order earn kind walletStatus
+        logInfo $ "[loyaltySvc] earn upserted wp id=" <> wp.id.getId <> " kind=" <> show wp.kind <> " status=" <> show wp.status
+        when isOrderCharged $ do
+          res <- recordLoyaltyEarn wallet wp
+          case res of
+            Left err -> logError $ "recordLoyaltyEarn failed: " <> show err
+            Right (WrittenNew _) -> do
+              logInfo $ "[loyaltySvc] earn ledger WrittenNew; bumping aggregates walletId=" <> wallet.id.getId
+              bumpWalletAggregatesOnEarn wallet kind earn.points
```
khuzema786: Different reference type

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+  Maybe Text -> -- merchantOperatingCityId
+  m (DWallet.Wallet, FAccount.Account)
+getOrCreateWalletForPerson personId programType programId currency merchantId mbMerchantOperatingCityId = do
+  mbWallet <- QWallet.findByPersonAndProgram personId programType
+  case mbWallet of
+    Just w -> do
+      mbAcc <- QFAccount.findById w.accountId
+      case mbAcc of
+        Just acc -> pure (w, acc)
+        Nothing -> createAccountAndWallet -- account somehow missing; recreate
```
khuzema786: only createAccount

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+  case mbExisting of
+    Just wp -> do
+      if wp.status == walletStatus
+        then pure wp
+        else do
+          QWP.updateStatus walletStatus wp.id
+          pure wp {DWP.status = walletStatus, DWP.updatedAt = now}
+    Nothing -> do
+      wpId <- Id <$> generateGUID
+      let firstCampaignId = fmap (.campaignId) (listToMaybe earn.campaigns)
```
khuzema786: Can we get multiple campaigns ?

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+              || mismatched mbJuspayTopup wallet.topupEarned
+              || mismatched mbJuspayCashback wallet.cashbackEarned
+          syncAvailable = fromMaybe wallet.availableBalance mbJuspayAvailable
+          syncCurrent = fromMaybe wallet.currentAvailablePoints mbJuspayAvailable
+          syncLifeEarned = fromMaybe wallet.lifetimeEarned mbJuspayLifeEarned
+          syncLifeBurned = fromMaybe wallet.lifetimeBurned mbJuspayLifeRedeemed
+          syncTopup = fromMaybe wallet.topupEarned mbJuspayTopup
+          syncCashback = fromMaybe wallet.cashbackEarned mbJuspayCashback
+      when drifted $ do
+        logInfo $
```
khuzema786: logError

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+            <> " lifeEarned="
+            <> show syncLifeEarned
+            <> " lifeBurned="
+            <> show syncLifeBurned
+            <> " topup="
+            <> show syncTopup
+            <> " cashback="
+            <> show syncCashback
+        QWallet.syncFromLoyaltyInfo syncCurrent syncAvailable syncLifeEarned syncLifeBurned syncTopup syncCashback wallet.id
+      whenJust mbJuspayAvailable $ \v -> QFAccount.updateBalance v wallet.accountId
```
khuzema786: dont update balance, use finance ctx to interact with ledger, finacne module.

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,564 @@
...
+  DWallet.Wallet ->
+  Text -> -- refId
+  HighPrecMoney -> -- points to burn
+  Currency ->
+  Text -> -- merchantId
+  Text -> -- merchantOperatingCityId
+  m (Either FinanceError LedgerWriteOutcome)
+recordLoyaltyBurn wallet refId points currency merchantId cityId = do
+  existing <- FLedger.getEntriesByReference (loyaltyReferenceTypeText LOYALTY_BURN) refId
+  case nonReversalEntries existing of
```
khuzema786: call ledger `transfer`. using finance ctx

khuzema786 (changes requested):

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
Hari-Krishna-Dupaguntla: done

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
Hari-Krishna-Dupaguntla: created fallbackstatushandler

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
Hari-Krishna-Dupaguntla: keeping it in the sharedlogic

Backend/dev/integration-tests/collections/LoyaltyWalletFlow/README.md
Hari-Krishna-Dupaguntla: deleted

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: done

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: throwing error now

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: yaa updated code to handle that thing

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: done

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: done

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: now account only updates in case of mismatch

Hari-Krishna-Dupaguntla (commented):
## Self-review: blockers before this can merge

Functional design is sensible (separating wallet aggregate state from the ledger entries, idempotent earn/burn via `getEntriesByReference`, drift detection in reconcile), but the following need addressing before merge.

### CRITICAL

- **Dev seed migration is unguarded**: `Backend/dev/feature-migrations/0008-loyalty-wallet-program-map.sql` force-overwrites `url` to `http://localhost:8080/juspay/` and `walletRewardApiVersion` on the `JuspayWallet_Juspay` `merchant_service_config` row using `||`. There is no `WHERE` guard tying it to a non-prod env or to the test merchant. If this ever runs against a non-local DB, every customer's loyalty / order-status calls would silently start hitting `localhost`. Add a guard (e.g. check for the test merchant's known short_id, or wrap in `IF current_database() = 'atlas_dev'`).

- **Behavior change in fallback fulfillment mapping (RIDE_FLOW)**. Three call sites changed from `_ -> pure (DPayment.FulfillmentPending, Nothing, Nothing)` to `_ -> SPayment.fallbackOrderStatusHandler paymentStatusResp`:
  - `Domain/Action/UI/Payment.hs:707` (`mkFulfillmentHandler`)
  - `SharedLogic/Scheduler/Jobs/CheckRefundStatus.hs:164`
  - `SharedLogic/Scheduler/Jobs/PaymentOrderStatusCheck.hs:153`
  
  Previously, any `paymentServiceType` not explicitly handled stayed `FulfillmentPending`. Now `CHARGED` → `FulfillmentSucceeded`, `AUTO_REFUNDED` → `FulfillmentRefunded`, several failure statuses → `FulfillmentFailed`, possibly transitioning orders that downstream code (and scheduler retry logic) assumed would stay pending. Affects `MetroBooking` / `FRFSBooking` / `OnlineRideBooking` / `LoyaltyWallet` etc. **QA must test** end-to-end flows for every `paymentServiceType` that hits this fallback.

### IMPORTANT

- **Missing indexes** on new tables. `dev/migrations-read-only/{rider-app,dynamic-offer-driver-app}/wallet*.sql` create the tables but no `CREATE INDEX`. On KV cache miss / non-KV reads, Postgres needs:
  - `wallet(person_id)`, `wallet(account_id)`, `wallet(person_id, program_type)`
  - `wallet_payments(order_id)`, `wallet_payments(wallet_id)`, `wallet_payments(order_id, program_id, kind)`

- **Race in `getOrCreateWalletForPerson`** (`Lib/Payment/Wallet/Service.hs`): between `findByPersonAndProgram` and `QFAccount.create + QWallet.create`, two concurrent requests can both create a `wallet` (and finance `account`) for the same `(personId, programType)`. No DB-level unique constraint on `(person_id, program_type)`. Two parallel webhooks for the same person are realistic.

- **Acknowledged TODO** in `LoyaltyWallet.yaml`: unique constraint on `(orderId, programId, kind)` for `wallet_payments`. The dedup logic in `upsertWalletPayment` (find-then-create) has the same race without a DB constraint to backstop it.

- **PR description is empty**. For a 1728-line PR touching payment + ledger, please fill in: motivation, what behavior changed, rollback plan, and which integration test scenarios were exercised.

- **"added unit tests" commit (df78528f) does not add any Haskell `*Spec.hs` files** — the only tests are the new Postman integration collection. Either rename the commit, or add the missing unit tests for `mapOrderStatusToWalletStatus`, `bumpWalletAggregatesOnEarn/Burn`, `reconcileWalletFromLoyaltyInfo`.

- **`fromMaybe ""` for missing `merchantOperatingCityId`** in `Lib/Payment/Wallet/Service.hs` (Account creation, `mkLoyaltyFinanceCtx`). Silently writing empty string into a non-nullable column is a code smell — joins/group-bys by city will lump these together. Prefer threading `Maybe` through or failing loudly.

- **Legacy path removed**: `Lib/Payment/Domain/Action.hs` removes `handleWalletOrder` / `handleWalletOrderStatus` (PersonWallet + WalletRewardPosting). Please confirm no remaining consumer (analytics, dashboards, ClickHouse pipeline) reads `person_wallet` / `wallet_reward_posting` for these flows.

- **CI coverage**: only CodeRabbit + HLint ran. Given `-Werror`, this does NOT mean the project compiles. Please post the result of `cabal build all` or trigger a real build CI before approval.

### Good to have

- `wallet.programType` is duplicated on `Account.counterpartyType` with no enforcement of consistency — document the invariant.
- Unrelated TH-style cleanup in `src-read-only/Storage/Beam/FleetOperatorDailyStats.hs` is unrelated to the wallet feature; consider splitting.
- `[loyaltySvc]` `logInfo` lines are very chatty — convert to `logDebug` once stable.

### Blockers before merge
1. Guard the `0008-loyalty-wallet-program-map.sql` URL override so it cannot run outside dev.
2. Verify and document impact of the `fallbackOrderStatusHandler` change for every `paymentServiceType`.
3. Add explicit indexes on the new tables.
4. Fill in the PR description and confirm `cabal build all` is green.
5. Either add the claimed Haskell unit tests or rename the commit.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -101,6 +104,27 @@ type FulfillmentStatusHandler m =
...
 
+fallbackOrderStatusHandler ::
+  (MonadFlow m) =>
+  DPayment.PaymentStatusResp ->
+  m (DPayment.PaymentFulfillmentStatus, Maybe Text, Maybe Text)
+fallbackOrderStatusHandler paymentStatusResp = do
+  status <- DPayment.getTransactionStatus paymentStatusResp
+  let fulfillment = case status of
+        Payment.CHARGED -> DPayment.FulfillmentSucceeded
+        Payment.AUTHENTICATION_FAILED -> DPayment.FulfillmentFailed
```
khuzema786: make DPayment.FulfillmentPending

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -101,6 +104,27 @@ type FulfillmentStatusHandler m =
...
+        Payment.AUTHORIZATION_FAILED -> DPayment.FulfillmentFailed
```
khuzema786: make DPayment.FulfillmentPending

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -241,6 +265,33 @@ orderStatusHandlerWithRefunds fulfillmentHandler paymentService paymentOrder upd
     Right (_, _, domainTransactionId) -> do
       QPaymentOrder.updatePaymentFulfillmentStatus paymentOrder.id finalPaymentStatusResponse.paymentFulfillmentStatus finalPaymentStatusResponse.domainEntityId domainTransactionId
     _ -> pure ()
+
+  case (finalPaymentStatusResponse, updatedPaymentOrder.merchantOperatingCityId) of
+    (DPayment.PaymentStatus {orderLoyaltyInfo = Just loyalty}, Just commonMerchantOperatingCityId) -> do
+      whenJust paymentOrder.paymentFulfillmentStatus $ \oldPaymentFulfillmentStatus -> do
+        when (finalPaymentStatusResponse.paymentFulfillmentStatus /= Just oldPaymentFulfillmentStatus) $ do
```
khuzema786, line 271: `when (finalPaymentStatusResponse.paymentFulfillmentStatus == DPayment.FulfillmentSucceeded && finalPaymentStatusResponse.paymentFulfillmentStatus /= Just oldPaymentFulfillmentStatus)`

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,570 @@
...
+        DWP.TOPUP -> LOYALTY_EARN_TOPUP
+        DWP.CASHBACK -> LOYALTY_EARN_CASHBACK
+  existing <- FLedger.getEntriesByReference refType wp.id.getId
+  case nonReversalEntries existing of
+    (e : _) -> pure (Right (AlreadyWritten e.id))
+    [] ->
+      runLoyaltyTransfer
+        wallet
+        wp.id.getId
+        (Finance.transfer Finance.PlatformAsset Finance.OwnerLiability wp.points refType)
```
khuzema786: If customer paid and top up then these should be the entry:
```
BuyerAsset -> OwnerLiability
```

If buyer app topped up as incentives coins or rewards without customer paying then these should be the entry:
```
BuyerExpense -> OwnerLiability
```

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,570 @@
...
+  Text -> -- refId
+  HighPrecMoney -> -- points to burn
+  m (Either FinanceError LedgerWriteOutcome)
+recordLoyaltyBurn wallet refId points = do
+  let refType = loyaltyReferenceTypeText LOYALTY_BURN
+  existing <- FLedger.getEntriesByReference refType refId
+  case nonReversalEntries existing of
+    (e : _) -> pure (Right (AlreadyWritten e.id))
+    [] ->
+      runLoyaltyTransfer
```
khuzema786, line 210: If customer reedemed:
`OwnerLiability -> BuyerAsset`

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,570 @@
...
+        wallet
+        refId
+        (Finance.transfer Finance.OwnerLiability Finance.PlatformAsset points refType)
+
+recordLoyaltyEarnReversal ::
+  (FBeamFlow.BeamFlow m r) =>
+  DWallet.Wallet ->
+  DWP.WalletPayments ->
+  m (Either FinanceError LedgerWriteOutcome)
+recordLoyaltyEarnReversal _wallet wp = do
```
khuzema786: when does this happen ?

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,570 @@
...
+        Nothing -> do
+          res <- FLedger.createReversal original.id "Juspay loyalty earn reversed"
+          pure $ fmap (WrittenNew . (.id)) res
+
+recordLoyaltyBurnReversal ::
+  (FBeamFlow.BeamFlow m r) =>
+  DWallet.Wallet ->
+  Text -> -- refId of the original burn
+  m (Either FinanceError LedgerWriteOutcome)
+recordLoyaltyBurnReversal _wallet refId = do
```
khuzema786: when does this happen ?

Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml
```diff
@@ -0,0 +1,147 @@
...
+    walletId: Id Wallet
+    personId: Text
+    orderId: Id PaymentOrder
+    programId: Text
+    kind: WalletPaymentKind
+    points: HighPrecMoney
+    benefitValue: Maybe HighPrecMoney
+    currency: Currency
+    status: WalletPaymentStatus
+    campaignId: Maybe Text
```
khuzema786: Multiple Campaigns ?

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
Hari-Krishna-Dupaguntla: no bro, in case of refund that money will get reversed

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: if refund happen then it happens

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: same. If I used 10 rupees in wallet to get the ticket and it will get refunded here

Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml
Hari-Krishna-Dupaguntla: if multiple campaign are there means multiple records will be created

Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml
```diff
@@ -0,0 +1,187 @@
...
+    id: PrimaryKey
+    walletId: "!SecondaryKey"
+    orderId: "!SecondaryKey"
+
+  queries:
+    findById:
+      kvFunction: findOneWithKV
+      where: id
+    findByOrderId:
+      kvFunction: findAllWithKV
```
khuzema786: findOneWithKV

Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml
```diff
@@ -0,0 +1,187 @@
...
+  constraints:
+    id: PrimaryKey
+    walletId: "!SecondaryKey"
+    orderId: "!SecondaryKey"
+
+  queries:
+    findById:
+      kvFunction: findOneWithKV
+      where: id
+    findByOrderId:
```
khuzema786, line 98: remove

Backend/lib/payment/spec/Storage/LoyaltyWallet.yaml
```diff
@@ -0,0 +1,187 @@
...
+      kvFunction: findAllWithKV
+      where: orderId
+    findByOrderAndProgramAndKind:
+      kvFunction: findOneWithKV
+      where:
+        and: [orderId, programId, kind]
```
khuzema786: have unique constraint on `orderId`

Backend/dev/migrations-read-only/dynamic-offer-driver-app/wallet_history.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE atlas_driver_offer_bpp.wallet_history ();
+
+ALTER TABLE atlas_driver_offer_bpp.wallet_history ADD COLUMN benefit_value double precision ;
+ALTER TABLE atlas_driver_offer_bpp.wallet_history ADD COLUMN campaign_id text ;
+ALTER TABLE atlas_driver_offer_bpp.wallet_history ADD COLUMN created_at timestamp with time zone NOT NULL default CURRENT_TIMESTAMP;
+ALTER TABLE atlas_driver_offer_bpp.wallet_history ADD COLUMN domain_entity_id text NOT NULL;
```
khuzema786, line 6: should be part of wallet_payments ?

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+              )
+        pure (w, acc)
+      Nothing -> createAccountAndWallet
+  where
+    lockKey = "Loyalty:Wallet:GetOrCreate:Person:" <> personId <> ":Program:" <> show programType
+    createAccountAndWallet = do
+      now <- getCurrentTime
+      accountId <- Id <$> generateGUID
+      walletId <- Id <$> generateGUID
+      let account =
```
khuzema786: Don't create account, that library is handling already. Just create wallet here and return.

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+                updatedAt = now
+              }
+      QWallet.create wallet
+      pure (wallet, account)
+
+mkLoyaltyFinanceCtx :: DWallet.Wallet -> Text -> Finance.FinanceCtx
+mkLoyaltyFinanceCtx wallet refId =
+  Finance.FinanceCtx
+    { merchantId = wallet.merchantId,
+      merchantOpCityId = fromMaybe "" wallet.merchantOperatingCityId,
```
khuzema786, line 133: make `merchantOperatingCityId` Non Maybe

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+  DWallet.Wallet ->
+  Text -> -- referenceId for the FinanceCtx (== ledger entry's reference_id)
+  Finance.FinanceM m (Maybe (Id FLE.LedgerEntry)) ->
+  m (Either FinanceError LedgerWriteOutcome)
+runLoyaltyTransfer wallet refId action = do
+  result <- Finance.runFinance (mkLoyaltyFinanceCtx wallet refId) action
+  case result of
+    Left err -> pure (Left err)
+    Right (Just entryId, _) -> pure (Right (WrittenNew entryId))
+    Right (Nothing, _) ->
```
khuzema786: When will Nothing come ?
Should we throw error in this case and break the flow or just logError ?

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+  (FBeamFlow.BeamFlow m r, MonadFlow m) =>
+  DWallet.Wallet ->
+  DWP.WalletPaymentKind ->
+  HighPrecMoney -> -- aggregated points across this (program, kind) for the transaction
+  Text -> -- domainEntityId (used as refId)
+  m (Either FinanceError LedgerWriteOutcome)
+recordLoyaltyHistory wallet kind points domainEntityId = do
+  let (refTypeEnum, source, dest) = historyLedgerLegs kind
+      refType = loyaltyReferenceTypeText refTypeEnum
+  entriesForWallet <- findEntriesForWallet kind refType domainEntityId wallet.accountId
```
khuzema786, line 206: Here upsert don't handle...only call `runLoyaltyTransfer` when state transitions from `/= CHARGED` to `== CHARGED`

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+            <> burn.programId
+            <> " not configured for merchant "
+            <> merchantId
+      Just programType -> do
+        let totalPoints = sum (map (.points) burn.burnOptions)
+        when (totalPoints > 0) $ do
+          (wallet, _acc) <-
+            getOrCreateWalletForPerson personId programType burn.programId currency merchantId mbMerchantOperatingCityId
+          void $ upsertWalletHistory wp wallet programType DWP.BURN Nothing totalPoints Nothing domainEntityId
+          processProgramLedger wallet DWP.BURN totalPoints domainEntityId isOrderCharged burn.reversedPoints
```
khuzema786, line 417: store this reversal also somewhere in WalletHistort

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
```diff
@@ -0,0 +1,546 @@
...
+              if null earn.campaigns
+                then [(Nothing, earn.points)]
+                else map (\c -> (c.campaignId, c.points)) earn.campaigns
+            mbBenefit = case kind of
+              DWP.TOPUP -> Just (max 0 (earn.points - order.amount))
+              DWP.CASHBACK -> Nothing
+              DWP.BURN -> Nothing
+        forM_ buckets $ \(mbCampaignId, pts) ->
+          void $ upsertWalletHistory wp wallet programType kind mbCampaignId pts mbBenefit domainEntityId
+        processProgramLedger wallet kind earn.points domainEntityId isOrderCharged earn.reversedPoints
```
khuzema786, line 400: store this reversal also somewhere in WalletHistort

khuzema786 (changes requested):

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: I am getting this merchantOperatingCityId from the paymentOrder.
In that we have it as maybe so I am keeping it as maybe

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: updated this one bro

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: I have udpated this thing to logError only.
transfer function return nothing when amount is less that or equal to 0
or emitLedgerEntries is false

Backend/dev/migrations-read-only/dynamic-offer-driver-app/wallet_history.sql
Hari-Krishna-Dupaguntla: add domain entity id in wallet payments

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: added reversed points

Backend/lib/payment/src/Lib/Payment/Wallet/Service.hs
Hari-Krishna-Dupaguntla: add reversed points

khuzema786 (dismissed):

khuzema786 (approved):

PR #14591, opened 2026-04-29
backend/feat/otp-ride-offers

yashikakaushik:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Ride quotes now include customer-specific ride-hailing offers when enabled per city, using service-tier and fare details to compute applicable discounts and taxes.
  * Offer retrieval is resilient: failures during offer lookup no longer block quote creation—quotes are returned without offers if needed.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Quote.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Quote.hs
@@ -41,6 +41,7 @@ import qualified Domain.Action.UI.MerchantPaymentMethod as DMPM
 import qualified Domain.SharedLogic.RideDiscount as RD
 import Domain.Types.Booking
 import Domain.Types.Booking as DBooking
+import Domain.Types.BppDetails (BppDetails)
 import qualified Domain.Types.BookingCancellationReason as SBCR
 import Domain.Types.CancellationReason
 import Domain.Types.FRFSRouteDetails
@@ -51,6 +52,7 @@ import qualified Domain.Types.Merchant as DM
 import qualified Domain.Types.MerchantOperatingCity as DMOC
 import qualified Domain.Types.Person as Person
 import qualified Domain.Types.Quote as SQuote
+import qualified Domain.Types.QuoteBreakup as DQB
 import qualified Domain.Types.RideStatus as DRide
 import Domain.Types.RiderConfig (VehicleServiceTierOrderConfig)
 import qualified Domain.Types.RiderPreferredOption as DRPO
@@ -64,7 +66,6 @@ import Kernel.Beam.Functions
 import Kernel.External.Maps.Types
 import Kernel.Prelude hiding (whenJust)
 import Kernel.Storage.Esqueleto (EsqDBReplicaFlow)
-import Kernel.Storage.Hedis as Hedis
 import qualified Kernel.Storage.Hedis as Redis
 import Kernel.Streaming.Kafka.Producer.Types (KafkaProducerTools)
 import Kernel.Streaming.Kafka.Topic.PublicTransportQuoteList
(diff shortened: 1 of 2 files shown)


khuzema786:

khuzema786:
I'll review PR #14591 directly from the context provided. Let me fetch the details and diff now.Now let me explore the related files in the workspace for deeper context.Now let me look at a few more specific things before writing the final review.Now I have enough context. Let me write the full review.

---

## Code Review: PR #14591 — `backend/feat/otp-ride-offers`

**Summary:** This PR extends the existing offer-fetching logic (already present in `getEstimates`) to also apply to quote responses (`getOffers`). It adds `customerOffers :: Maybe CumulativeOfferResp` to `QuoteAPIEntity` and introduces a `mkQuoteAPIEntitiesWithOffers` helper. The approach mirrors the existing estimate flow closely, which is good — but there are two significant issues worth fixing before merge.

---

### Critical

#### 1. Type constraint mismatch — likely compile error
**File:** `Domain/Action/UI/Quote.hs:255` | **Confidence: 85**

`getOffers` is declared with:
```haskell
getOffers :: (HedisFlow m r, CacheFlow m r, EsqDBFlow m r, EsqDBReplicaFlow m r) => SSR.SearchRequest -> m [OfferRes]
```

But the new `mkQuoteAPIEntitiesWithOffers` (in its `where` clause) calls:
- `SOffer.offerListWithBasket` — requires `MonadFlow m, EncFlow m r, ServiceFlow m r, BeamFlow m r`
- `SOffer.mkCumulativeOfferResp` — requires `MonadFlow m, EncFlow m r, BeamFlow m r, ClickhouseFlow m r`

These constraints are **not** in `getOffers`'s signature and are not implied by `HedisFlow`. Note that `getEstimates` avoids this by using the concrete `Flow` type rather than a polymorphic `m`.

**Fix:** Either switch `getOffers` to use `Flow` directly (like `getEstimates`):
```haskell
getOffers :: SSR.SearchRequest -> Flow [OfferRes]
```
or extend the type signature with all needed constraints. The `Flow` approach is simpler and consistent with how `getEstimates` is written.

---

### Warnings

#### 2. Duplicate `productId` keys in basket — silently broken for multi-BPP results
**File:** `Domain/Action/UI/Quote.hs` (new `mkQuoteAPIEntitiesWithOffers`) | **Confidence: 85**

`productId` is derived as `show q.vehicleServiceTierType` (e.g., `"SEDAN"`). Multiple BPPs can return quotes for the same tier. This means `products` may contain duplicate `("SEDAN", price1)`, `("SEDAN", price2)` entries sent to `offerListWithBasket`.

Downstream, `Map.fromList productOffers` silently drops all but the last entry for a duplicate key. If `offerListWithBasket` returns one result per productId, every quote for `"SEDAN"` will correctly pick up that one offer — but if the duplicate products cause unexpected behavior in the offer service (e.g., double-counting for fraud checks), the results will be incorrect.

In `getEstimates` this is less risky because there is typically one estimate per tier. In `getOffers`, multiple SEDAN quotes from competing BPPs are the norm.

**Fix:** Deduplicate `products` before sending to `offerListWithBasket`:
```haskell
products = nubBy (\(a, _) (b, _) -> a == b) $
  map (\(_, (productId, _, price)) -> (productId, price)) quoteEntitiesWithCtx
```

#### 3. Error swallowed silently on offer fetch failure
**File:** `Domain/Action/UI/Quote.hs` (new `mkQuoteAPIEntitiesWithOffers`) | **Confidence: 75**

```haskell
>>= \case
  Left _ -> pure []   -- error discarded
  Right r -> pure r
```

The `Left` case drops the error with no log entry. The equivalent in `getEstimates` has the same issue. Degraded-mode silencing is fine for resilience, but without a log the on-call team has no signal when the offer service is degrading.

**Fix:**
```haskell
>>= \case
  Left err -> do
    logWarning $ "getOffers:offerListWithBasket failed: " <> show err
    pure []
  Right r -> pure r
```

#### 4. `DOrder.RideHailing` hardcoded for Rental and MeterRide branches
**File:** `Domain/Action/UI/Quote.hs` (new `mkQuoteAPIEntitiesWithOffers`) | **Confidence: 65**

Both the `Just _` (Rental/OnDemand) and `Nothing` (MeterRide/Rental) branches call `offerListWithBasket` with `DOrder.RideHailing` as `PaymentServiceType`. For rental and meter-ride searches this may return irrelevant offers or fail fraud checks that gate on service type.

If this is intentional (all rider-side searches are `RideHailing` from the offer service's POV), a comment explaining the choice would prevent future confusion.

---

### Suggestion

#### 5. Significant code duplication between `mkQuoteAPIEntitiesWithOffers` and `getEstimates`
**Confidence: 90 (it IS duplicated)**

The offer-fetch logic (breakup parsing, `offerBaseAmount` computation, `withTryCatch`, `Map.fromList`, `mkCumulativeOfferResp` call) is nearly identical between `getEstimates` and the new `mkQuoteAPIEntitiesWithOffers`. Any future change to offer policy (e.g., switching the `DOrder.PaymentServiceType`, changing the basket format) will need to be applied in two places.

Consider extracting a shared helper in `SharedLogic/Offer.hs` once this works end-to-end. Not a blocker for this PR, but worth a TODO comment.

---

### What's Good

- The `enableRideHailingOffers` flag is correctly read once at the top of `getOffers` and shared between both branches — no duplicated config lookups.
- Using `withTryCatch` ensures offer failures never break the quote flow — correct resilience posture.
- `customerOffers = Nothing` default in `makeQuoteAPIEntity` is correct; it makes the API additive and backward-compatible with clients that don't consume the field.
- The `offerBaseAmount` logic correctly excludes tolls, cancellation fees, and parking charges from the discountable base, consistent with `getEstimates`.

---

**Verdict:** Two issues should be resolved before merge — the constraint mismatch is a likely compile error, and the duplicate-productId problem can cause silently incorrect offer lookup for multi-BPP scenarios. The rest are low-risk. Would suggest running `cabal build rider-app` to confirm the constraint issue before proceeding.

khuzema786 (approved):

khuzema786 (approved):

PR #14641, opened 2026-05-04
Backend/enh/driver info sync

H-C-21:
## Summary by CodeRabbit

* **Chores**
  * Added optional secondary long‑term Redis; startup/shutdown handle connection lifecycle and log failures.
  * Standardized Redis LTS environment wiring across the codebase for consistent runtime integration.

* **Refactor**
  * Driver-pool sync now merges from a fallback store when primary is missing, writes merged data, deletes fallback entry, and logs if nothing found.
  * Added a conservative default driver-pool record for missing entries.


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool/LTSDataSync.hs
```diff
@@ -146,22 +146,34 @@ emptyUpdate =
...
+      DPD.setDriverPoolData merged
+    Nothing -> do
+      mbExisting' <- Redis.withSecondaryLTSRedis $ Redis.safeGet (DPD.driverPoolDataKey driverId)
+      case mbExisting' of
+        Just existing' -> do
+          let merged = applyUpdate update existing'
+          DPD.setDriverPoolData merged
+          Redis.withSecondaryLTSRedis $ Redis.del (DPD.driverPoolDataKey driverId)
+        Nothing -> do
+          let merged = applyUpdate update (DPD.defaultDriverPoolData driverId)
```
hkmangla: Don't make default data.. it can cause issues. Please fetch it using fetch function

hkmangla (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverPool/LTSDataSync.hs
H-C-21: Removed the default data case, Also we don't need to do fetch data  from db either, since if cloud is switched, we'll keep reading older cloud details until any update happens in which case we are deleting it from the other cloud and setting latest value in current cloud.

hkmangla (dismissed):

hkmangla (approved):

hkmangla:
Not required.

PR #14758, opened 2026-05-12
updated juspay webhook to handle multiple transaction

Hari-Krishna-Dupaguntla:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end Postman integration collection and local environment for multi-transaction payment webhook flows, validating two distinct transactions per webhook.

* **Refactor**
  * Payment status processing updated to iterate, process, and persist multiple transactions per order reliably.

* **Enhancement**
  * Loyalty responses enriched with program-type resolution; wallet-history bucket derivation adjusted for top-ups and campaign earns.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14758)


khuzema786:
@ClaudeWeb Can you review this PR

khuzema786:
I'll fetch the PR details and diff for PR #14758 now.Let me try fetching the PR data via the REST API instead.Now let me fetch the raw diff and look at the key changed files.Let me read the full diff and the key files related to the webhook change.Let me look at the remaining files - particularly the payment status/webhook handler and the wallet history query.Let me look at the full SharedLogic/Payment.hs diff and any Postman collection changes.Let me look at the full rider-platform Payment.hs diff (the webhook handler for multiple transactions) and look at the FRFS status change around line 417.Now I have a full picture of the PR. Let me compile the review.

---

## PR #14758 — Updated Juspay Webhook to Handle Multiple Transactions

**Verdict**: The PR delivers three distinct features bundled together with mostly correct implementation, but has two backward-compatibility breaks that go unacknowledged in the checklist, and one non-deterministic selection that could cause subtle bugs in production.

---

### Critical

#### 1. Breaking API change: `programId :: Maybe Text` → `programId :: Text`
**File**: `Backend/app/rider-platform/rider-app/Main/src/API/Types/UI/Payment.hs`

```diff
-    programId :: Maybe Text
+    programId :: Text
```

Making a previously optional field required is a **breaking change** for all existing mobile/web clients that either:
- Omit `programId` from the JSON body, or
- Send `"programId": null`

Both will now fail deserialization with a 400/422. The PR checklist item _"This PR contains API breaking changes"_ is **unchecked**.

**Fix**: Either keep `Maybe Text` and unwrap downstream, or explicitly check the box and version the API endpoint.

---

#### 2. Breaking API response shape: `/wallet/loyaltyInfo`
**File**: `Backend/app/rider-platform/rider-app/Main/src-read-only/API/Action/UI/Loyalty.hs`

```diff
-type API = ... Post '[JSON] Kernel.External.Wallet.Interface.Types.LoyaltyInfoResponse
+type API = ... Post '[JSON] API.Types.UI.Loyalty.LoyaltyInfoAPIResponse
```

The response was previously the flat `LoyaltyInfoResponse` object. It is now a wrapper with three fields (`loyaltyInfo`, `avgTicketPriceLast30Days`, `weeklyWalletCredits`). Any client that was parsing the loyalty response will break. Again, the breaking change checklist item is **unchecked**.

---

### Warning

#### 3. Non-deterministic transaction selection in `frfsBookingStatus`
**File**: `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSStatus.hs`, line ~490

```diff
-        [transaction] -> return transaction.id
-        _ -> throwError $ InvalidRequest "Multiple successful transactions found"
+        (transaction : _) -> return transaction.id
```

This is the core fix from the PR title. When multiple `CHARGED` transactions exist for the same order, `filter` preserves the DB query's return order — which has no guaranteed `ORDER BY`. The "first" transaction picked could change between invocations depending on DB row ordering, index scans, or replica lag.

**Fix**: Sort transactions by `createdAt` descending before taking the head, or select `maximumBy (comparing .createdAt)`:

```haskell
let successTransactions = filter (\t -> t.status == Payment.CHARGED) transactions
case NE.nonEmpty successTransactions of
  Nothing -> throwError $ InvalidRequest "No successful transaction found"
  Just ts -> return (NE.last (NE.sortWith (.createdAt) ts)).id
```

---

#### 4. Unbounded DB scan in `getAvgTicketPriceLast30Days`
**File**: `Backend/app/rider-platform/rider-app/Main/src/Storage/Queries/FRFSTicketBookingExtra.hs`

```haskell
findAllConfirmedByRiderIdSince riderId merchantId merchantOperatingCityId since =
  findAllWithKV [...]  -- no LIMIT
```

For a frequent commuter over 30 days this could return hundreds of rows just to compute a sum. Since the caller only needs `sum(.totalPrice.amount)`, consider pushing the aggregation into the query (sum+count at DB level), or at minimum add a sanity `LIMIT` to bound memory usage.

---

#### 5. Silent program-type degradation in `getLoyaltyInfo`
**File**: `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs`

```haskell
mbProgramMap <- TPayment.loadLoyaltyProgramMap merchantId merchantOperatingCityId
let resolveProgramType pid = T.pack . show <$> (Map.lookup pid =<< mbProgramMap)
    enrich p = p {WalletTypes.programType = resolveProgramType (p.id_)}
```

If `loadLoyaltyProgramMap` returns `Nothing` (config missing, service down, etc.), `resolveProgramType` silently sets `programType = Nothing` for every program. The response is returned without any error or log warning. Clients relying on `programType` being populated will silently see nulls.

**Fix**: Add a log warning when `mbProgramMap` is `Nothing` so this is observable:

```haskell
when (isNothing mbProgramMap) $
  logWarning "loadLoyaltyProgramMap returned Nothing — programType will be absent in loyalty response"
```

---

### Suggestion

#### 6. `getWeeklyWalletCreditsLast30Weeks` — `since` filter is computed but unused in first query
**File**: `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/LoyaltyEnrichment.hs`

```haskell
let since = Time.addUTCTime (negate $ fromIntegral (weeksInWindow * 7 * 86400 :: Int)) now
wallets <- QWalletExtra.findAllByPersonAndMerchantOpCity ...  -- ← `since` not passed here
```

`since` is only applied to the history query, not the wallet lookup. This is probably fine (wallets don't expire), but worth a comment explaining why.

---

#### 7. `weekStartUtc` uses ISO week Monday correctly — but only returns weeks with data

`aggregateByWeek` returns only weeks that have at least one credit. If the UI expects a continuous 30-week series (with zeros for empty weeks), the caller will need to backfill gaps. This should be documented or handled on the client side.

---

### Non-Issues (cosmetic only)
- Import reordering across 8+ files — likely automated formatter output, no logic change.
- Nix flake indentation normalization — pure whitespace.
- `PARTIAL_CHARGED -> ...` additions in `Handler.hs`, `WalletTransaction.hs`, `BBPS.hs`, `FRFSStatus.hs`, `Payment.hs` — consistently handled across all status-mapping sites. ✓

---

### Checklist Items Missing
- [ ] "This PR contains API breaking changes" should be **checked** (two breaking changes)
- [ ] No tests added for the multi-transaction selection behavior (the core fix)
- [ ] The PR description body has no description of what changed or why

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -122,6 +126,7 @@ fallbackOrderStatusHandler paymentStatusResp = do
         Payment.JUSPAY_DECLINED -> DPayment.FulfillmentPending
         Payment.AUTHORIZATION_FAILED -> DPayment.FulfillmentPending
         Payment.AUTHENTICATION_FAILED -> DPayment.FulfillmentPending
+        Payment.PARTIAL_CHARGED -> DPayment.FulfillmentPending
```
khuzema786, line 129: Check if refunds for this would be handled.

Backend/lib/payment/src/Lib/Payment/Domain/Action.hs
```diff
@@ -1566,41 +1566,55 @@ orderStatusService merchantOpCityId personId orderId orderStatusCall = do
...
                 cardType = (.cardType) =<< card,
-                surchargeAmount = (.surchargeAmount) =<< txnDetail,
-                taxAmount = (.taxAmount) =<< txnDetail,
-                netAmount = (.netAmount) =<< txnDetail,
-                epgTxnId = (.epgTxnId) =<< paymentGatewayResponse,
+                surchargeAmount = (.surchargeAmount) =<< txnObj.txnDetail,
+                taxAmount = (.taxAmount) =<< txnObj.txnDetail,
+                netAmount = (.netAmount) =<< txnObj.txnDetail,
+                epgTxnId = (.epgTxnId) =<< txnObj.paymentGatewayResponse,
                 cardBrand = (.cardBrand) =<< card,
```
khuzema786, line 1598: In multiple txn then card things should be inside txnList

Backend/lib/payment/src/Lib/Payment/Domain/Action.hs
```diff
@@ -1566,41 +1566,55 @@ orderStatusService merchantOpCityId personId orderId orderStatusCall = do
...
                 cardIsin = (.cardIsin) =<< card,
                 cardLastFourDigits = (.lastFourDigits) =<< card,
                 cardIssuer = (.cardIssuer) =<< card,
-                authorizationDateTime = (.created) =<< paymentGatewayResponse,
+                authorizationDateTime = (.created) =<< txnObj.paymentGatewayResponse,
                 captureDateTime = dateCreated,
                 vpa = payerVpa,
```
khuzema786, line 1604: In multiple txn then payerVpa things should be inside txnList

khuzema786 (approved):

Backend/lib/payment/src/Lib/Payment/Domain/Action.hs
```diff
@@ -1827,29 +1852,32 @@ juspayWebhookService merchantOpCityId resp respDump = do
...
+
+      forM_ txnList $ \txnObj -> do
+        let orderTxn = mkOrderTxn txnObj
+        maybe
+          (updateOrderTransaction merchantOpCityId order orderTxn Nothing)
+          ( \transactionUUID' ->
+              Redis.whenWithLockRedis (txnProccessingKey transactionUUID') 60 $ updateOrderTransaction merchantOpCityId order orderTxn $ Just respDump
+          )
+          txnObj.txnUuid
       mapM_ (upsertRefundStatus merchantOpCityId order) refunds
```
khuzema786, line 1881: If refunds happen then one order can have multiple refunds, whether that is handled correctly ?

khuzema786 (approved):

PR #14791, opened 2026-05-14
ci: push builds to new system

AVBharath10:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **Chores**
  * CI updated to use more specific runners and incremental builds with per-runner caches, restoring cached build state and staging built executables, configs, and API specs to improve build reliability and cache reuse.

* **Chores / Deployment**
  * Container packaging now supports using locally staged binaries and performs an impure build path to produce consistent Docker images that include the staged artifacts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14791?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/cabal-main-push.yaml
+++ b/.github/workflows/cabal-main-push.yaml
@@ -50,7 +50,7 @@ jobs:
 
       - name: Restore cabal cache
         run: |
-          CACHE_ROOT=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}
+          CACHE_ROOT=$HOME/ny-cabal-cache/${RUNNER_NAME}
           KEY_FILE="$CACHE_ROOT/.cache-key"
           CURRENT_KEY="${{ steps.cache-key.outputs.key }}"
 
@@ -67,7 +67,7 @@ jobs:
         run: |
           set -o pipefail
           nix develop .#backend --command bash -c "
-            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            export CABAL_DIR=$HOME/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
             cd Backend
             cabal build all --enable-optimization=0 --ghc-options='-funfolding-use-threshold20 -fno-cse -fmax-simplifier-iterations1 -fno-specialise-aggressively' 2>&1
           " | while IFS= read -r line; do
@@ -76,12 +76,12 @@ jobs:
 
       - name: Collect binaries to staging
         run: |
-          STAGING=/home/khuzema/ny-staging/$GITHUB_RUN_ID
+          STAGING=$HOME/ny-staging/$GITHUB_RUN_ID
           rm -rf "$STAGING"
           mkdir -p "$STAGING/bin" "$STAGING/opt/app"
 
           nix develop .#backend --command bash -c "
-            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            export CABAL_DIR=$HOME/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
             cd Backend
             cabal install all:exes --installdir=$STAGING/bin --overwrite-policy=always
           "
@@ -107,7 +107,7 @@ jobs:
         env:
           DOCKER_REGISTRY: ghcr.io
           DOCKER_TAG_PREFIX: "cabal-master-"
-          LOCAL_BINARIES_PATH: /home/khuzema/ny-staging/${{ github.run_id }}
+          LOCAL_BINARIES_PATH: $HOME/ny-staging/${{ github.run_id }}
         run: |
           nix build .#packages.${{ matrix.system }}.dockerImage --impure
 
@@ -121,7 +121,7 @@ jobs:
 
       - name: Cleanup staging dir
         if: always()
-        run: rm -rf /home/khuzema/ny-staging/${{ github.run_id }}
+        run: rm -rf $HOME/ny-staging/${{ github.run_id }}
   
   analyze-master-build-times:
     name: Analyze Master Build Times
@@ -191,7 +191,7 @@ jobs:
 
       - name: Restore cabal cache
         run: |
-          CACHE_ROOT=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}
+          CACHE_ROOT=$HOME/ny-cabal-cache/${RUNNER_NAME}
           KEY_FILE="$CACHE_ROOT/.cache-key"
           CURRENT_KEY="${{ steps.cache-key.outputs.key }}"
 
@@ -208,7 +208,7 @@ jobs:
         run: |
           set -o pipefail
           nix develop .#backend --command bash -c "
-            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            export CABAL_DIR=$HOME/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
             cd Backend
             cabal build all 2>&1
           " | while IFS= read -r line; do
@@ -217,12 +217,12 @@ jobs:
 
       - name: Collect binaries to staging
         run: |
-          STAGING=/home/khuzema/ny-staging/$GITHUB_RUN_ID
+          STAGING=$HOME/ny-staging/$GITHUB_RUN_ID
           rm -rf "$STAGING"
           mkdir -p "$STAGING/bin" "$STAGING/opt/app"
 
           nix develop .#backend --command bash -c "
-            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            export CABAL_DIR=$HOME/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
             cd Backend
             cabal install all:exes --installdir=$STAGING/bin --overwrite-policy=always
           "
@@ -248,7 +248,7 @@ jobs:
         env:
           DOCKER_REGISTRY: ghcr.io
           DOCKER_TAG_PREFIX: "cabal-prod-"
-          LOCAL_BINARIES_PATH: /home/khuzema/ny-staging/${{ github.run_id }}
+          LOCAL_BINARIES_PATH: $HOME/ny-staging/${{ github.run_id }}
         run: |
           nix build .#packages.${{ matrix.system }}.dockerImage --impure
 
@@ -262,7 +262,7 @@ jobs:
 
       - name: Cleanup staging dir
         if: always()
-        run: rm -rf /home/khuzema/ny-staging/${{ github.run_id }}
+        run: rm -rf $HOME/ny-staging/${{ github.run_id }}
 
   analyze-prod-build-times:
     name: Analyze Prod Build Times
--- a/.github/workflows/cabal-main-push.yaml
+++ b/.github/workflows/cabal-main-push.yaml
@@ -1,4 +1,4 @@
-name: Nix CI (Main Push)
+name: Nix CI (Cabal Push Main)
 
 on:
   push:
@@ -32,7 +32,7 @@ jobs:
       matrix:
         include:
           - system: x86_64-linux
-            host: x86_64-linux
+            host: x86_64-linux-ny-push-main
             arch: amd64
             name: Master Local Build
             continue-on-error: false
@@ -42,27 +42,56 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Build all flake outputs (Master Local mode)
+      - name: Compute cache key
+        id: cache-key
+        run: |
+          KEY=$(cat flake.lock Backend/cabal.project | sha256sum | cut -c1-16)
+          echo "key=$KEY" >> $GITHUB_OUTPUT
+
+      - name: Restore cabal cache
+        run: |
+          CACHE_ROOT=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}
+          KEY_FILE="$CACHE_ROOT/.cache-key"
+          CURRENT_KEY="${{ steps.cache-key.outputs.key }}"
+
+          if [ -f "$KEY_FILE" ] && [ "$(cat $KEY_FILE)" != "$CURRENT_KEY" ]; then
+            echo "Cache key changed, busting dist-newstyle"
+            rm -rf "$CACHE_ROOT/dist-newstyle"
+          fi
+
+          mkdir -p "$CACHE_ROOT/dist-newstyle" "$CACHE_ROOT/.cabal-dir"
+          echo "$CURRENT_KEY" > "$KEY_FILE"
+          ln -sfn "$CACHE_ROOT/dist-newstyle" Backend/dist-newstyle
+
+      - name: Build (incremental cabal)
         run: |
           set -o pipefail
-          NIX_LOCAL_BUILD="1" nix build github:srid/devour-flake#default --impure --system ${{ matrix.system }} -L --print-out-paths --no-write-lock-file --override-input flake . --out-link ./result 2>&1 | while IFS= read -r line; do
+          nix develop .#backend --command bash -c "
+            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            cd Backend
+            cabal build all --enable-optimization=0 --ghc-options='-funfolding-use-threshold20 -fno-cse -fmax-simplifier-iterations1 -fno-specialise-aggressively' 2>&1
+          " | while IFS= read -r line; do
             printf '%s %s\n' "$(date -u '+%Y-%m-%dT%H:%M:%S.%NZ')" "$line"
           done | tee master-build-log.txt
 
-      - name: Push to Cache (Local mode)        
-        continue-on-error: true
-        env:
-          ATTIC_TOKEN: ${{ secrets.ATTIC_TOKEN }}
-          CACHIX_AUTH_TOKEN: ${{ secrets.CACHIX_AUTH_TOKEN }}
+      - name: Collect binaries to staging
         run: |
-          PATHS=$(jq -r '.[].outPaths[]' ./result)
-          echo "Paths to push: $(echo "$PATHS" | wc -l)"
+          STAGING=/home/khuzema/ny-staging/$GITHUB_RUN_ID
+          rm -rf "$STAGING"
+          mkdir -p "$STAGING/bin" "$STAGING/opt/app"
 
-          # Try attic, fall back to cachix on any failure
-          attic login cache-nixos-asia https://cache.nixos.asia "$ATTIC_TOKEN" && \
-            echo "$PATHS" | attic push cache-nixos-asia:oss --stdin && \
-            echo "✅ Pushed to attic" || \
-            { echo "⚠️ Attic failed, falling back to cachix..."; echo "$PATHS" | cachix push nammayatri; }
+          nix develop .#backend --command bash -c "
+            export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            cd Backend
+            cabal install all:exes --installdir=$STAGING/bin --overwrite-policy=always
+          "
+
+          for f in $STAGING/bin/*; do
+            ln -sf "$f" "$STAGING/opt/app/$(basename $f)"
+          done
+
+          cp -rT Backend/dhall-configs "$STAGING/opt/app/dhall-configs"
+          cp -rT Backend/swagger "$STAGING/opt/app/swagger"
 
       - name: Upload master build log
         continue-on-error: true
@@ -76,10 +105,11 @@ jobs:
         id: docker
         if: matrix.system == 'x86_64-linux' || matrix.system == 'aarch64-linux'
         env:
-          DOCKER_REGISTRY: ghcr.io          
-          DOCKER_TAG_PREFIX: "master-"          
+          DOCKER_REGISTRY: ghcr.io
+          DOCKER_TAG_PREFIX: "cabal-master-"
+          LOCAL_BINARIES_PATH: /home/khuzema/ny-staging/${{ github.run_id }}
         run: |
-          NIX_LOCAL_BUILD="1" nix build .#packages.${{ matrix.system }}.dockerImage --impure
+          nix build .#packages.${{ matrix.system }}.dockerImage --impure
 
           BASE_IMAGE="$(nix eval --raw .#packages.${{ matrix.system }}.dockerImage.imageName --impure):$(nix eval --raw .#packages.${{ matrix.system }}.dockerImage.imageTag --impure)"
           echo "base_image=${BASE_IMAGE}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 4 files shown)


khuzema786:

```
nix build github:srid/devour-flake#default --system ${{ matrix.system }} -L --print-out-paths --no-write-lock-file --override-input flake . --out-link ./result 2>&1 | while IFS= read -r line; do                                                                                                                                                                                       
      printf '%s %s\n' "$(date -u '+%Y-%m-%dT%H:%M:%S.%NZ')" "$line"                                                                                                                     
 done | tee prod-build-log.txt
```

If we replace this with cabal build all inside nix develop .#backend shell like in this PR, then for prod build what all top things we should take care.

khuzema786:
I'll fetch the PR details and diff to give you a thorough review.Let me look at the docker.nix and flake files for deeper context.Let me use the GitHub MCP tool to get more context.Now I have everything I need for a thorough review. Let me write the full analysis.

---

## PR #14791 Review — `ci: push builds to new system`

**Verdict:** The strategic shift (incremental cabal cache for speed) is sound, but there are **two critical bugs** that will produce broken Docker images and one concurrency hazard that will cause silent production contamination under load. These need to be fixed before merge.

---

## Critical

### 1. Broken symlinks in Docker image (`/opt/app/` → dead host paths)
**Confidence: 92 | Files: `nix-main-push.yaml`, `nix-prod-hot-fix-push.yaml`, `docker.nix`**

The "Collect binaries" step creates symlinks that point to absolute host paths:
```bash
for f in $STAGING/bin/*; do
  ln -sf "$f" "$STAGING/opt/app/$(basename $f)"
  # Creates: /home/khuzema/ny-staging/opt/app/rider-app-exe -> /home/khuzema/ny-staging/bin/rider-app-exe
done
```

`builtins.path { path = /. + localBinariesPath; }` then copies this entire directory tree (including the symlinks) into the Nix store. Inside the Docker image, `/home/khuzema/ny-staging/bin/` does not exist, so every symlink in `/opt/app/` is broken at runtime.

The `extraCommands` test (`ls opt/app/rider-app-exe`) passes because `ls` resolves the name, not the target — but any process trying to execute through that path will fail.

The `Cmd` points at `${nammayatriBinaries}/bin/rider-app-exe` (the direct binary, not the symlink), so the container *starts*, but any service-discovery or wrapper that looks in `/opt/app/` gets a broken link.

**Fix:** Don't create symlinks; copy the binaries directly:
```bash
for f in "$STAGING/bin/"*; do
  cp "$f" "$STAGING/opt/app/$(basename "$f")"
done
```

---

### 2. Shared staging directory — concurrent workflow runs corrupt each other
**Confidence: 88 | File: `nix-main-push.yaml`, `nix-prod-hot-fix-push.yaml`**

`STAGING=/home/khuzema/ny-staging` is a fixed host-level path. The step starts with `rm -rf "$STAGING"`, then rebuilds it. If two workflow runs execute on the same runner concurrently (two pushes to `main` in quick succession, or a `workflow_dispatch` overlapping a push), one run's `rm -rf` wipes the other's staging area mid-build — potentially packaging empty or partially-built binaries into a prod Docker image with no error surfaced.

**Fix:** Isolate per run:
```bash
STAGING=/home/khuzema/ny-staging/${{ github.run_id }}
```
Add cleanup at the end of the job, or rely on OS-level tmpdir cleanup.

---

### 3. Runtime dependency gap — cabal binaries are dynamically linked but the Docker closure has no mechanism to include their `.so` deps
**Confidence: 82 | File: `docker.nix`**

The old `nix build`-based approach computed the full closure of each binary (all linked `.so` files — `libgmp`, `libz`, `libpq`, Haskell RTS libs, etc.) and automatically bundled them. `builtins.path` just copies files; it has no closure-tracking semantics.

When `nix develop .#backend` builds the binary, GHC links dynamically against the Nix-provided libraries (e.g., `/nix/store/xxxx-gmp-6.x/lib/libgmp.so`). Those store paths are burned into the binary's RPATH. The Docker image will not contain those paths unless they appear in `copyToRoot`.

The image will likely `SIGKILL` at startup with `cannot open shared object file`.

**Fix options (pick one):**
- Keep the Nix package (`self'.packages.nammayatri`) as the source of truth for the Docker image; use the local cabal cache only to speed up the Nix build.
- Or: statically link with `--ghc-options=-optl-static -optl-pthread` and add `--enable-executable-static` to the cabal flags, then the binary is self-contained.
- Or: use `pkgs.lddInputs` / `pkgs.makeWrapper` to explicitly pull the runtime deps into the image.

---

## Warning

### 4. `for f in $STAGING/bin/*` — unquoted glob with no empty-dir guard
**Confidence: 85 | File: both workflow files**

If `cabal install` produces no executables (partial failure, wrong target), the glob `$STAGING/bin/*` expands to the literal string `$STAGING/bin/*`. The subsequent `ln -sf` creates a symlink named `*` in `/opt/app/` — silently wrong, no error.

```bash
# Fix: use nullglob or check explicitly
shopt -s nullglob
for f in "$STAGING/bin/"*; do
  [[ -f "$f" ]] && cp "$f" "$STAGING/opt/app/$(basename "$f")"
done
```

---

### 5. `cabal install` missing `--enable-optimization=2`
**Confidence: 75 | File: both workflow files**

`cabal build all --enable-optimization=2` compiles with full optimization. `cabal install all:exes` does not pass `--enable-optimization=2`. If cabal decides any artifact is stale (flag mismatch), it silently recompiles at `O1` (default). The prod binary may be slower than expected.

```bash
cabal install all:exes \
  --enable-optimization=2 \
  --installdir=$STAGING/bin \
  --overwrite-policy=always
```

---

### 6. `prodHotPush-MMFRFS` is a dead reference in the docker-push condition
**Confidence: 80 | File: `nix-prod-hot-fix-push.yaml`**

The `check-docker-push` step includes `refs/heads/prodHotPush-MMFRFS` in its condition, but the workflow only triggers on `prodHotPush-Common`, `prodHotPush-BAP`, and `prodHotPush-BPP`. That branch can never trigger this workflow, making the condition dead code. Either add the branch to `on.push.branches` or remove it from the condition.

---

### 7. Hardcoded `/home/khuzema/` on self-hosted runner
**Confidence: 80 | File: both workflow files**

Paths like `/home/khuzema/ny-cabal-cache/` and `/home/khuzema/ny-staging` are baked in. Using `$HOME` or a configurable env var would make this more robust to runner user changes:
```yaml
env:
  NY_CACHE_BASE: /home/khuzema/ny-cabal-cache
  NY_STAGING_BASE: /home/khuzema/ny-staging
```

---

## On Your Specific Question

> If we replace `nix build … devour-flake` with `cabal build all` inside `nix develop .#backend`, what top things should we take care of for prod builds?

Here's the ordered checklist:

| # | Concern | Why it matters for prod |
|---|---------|------------------------|
| 1 | **Runtime deps in Docker** | `nix build` auto-includes the closure; `cabal install` doesn't. Binaries that link `libgmp`/`libpq`/GHC RTS dynamically will crash on startup in Docker unless those `.so`s are explicitly included or you go static. |
| 2 | **Static vs dynamic linking** | Decide before this merges. Static (`--enable-executable-static`) is safest for Docker portability. Dynamic requires carefully auditing `docker.nix` to include all `.so` paths. |
| 3 | **Staging dir isolation per run** | Fixed path = silent contamination between concurrent workflow runs. |
| 4 | **Cache invalidation strategy** | Incremental cabal caches produce wrong output when GHC/package versions drift. Key the cache on `ghc --version` + a hash of `cabal.project` / `cabal.lock`. The old Nix approach was immune to this (content-addressed). |
| 5 | **Stripping binaries** | `nix build` strips by default. `cabal install` may not. Add `strip $STAGING/bin/*` before packaging to keep Docker image size down. |
| 6 | **Prod vs master build — same runner, same cache** | Both `prod-build` and `master-build` share `/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/`. They run sequentially (prod needs master), but the shared cache means a failed master can leave a partial `dist-newstyle` that prod picks up without recompiling from scratch. |
| 7 | **Nix cache loss** | You dropped Attic/Cachix push. The incremental cabal cache is local-disk only, not shared across machines. If you ever add a second runner or rebuild the host, you pay full cold-build time. Consider keeping a Nix binary cache push for the nix shell itself even if the Haskell build is now cabal-based. |
| 8 | **Symlink vs copy in `/opt/app/`** | As documented above — use `cp`, not `ln -sf`. |
| 9 | **Optimization flag consistency** | Pass `--enable-optimization=2` to `cabal install` too. |
| 10 | **`extraCommands` test validity** | The `ls opt/app/rider-app-exe` test in `docker.nix` only checks existence, not executability. Add a smoke test: `` extraCommands = '' ls opt/app/rider-app-exe file opt/app/rider-app-exe | grep -q ELF ''; `` |

khuzema786:
@ClaudeWeb what would be the best approach to not affect performance that nix gave and also make builds incremental ?

khuzema786:
Great question. Let me reason through this carefully based on the PR structure and what Nix was actually providing.Let me read the Backend's Nix setup to give you a precise answer.This is a rich question. With `Backend/default.nix` fully read, I can give you a precise, project-specific answer — not a generic one.

---

## What Nix Was Actually Giving You (the real picture)

Before suggesting anything, let's be precise about what the old pipeline did and why it was correct.

### The `nammayatri` Nix package is a `symlinkJoin`

From `Backend/default.nix`:
```nix
nammayatri = pkgs.symlinkJoin {
  name = "nammayatri";
  paths = localCabalPackages;    # ← all per-package Nix derivations
  postBuild = ''
    mkdir -p $out/opt/app
    for f in `fd . $out/bin/`; do ln -s $f $out/opt/app/; done
    cp -r ${./dhall-configs} $out/opt/app/dhall-configs
    cp -r ${./swagger} $out/opt/app/swagger
  '';
};
```

Each `localCabalPackage` (e.g. `rider-app`, `dynamic-offer-driver-app`) is its own Nix derivation, cached individually in Attic. **Nix is already incremental at the per-package level.** If only `rider-app`'s source changes, only `rider-app` (and its dependents) rebuild. Everything else is a cache hit from Attic.

### The project already had `NIX_LOCAL_BUILD` fast-path

Also in `Backend/default.nix`:
```nix
localBuild = builtins.getEnv "NIX_LOCAL_BUILD" != "";
# ...
(lib.optionals localBuild [
  "--ghc-options=-O0"
  "--ghc-options=-funfolding-use-threshold20"
  "--ghc-options=-fno-cse"
  "--ghc-options=-fmax-simplifier-iterations1"
  "--ghc-options=-fno-specialise-aggressively"
])
```

The old master-build already used `NIX_LOCAL_BUILD="1" nix build …` which traded full optimization for faster Nix evaluation. Prod build used full optimization. **This mechanism is gone in the PR but is the correct lever to pull.**

### The `cac_client` dynamic lib wrapping

Every major exe (`rider-app`, `dynamic-offer-driver-app`, etc.) is wrapped via `cacConfig` using `makeWrapper` to set `LD_LIBRARY_PATH` to `cac_client/lib`. The Nix Docker image has this wrapper embedded. **A raw `cabal install` binary skips this wrapper entirely — `cac_client` will fail to load at runtime in Docker.**

---

## Why Replacing `nix build` With `cabal build` Breaks Things

| What Nix did automatically | What `cabal install` gives you |
|---|---|
| Computes full runtime closure (all `.so` deps) | Just the binary — no dep tracking |
| RPATH points into `/nix/store/…` (hermetic) | RPATH points into `/nix/store/…` on the **build runner** — those paths don't exist in Docker |
| Applies `makeWrapper` for `cac_client` `LD_LIBRARY_PATH` | Raw binary, `cac_client` dlopen fails at startup |
| `symlinkJoin` creates proper `/opt/app/` tree | Manual symlinks pointing to `/home/khuzema/…` (broken in Docker) |
| Per-package Nix cache → only rebuild changed packages | Always recompiles everything on cold runner |

---

## The Right Architecture: Two-Phase Build

The insight is: **use cabal for fast feedback, keep Nix for the Docker image.** They serve different purposes and don't need to replace each other.

```
┌─────────────────────────────────────────────────────────┐
│ Phase 1 — Incremental cabal (fast, catches errors)      │
│                                                         │
│  Restore dist-newstyle from per-runner disk cache       │
│  → cabal build all --enable-optimization=2              │
│    (incremental: only recompiles changed modules)       │
│  → repopulate dist-newstyle cache                       │
└───────────────────────┬─────────────────────────────────┘
                        │ (if compile fails → stop here, fast feedback)
                        ▼
┌─────────────────────────────────────────────────────────┐
│ Phase 2 — Nix build for Docker (correct, cached)        │
│                                                         │
│  nix build .#packages.x86_64-linux.dockerImage          │
│  → per-package cache hits from Attic for unchanged pkgs │
│  → only rebuilds packages whose source hash changed     │
│  → full closure: runtime .so deps + cac_client wrapper  │
│  → correct /opt/app/ layout with working symlinks       │
│  → push new derivations back to Attic                   │
└─────────────────────────────────────────────────────────┘
```

### Why Phase 2 is fast (when Attic is warm)

Each `rider-app`, `dynamic-offer-driver-app`, etc. is a separately hashed Nix derivation. With Attic populated from previous builds:
- A push that changes only `rider-app` source → Nix fetches ~50 packages from cache, rebuilds 1
- A push that changes a shared library → Nix rebuilds that library + its downstreams only
- Full cold build only on first run or `flake.lock` update

The PR removes the Attic push, which is precisely what would make Phase 2 fast.

---

## Concrete Workflow Changes

**Restore `NIX_LOCAL_BUILD` for master-build** (faster Nix build, -O0):
```yaml
- name: Build (incremental cabal — fast feedback)
  run: |
    # Phase 1: incremental, catches errors fast
    nix develop .#backend --command bash -c "
      export CABAL_DIR=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
      cd Backend
      cabal build all --enable-optimization=2 2>&1
    " | while IFS= read -r line; do
      printf '%s %s\n' "$(date -u '+%Y-%m-%dT%H:%M:%S.%NZ')" "$line"
    done | tee master-build-log.txt

- name: Build And Push Docker image (Master mode)
  env:
    DOCKER_TAG_PREFIX: "master-"
    NIX_LOCAL_BUILD: "1"          # ← use -O0 flags, faster Nix build
  run: |
    # Phase 2: Nix build for correct Docker closure, uses Attic cache
    nix build .#packages.${{ matrix.system }}.dockerImage --impure
    ...

- name: Push to Attic cache        # ← restore this, it's what makes Phase 2 fast
  run: |
    attic push cache-nixos-asia:oss $(nix-store -qR ./result)
```

**For prod-build**: drop `NIX_LOCAL_BUILD` (full -O2 optimization via Nix):
```yaml
- name: Build (incremental cabal — fast feedback)
  run: |
    # Phase 1: same as above — fast incremental check

- name: Build And Push Docker image (Prod mode)
  run: |
    # Phase 2: full Nix build, no NIX_LOCAL_BUILD, full -O2
    nix build .#packages.${{ matrix.system }}.dockerImage
    # push to Attic after success
```

---

## Summary of the Correct Mental Model

| Concern | Right tool |
|---|---|
| Fast incremental feedback ("did it compile?") | `cabal build` in `nix develop` |
| Correct runtime deps in Docker | `nix build .#dockerImage` |
| Speed up Nix build across runs | Attic/Cachix binary cache (per-package granularity) |
| Speed up Nix build for master (not prod) | `NIX_LOCAL_BUILD=1` (already in the codebase, sets -O0) |
| Cross-runner cache sharing | Attic (works because Nix packages are content-addressed) |

**The PR's fundamental mistake**: it treats incremental compilation and Docker correctness as the same problem. They're not — cabal solves the first, Nix solves the second. The project already had the right building blocks; the PR removed Phase 2 and replaced it with something broken. The fix is to restore Phase 2 (Nix Docker build + Attic push) and keep Phase 1 (cabal incremental) as a parallel fast-feedback step.

khuzema786:
@ClaudeWeb we want to make the package build also incremental and not build all modules from scratch even on 1 line change

khuzema786:
This is the core hard problem with Nix + Haskell. Let me read the devShell definition carefully to understand exactly what runtime deps the binaries need, then give you a precise answer.Now I have the complete picture. The project has ~50 cabal packages and uses `haskell-flake`. Let me give you the precise technical answer.

---

## The Root Problem: Where Nix's Incrementality Stops

With `haskell-flake`, Nix caches at the **cabal package level** — one Nix derivation per `.cabal` file. You already have ~50 packages, so you get ~50-package granularity for free.

But **within a single package**, Nix re-runs `cabal build` from a clean sandbox every time any source file in that package changes. It has no way to reuse `.hi` / `.o` files from a previous build. That's the gap you want to close.

---

## The Three Options, Ranked by Fit

### Option 1 — `haskell.nix` (gold standard, migration cost)

[`haskell.nix`](https://github.com/input-output-hk/haskell.nix) from IOHK compiles **each Haskell module as its own Nix derivation**. `.hi` and `.o` files are individually cached in the Nix store. A 1-line change in one module only rebuilds that module + its dependents.

- This is what IOHK uses for Cardano — a comparably large Haskell monorepo
- Integrates with Nix binary cache natively → Docker images are always correct
- **Migration from `haskell-flake` is significant** (different project configuration model, different `cabal.project` integration)
- Worth doing if build speed is a recurring pain and you can invest the time

---

### Option 2 — Cabal incremental + `autoPatchelfHook` in `docker.nix` (fixes this PR)

This is the **correct version of what this PR is trying to do.** The idea:
- Keep `cabal build` for module-level incremental compilation ✓
- Fix the Docker runtime dep problem using Nix's `autoPatchelfHook` ✓

When `cabal build` runs inside `nix develop .#backend`, every binary it produces already has RPATHs pointing to `/nix/store/…` paths (the libs from the dev shell). Those store paths exist on the build runner. The issue is they don't get into the Docker image.

`pkgs.autoPatchelfHook` solves this: it scans the binary's ELF needed-libs, finds matching `.so` files in the listed `buildInputs`, and patches the RPATH. Nix then automatically includes those store paths in the Docker image closure.

The fix in `docker.nix`:

```nix
{ self, inputs, ... }:
let
  ...
  localBinariesPath = builtins.getEnv "LOCAL_BINARIES_PATH";
  localBuild = localBinariesPath != "";
in {
  config.perSystem = { self', pkgs, lib, config, ... }:
    let
      # When using cabal-built binaries, re-derive the runtime closure
      # using the same library set the nix develop shell provides.
      # autoPatchelfHook scans each ELF binary, finds matching .so files
      # in buildInputs, and patches RPATH — so Nix includes them in the
      # Docker closure automatically.
      nammayatriBinaries = if localBuild
        then pkgs.runCommand "nammayatri-local" {
          nativeBuildInputs = [ pkgs.autoPatchelfHook pkgs.makeWrapper ];

          # These are the exact same libs the nix develop shell provides —
          # so they match what cabal linked against.
          buildInputs =
            config.haskellProjects.default.outputs.devShell.buildInputs
            ++ [ config.haskellProjects.default.outputs.finalPackages.cac_client ];

          src = /. + localBinariesPath;
        } ''
          cp -r $src $out
          chmod -R +w $out/bin

          # Re-apply the cac_client wrapper that cacConfig normally sets up.
          # cabal install skips makeWrapper, so we do it here.
          for exe in $out/bin/*; do
            wrapProgram "$exe" \
              --set LD_LIBRARY_PATH "${config.haskellProjects.default.outputs.finalPackages.cac_client}/lib"
          done
        ''
        else self'.packages.nammayatri;
    in { ... };  # rest of docker.nix unchanged
}
```

What this gives you:
- `autoPatchelfHook` patches RPATHs → all `.so` deps are in the Docker closure ✓
- `wrapProgram` restores the `cac_client` `LD_LIBRARY_PATH` that `cacConfig` normally adds ✓
- No symlinks pointing to host paths ✓
- `nix build .#dockerImage --impure` remains the image build command ✓

---

### Option 3 — Fine-grained cabal package splitting (incremental improvement, no migration)

Your ~50 cabal packages give Nix ~50 cache slots. If the slowest packages (`rider-app`, `dynamic-offer-driver-app`) are large, you can split them into finer cabal libraries.

E.g. split `rider-app` into:
- `rider-app-core` (stable domain logic — rarely changes)
- `rider-app-api` (API layer — changes often)
- `rider-app-exe` (thin executable glue)

Any change to `rider-app-api` only rebuilds `rider-app-api` and `rider-app-exe` in Nix — not `rider-app-core`. No tooling migration, pure structural change.

---

## The Cache Invalidation Problem (Critical for all options)

The current PR's cabal cache has **no invalidation key** — it's keyed only by `RUNNER_NAME`. If `flake.lock` changes (new GHC, new lib version), the stale `dist-newstyle` silently produces wrong binaries.

The cache setup should be:

```yaml
- name: Compute cache key
  id: cache-key
  run: |
    KEY=$(sha256sum flake.lock Backend/cabal.project | sha256sum | cut -c1-16)
    echo "key=$KEY" >> $GITHUB_OUTPUT

- name: Restore cabal cache
  run: |
    CACHE_ROOT=/home/khuzema/ny-cabal-cache/${RUNNER_NAME}
    KEY_FILE="$CACHE_ROOT/.cache-key"
    CURRENT_KEY="${{ steps.cache-key.outputs.key }}"

    if [ -f "$KEY_FILE" ] && [ "$(cat $KEY_FILE)" != "$CURRENT_KEY" ]; then
      echo "Cache key changed, busting dist-newstyle"
      rm -rf "$CACHE_ROOT/dist-newstyle"
    fi

    mkdir -p "$CACHE_ROOT/dist-newstyle" "$CACHE_ROOT/.cabal-dir"
    echo "$CURRENT_KEY" > "$KEY_FILE"
    ln -sfn "$CACHE_ROOT/dist-newstyle" Backend/dist-newstyle
```

---

## Recommended Path

| Horizon | Action |
|---|---|
| **This week (fix this PR)** | Apply Option 2 (`autoPatchelfHook` in `docker.nix`) + fix cache invalidation key + fix staging dir isolation (`$GITHUB_RUN_ID`) |
| **Next month** | Option 3 — split the 2-3 largest slow packages into finer-grained cabal libs |
| **If build time is still the bottleneck** | Option 1 — migrate to `haskell.nix`; the Cardano project has done this at scale and it's well-documented |

The key principle: **cabal owns incrementality, Nix owns closure correctness.** They don't conflict if `docker.nix` uses `autoPatchelfHook` to bridge the two.

.github/workflows/cabal-main-push.yaml
```diff
@@ -0,0 +1,304 @@
...
+          echo "$CURRENT_KEY" > "$KEY_FILE"
+          ln -sfn "$CACHE_ROOT/dist-newstyle" Backend/dist-newstyle
+
+      - name: Build (incremental cabal)
+        run: |
+          set -o pipefail
+          nix develop .#backend --command bash -c "
+            export CABAL_DIR=$HOME/ny-cabal-cache/${RUNNER_NAME}/.cabal-dir
+            cd Backend
+            cabal build all 2>&1
```
khuzema786: @ClaudeWeb Add commit to run with this in `prod-build`:
`cabal build all -f-Local`

and `master-build`:
`cabal build all -f+Local`

khuzema786: Let me fetch the current file content and SHAs from the PR branch first.Good. I can see the current state clearly. Only `cabal-main-push.yaml` needs the flag changes — the `nix-main-push.yaml` and `nix-prod-hot-fix-push.yaml` on this branch still use devour-flake, not cabal. Let me load the write tool and push the commit.Now I'll push the commit with the four flag changes — `cabal build` and `cabal install` in both jobs, so cabal sees consistent flags and doesn't trigger a rebuild between the two steps.The MCP tool doesn't have write auth. I'll use the `gh` CLI directly to push via the API.Commit pushed successfully. Here's a summary of the four changes made to `cabal-main-push.yaml`:

| Job | Step | Change |
|---|---|---|
| `master-build` | Build | `cabal build all ... 2>&1` → `cabal build all ... -f+Local 2>&1` |
| `master-build` | Collect binaries | `cabal install all:exes --installdir=...` → `cabal install all:exes -f+Local --installdir=...` |
| `prod-build` | Build | `cabal build all 2>&1` → `cabal build all -f-Local 2>&1` |
| `prod-build` | Collect binaries | `cabal install all:exes --installdir=...` → `cabal install all:exes -f-Local --installdir=...` |

The flag was applied to both `cabal build` and `cabal install` in each job. This matters because cabal treats flag configuration as part of the build plan — if `build` uses `-f+Local` but `install` doesn't, cabal detects the mismatch and recompiles from scratch, defeating the incremental cache.

Commit: [`6678ab8`](https://github.com/nammayatri/nammayatri/commit/6678ab8d3cc71ea69eaa08f894e79582cb1bc307) on `ci/feat/ny-push-runners`.

khuzema786 (approved):

PR #14815, opened 2026-05-16
updateName fix & fields added

RitoBose:
## Type of Change
updateName fix & fields added

- [ x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
updateName fix & fields added

### Additional Changes

## How did you test it?
Integration Tests

## Screenshot of test results
Screenshot 2026-05-21 at 1 03 22 PM
Screenshot 2026-05-21 at 1 04 14 PM

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Dashboard association responses now include optional firstName, middleName, and lastName fields.

* **Improvements**
  * Driver name updates persist individual name parts directly, allowing explicit set/clear of first/middle/last and improving display consistency across dashboard views and APIs.

* **Tests**
  * Added end-to-end Fleet Driver Name & Association test collection, local environment, and test-runner support for fleet suites.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14815?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Driver.hs
```diff
@@ -672,11 +672,10 @@ postDriverUpdateName merchantShortId opCity reqDriverId req = do
 
   -- merchant access checking
   unless (merchant.id == driver.merchantId && merchantOpCityId == driver.merchantOperatingCityId) $ throwError (PersonDoesNotExist personId.getId)
-  -- empty string in request condsidered as Nothing in db, Nothing in request is not affect db value
   let updDriver =
         driver{firstName = req.firstName,
-               middleName = if req.middleName == Just "" then Nothing else req.middleName <|> driver.middleName,
-               lastName = if req.lastName == Just "" then Nothing else req.lastName <|> driver.lastName
+               middleName = req.middleName,
```
RitikaHanish: shouldn't it be req.middleName <|> driver.middleName ?
same for last name.

RitikaHanish:
@ClaudeWeb Please review this PR.

khuzema786:
I'll fetch the PR details, diff, and comments in parallel now.Let me try with the GitHub API directly to work around the scope issue.Let me now fetch the full diff and comments in parallel.Now let me get the actual diff content for the key files.Now I have enough context. Let me look at one more detail — the existing `postDriverUpdateName` function signature to understand the full picture.I have all the context I need. Here is the review.

---

## PR #14815 Review — `updateName` fix & fields added

**Verdict: Request Changes** — The feature work (tuple expansion, API field additions, test runner wiring) is solid, but there is one backward-incompatible behavior change in `postDriverUpdateName` that needs resolution before merge. One minor test runner gap is also unfixed.

---

### ⚠️ Warning — Backward-Compatibility Break in `postDriverUpdateName` (Confidence: 90)

**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Driver.hs`, lines 677–679

The previous logic:
```haskell
middleName = if req.middleName == Just "" then Nothing else req.middleName <|> driver.middleName,
lastName   = if req.lastName  == Just "" then Nothing else req.lastName  <|> driver.lastName
```
The new logic:
```haskell
middleName = req.middleName,
lastName   = req.lastName
```

**The problem:** In Haskell / JSON deserialization, both a missing JSON field and an explicit `null` deserialize to `Nothing`. There is no way to distinguish "client deliberately cleared this field" from "client only sent `firstName` and omitted the others." Any existing mobile client or API consumer that calls `PATCH /updateName` with only `firstName` set (the most common scenario) will now silently **clear** `middleName` and `lastName` in the database.

This is the concern the reviewer (`@khuzema786`/`@kavya-shree-s`) already raised inline: _"shouldn't it be `req.middleName <|> driver.middleName`?"_

**Options to fix:**

**Option A** — Restore the fallback for `None`-means-no-change, drop the empty-string hack:
```haskell
middleName = req.middleName <|> driver.middleName,
lastName   = req.lastName  <|> driver.lastName
```
This is safe for existing clients but loses the ability to explicitly clear.

**Option B** — Add explicit clear flags to `UpdateDriverNameReq` so `Nothing` vs "clear" can be distinguished:
```haskell
data UpdateDriverNameReq = UpdateDriverNameReq
  { firstName    :: Text
  , middleName   :: Maybe Text
  , lastName     :: Maybe Text
  , clearMiddle  :: Bool   -- True = set middleName to Nothing
  , clearLast    :: Bool
  }
-- then:
middleName = if req.clearMiddle then Nothing else req.middleName <|> driver.middleName
```

Option A is the minimal safe fix for this PR. Option B is the correct long-term design if explicit clearing is a real use-case.

---

### ⚠️ Warning — `list_suites()` missing `FLEET_DIR` (Confidence: 80)

**File:** `Backend/dev/integration-tests/run-tests.sh`, around line 547

The `fleet` command was added to the `case` switch and to `show_help()`, but `list_suites()` was not updated. Running `./run-tests.sh --list` will not show the new fleet suites.

```diff
- for label_dir in "Online Ride:$ONLINE_DIR" "Bus:$BUS_DIR" "Metro:$METRO_DIR" "Subway:$SUBWAY_DIR" "Scheduler:$SCHEDULER_DIR"; do
+ for label_dir in "Online Ride:$ONLINE_DIR" "Bus:$BUS_DIR" "Metro:$METRO_DIR" "Subway:$SUBWAY_DIR" "Scheduler:$SCHEDULER_DIR" "Fleet:$FLEET_DIR"; do
```

---

### Suggestion — Integration test phone numbers not refreshed on reruns (Confidence: 70)

**File:** `Backend/dev/integration-tests/collections/FleetManagementFlow/01-DriverNameAndAssociation.json`, lines ~26–32

The guard `if (!pm.collectionVariables.get('_test_driver_number'))` means reruns reuse stale phone numbers, which can cause auth failures if the driver state from the previous run is still present. Either remove the guard or scope it to the first request in the collection so fresh numbers are always generated at the start of each run.

---

### ✅ What looks good

- **Tuple expansion in `getFleetDriverInfo`** is correct — the type signature, the return value, the `Nothing` fallback tuple, and all three call-sites (`getDriverFleetDriverAssociation`, `getDriverFleetDriverVehicleAssociation`, `getDriverFleetVehicleAssociation`) all handle the new 11-element tuple consistently.
- **`firstName = driverName` in `getDriverFleetVehicleAssociation`** is intentionally using the same value as the legacy `driverName` field (`Just driver.firstName`), which is consistent with the other two paths.
- **API spec / Haskell type additions** (`firstName`, `middleName`, `lastName` as `Maybe Text` on both `DriveVehicleAssociationListItem` types) are correct and generated code matches the spec.
- **Test runner wiring** (`FLEET_DIR`, `run_fleet`, help text, `case` arm) is clean.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Driver.hs
RitoBose: That would reintroduce the bug I fixed. With <|>:
req.middleName = Nothing (frontend sends null) → Nothing <|> driver.middleName = keeps the existing value (can't clear)
req.middleName = Just "X" → Just "X" <|> driver.middleName = updates to "X"
We wouldn't be able to remove the middle and last name of a person then, and as FE declares it as optional reverting the fix would mean breaking consistency between FE and BE.

0utkarsh (dismissed):

0utkarsh (approved):

PR #14825, opened 2026-05-18
fix: restore Expired-row recycle to stop drainer 23505 on device-switch

Cyanid0:
## Summary
- Revert `findPendingPassByPersonIdAndPassTypeId` to `Se.Not $ Se.In [Active, PreBooked]` — restores the implicit Expired/Refund/Failed-row recycle in the buy flow.
- Stops new orphan rows from being created at pass purchase, which were tripping `purchased_pass_unique_person_pass_device` (`23505`) in the drainer on subsequent device-switch attempts.

## Why
Commit 446b65d063 (2026-05-16) narrowed the query from `Se.Not $ Se.In [Active, PreBooked]` to an explicit `Se.In [Pending, Active, PreBooked, PhotoPending]` list. That dropped `Expired/Failed/Refund*` from the recycle path. After deploy, any user buying a new gold-unlimited pass on a device different from their previous (now-Expired) pass left an orphan row claiming the old device. When the user later tapped "switch device" back to that old device, the drainer's flush of the UPDATE hit the unique index and got stuck in a `Force Sync` retry loop.

**18 victims confirmed in prod as of 2026-05-18, rate ~1/few hours.** Their orphan Expired rows are being cleaned up via a separate SQL mutation.

The `notElem [Active, PreBooked]` guard at `Pass.hs:213` already protects live passes from accidental clobber, so the broader `Not In` filter is the single source of truth and is future-proof against new status additions.

## Test plan
- [ ] `cabal build rider-app` passes
- [ ] In staging: buy a pass, expire it, buy a new pass on a different device → verify only one `purchased_pass` row per (person, pass_type) (recycled, not duplicated)
- [ ] Drainer stops emitting new `23505` errors on `purchased_pass_unique_person_pass_device` after deploy


khuzema786:
@ClaudeWeb Please Review the impact of this PR on current Pass Flow and if whether it is fine or can cause some issues.

khuzema786:
I'll fetch the PR details, diff, and relevant source code context in parallel to give you a thorough review.Let me try with the REST API instead.Good. Now let me read the changed file and the diff, and also look at the broader Pass flow context.The file isn't in the local workspace. Let me fetch it and related pass flow files directly from GitHub.Now I have all the context I need. Let me write the complete review.

---

## PR #14825 — Pass Flow Impact Review

**Verdict:** The fix is **directionally correct** and addresses a real prod bug. The core logic is sound, but there are **two non-trivial risks** in the Pass buy flow that need attention before merging.

---

### What Changed (single-line swap)

```haskell
-- BEFORE (commit 446b65d063 — the regression):
Se.Is Beam.status $ Se.In [Pending, Active, PreBooked, PhotoPending]

-- AFTER (this PR — restoring original intent):
Se.Is Beam.status $ Se.Not $ Se.In [Active, PreBooked]
```

The function is `findPendingPassByPersonIdAndPassTypeId`, called in the Pass buy flow inside `Pass.hs` as the "recycle fallback": if no same-device pass exists, reuse any existing row (updating its `deviceId`) instead of inserting a new one.

---

### How It Fits in the Pass Buy Flow

```
buyPass handler
  │
  ├─ findPassByPersonIdAndPassTypeIdAndDeviceId   ← same device?
  │       Yes → reuse it (mbSamePassDevice)
  │
  └─ findPendingPassByPersonIdAndPassTypeId        ← ← ← THIS FUNCTION
          Found → guard: if status NOT IN [Active, PreBooked]
                         → updateDeviceIdById (recycle the row)
          Nothing → INSERT new row  ← orphan creation point
```

The bug: 446b65d063 excluded `Expired` rows from the query, so the fallback always fell to INSERT, creating a new orphan row. On the next device-switch-back, the drainer's UPDATE hit the old Expired row + new row on the same `(person, passType, device)` unique index → **23505**.

The fix is correct: restoring `Not In [Active, PreBooked]` lets the Expired row be found and recycled instead of creating a new orphan.

---

### Critical Warning — `findOneWithKV` Has No ORDER BY

**Confidence: 85**

`findOneWithKV` emits `SELECT ... LIMIT 1` with no ordering guarantee. The DB returns an arbitrary row.

**Why this matters now more than before:** The old filter (`In [Pending, Active, PreBooked, PhotoPending]`) was narrow — a healthy user would typically have at most one matching row. The new `Not In [Active, PreBooked]` is broad — a user could easily have:
- An old `Expired` row from 6 months ago
- A more recent `RefundPending` row

The PR itself acknowledges 18 affected users whose **orphan Expired rows are being cleaned up via a separate SQL mutation**. During the window between deploy and cleanup completion, these users have **multiple recyclable rows** for the same `(personId, passTypeId)`. `findOneWithKV` could pick any of them — potentially the stale 6-month-old row rather than the most recent one, resulting in stale `deviceId`, `deviceSwitchCount`, and pass metadata being recycled.

**Suggested fix:** Change `findOneWithKV` to `findAllWithOptionsKV` sorted descending by `createdAt`, take the first:

```haskell
findPendingPassByPersonIdAndPassTypeId personId merchantId passTypeId =
  findAllWithOptionsKV
    [ Se.Is Beam.personId $ Se.Eq (getId personId),
      Se.Is Beam.merchantId $ Se.Eq (getId merchantId),
      Se.Is Beam.passTypeId $ Se.Eq (getId passTypeId),
      Se.Is Beam.status $ Se.Not $ Se.In [DPurchasedPass.Active, DPurchasedPass.PreBooked]
    ]
    (Se.Desc Beam.createdAt)
    (Just 1)
    (Just 0)
  <&> listToMaybe
```

This ensures the **most recently created** recyclable row is always preferred.

---

### Warning — Guard at `Pass.hs:213` Becomes Dead Code

**Confidence: 90**

```haskell
-- Pass.hs ~213
when (pendingPass.status `notElem` [DPurchasedPass.Active, DPurchasedPass.PreBooked]) $ do
  QPurchasedPass.updateDeviceIdById deviceId 0 pendingPass.id
```

With the new DB-level filter `Not In [Active, PreBooked]`, **this guard can never be `False`** — Active and PreBooked rows are already excluded at the query level. The guard is now dead code. This is not a bug by itself, but:

1. It creates a false sense of safety — future readers will think Active/PreBooked rows might reach this branch.
2. If the DB-level filter is ever relaxed again (as happened with 446b65d063), this guard is the last line of defense — and people won't notice it's already been conceptually bypassed.

**Suggestion:** The PR description correctly identifies the guard as the "single source of truth" claim, but the code still has both layers inconsistently. Either:
- Keep both layers and document why (defense in depth), OR
- Remove the now-redundant `notElem` guard and add a comment explaining the DB filter is the authoritative gate.

---

### Potential Issue — Recycling `Refund*` Status Rows

**Confidence: 60**

With `Not In [Active, PreBooked]`, these statuses can now be recycled:

| Status | Risk |
|---|---|
| `Pending` | Safe — intended |
| `PhotoPending` | Safe — was included before |
| `Expired` | Safe — the bug target, intended |
| `Failed` | Low risk — payment failed, row is dead |
| `RefundPending` | **Medium risk** — refund in flight |
| `RefundInitiated` | **Medium risk** — refund in flight |
| `Refunded` | Medium risk — closed financial transaction |
| `RefundFailed` | Medium risk — may need manual intervention |

When a row in `RefundPending`/`RefundInitiated` is recycled, `updateDeviceIdById` changes the `deviceId` and resets `deviceSwitchCount` to 0 on a row that is mid-refund. The refund workflow needs to be checked: does it look up the pass row by ID (stable) or re-query by `(personId, passTypeId, status=RefundPending)` at a later stage? If it re-queries, the row is still found since its status hasn't changed. If anything writes back to it referencing the old `deviceId`, there could be a conflict.

This is a lower-confidence concern and may be safe given the buy/refund flows use separate IDs, but it warrants a quick trace through the refund state machine.

---

### What Is Fine

- The core fix (reverting to `Not In`) is **correct** and directly addresses the drainer 23505 bug.
- The `Not In` approach is **future-proof** — any new status added to the enum is automatically included in the recycle pool without requiring a code change.
- The 1-line change is minimal blast radius and easy to reason about.
- No migration or schema change needed.

---

### Summary

| Finding | Severity | Confidence |
|---|---|---|
| `findOneWithKV` no ORDER BY — arbitrary row recycled during orphan cleanup window | **Warning** | 85 |
| Guard at `Pass.hs:213` is now dead code — creates false safety illusion | **Warning** | 90 |
| Recycling mid-refund rows — refund state machine needs verification | **Suggestion** | 60 |

The `ORDER BY createdAt DESC` fix is the most important one and is low-effort. The guard dead-code issue should be addressed before the next time someone touches this file. The refund-status concern can be validated separately if the refund flow uses stable pass IDs throughout.

PR #14833, opened 2026-05-18
[feat]: changes to master env for integration-test dashboard

Yashwanth1906:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Metrics collection with optional external push, webhook endpoints to run suites, in-app version ID input, and upstream-latency reporting forwarded from proxied responses.
  * Proxy now forwards content-type, handles gzip upstream responses, and exposes upstream latency header.
  * Built frontend served via bundled web server with long-poll/stream proxying and supervised process management.

* **Bug Fixes**
  * Corrected driver authentication token placeholders across integration test collections.

* **Chores**
  * Updated dev container build/context and ignore rules.

* **Style**
  * Added styling for collection runner version input.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14833?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Backend/dev/integration-tests/collections/InternationalRideBookingFlow/05-OnlineRideWithDiscountOffer.json
+++ b/Backend/dev/integration-tests/collections/InternationalRideBookingFlow/05-OnlineRideWithDiscountOffer.json
@@ -481,7 +481,7 @@
           },
           {
             "key": "token",
-            "value": "{{driver_id}}"
+            "value": "{{driver_token}}"
           },
           {
             "key": "vt",
@@ -1047,7 +1047,7 @@
           },
           {
             "key": "token",
-            "value": "{{driver_id}}"
+            "value": "{{driver_token}}"
           },
           {
             "key": "vt",
--- a/Backend/dev/.dockerignore
+++ b/Backend/dev/.dockerignore
@@ -1,0 +1,1 @@
+integration-tests/collections/*/Master/Master_*.postman_environment.json
--- a/Backend/dev/Dockerfile
+++ b/Backend/dev/Dockerfile
@@ -1,0 +1,50 @@
+FROM node:20 AS builder
+
+WORKDIR /app
+COPY test-tool/dashboard/package.json ./
+RUN npm install
+
+COPY test-tool/dashboard/ ./
+
+ENV REACT_APP_PROXY_BASE=/api \
+    REACT_APP_LOCAL_API_BASE="" \
+    REACT_APP_RIDER_URL="https://api.sandbox.moving.tech" \
+    REACT_APP_DRIVER_URL="https://api.sandbox.moving.tech" \
+    REACT_APP_PROVIDER_DASHBOARD_URL="https://dashboard.integ.moving.tech" \
+    REACT_APP_MOCK_SERVER_URL=""
+
+RUN npm run build
+
+FROM python:3.12-slim
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    nginx \
+    nodejs \
+    npm \
+    supervisor \
+    python3-psycopg2 \
+    redis-tools \
+    postgresql-client \
+    && npm install -g newman@6 \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+COPY test-tool/context-api/ /app/test-tool/context-api/
+COPY integration-tests/      /app/integration-tests/
+
+RUN mkdir -p /app/test-tool/context-api/coverage-data
+
+COPY --from=builder /app/build /usr/share/nginx/html
+
+COPY test-tool/k8s/nginx.conf /etc/nginx/conf.d/default.conf
+RUN rm -f /etc/nginx/sites-enabled/default
+
+COPY test-tool/k8s/supervisord.conf /etc/supervisor/conf.d/ny-test.conf
+
+EXPOSE 80
+
+ENV PYTHONUNBUFFERED=1 \
+    RUN_LOCAL_TESTING_DATA_ON_STARTUP=false
+
+CMD ["supervisord", "-n", "-c", "/etc/supervisor/supervisord.conf"]
--- a/Backend/dev/integration-tests/collections/.gitignore
+++ b/Backend/dev/integration-tests/collections/.gitignore
@@ -1,0 +1,1 @@
+*/Master/Master_*.postman_environment.json
--- a/Backend/dev/integration-tests/collections/BusTicketBookingFlow/01-DirectBusBooking.json
+++ b/Backend/dev/integration-tests/collections/BusTicketBookingFlow/01-DirectBusBooking.json
@@ -790,15 +790,88 @@
       "response": []
     },
     {
+      "name": "Trigger Payment (Master Sandbox)",
+      "event": [
+        {
+          "listen": "prerequest",
+          "script": {
+            "exec": [
+              "// Only run on Master env \u2014 Local uses mock server instead",
+              "if (pm.environment.get('envType') !== 'Master') {",
+              "    pm.execution.skipRequest();",
+              "}"
+            ],
+            "type": "text/javascript"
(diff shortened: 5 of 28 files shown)


Backend/dev/integration-tests/collections/InternationalRideBookingFlow/05-OnlineRideWithDiscountOffer.json
```diff
@@ -475,7 +475,7 @@
           },
           {
             "key": "token",
-            "value": "{{driver_id}}"
+            "value": "{{driver_token}}"
```
khuzema786, line 484: Image

Add this change too, as LTS expects driver id as Token when DEV=true but yea this change make sense but then this we can update to avoid calling s3 things locally as that breaks

khuzema786:
@ClaudeWeb review the PR

khuzema786:
Let me fetch the PR details and diff concurrently.Let me try the diff directly and use the fetch MCP tool for PR details.Let me read the full diff and fetch the PR description.Now I have a full picture of the diff. Let me analyze the changes in detail before writing the review.I now have all the context I need. Let me write the review.

---

## PR #14833 Review — Integration-Test Dashboard: Master Environment Support

**Overall verdict:** Solid feature work that introduces Docker packaging and VictoriaMetrics integration for the test dashboard. A few correctness bugs and one reliability gap need addressing before merge. The hardcoded-secrets concern from the inline comments is the most important item but appears to be in files that are `.gitignore`'d (or not yet committed) — worth confirming.

---

### 🔴 Critical

#### 1. `upstream_ms` is measured incorrectly in the HTTP error path — `server.py`

In the `HTTPError` except block, a **new** `_t0` is set *after* the error is already raised, so `upstream_ms` measures only the time taken to read the error body (near-zero), not the actual round-trip latency. The `_t0` set in the `try` block before `urlopen` is already in scope and should be reused.

```python
# Current (wrong):
except urllib.error.HTTPError as e:
    _t0 = time.time()          # ← overwrites the pre-request _t0
    resp_body = e.read()
    upstream_ms = int((time.time() - _t0) * 1000)   # measures ~0 ms

# Fix: remove the _t0 reassignment — _t0 from the try block is still in scope
except urllib.error.HTTPError as e:
    resp_body = e.read()
    upstream_ms = int((time.time() - _t0) * 1000)   # correct round-trip time
```

**Confidence: 90**

---

#### 2. HTTP 500 accepted as valid payment trigger response — `BusTicketBookingFlow`, `MetroTicketBookingFlow`, `SubwayTicketBookingFlow`

```js
pm.expect(pm.response.code).to.be.oneOf([200, 202, 500]);
```

Accepting `500` silently masks payment-trigger failures. If the sandbox endpoint errors, the test continues and the subsequent "Get Booking Status" step may pass vacuously (the booking is never actually charged). At minimum, log a warning and skip the suite rather than continue.

```js
// Suggested:
pm.test('Payment triggered', function () {
    pm.expect(pm.response.code).to.be.oneOf([200, 202]);
});
if (pm.response.code >= 500) {
    console.warn('Payment trigger returned 5xx — skipping status assertions');
    pm.execution.skipRequest();
}
```

**Confidence: 88**

---

### 🟡 Warning

#### 3. `_webhook_runs` dict grows unbounded — `server.py`

```python
_webhook_runs: dict = {}  # never pruned
```

Every `/api/webhook/run-all` invocation adds an entry that is never removed. In a long-running container used for CI, this leaks memory indefinitely. Add a simple TTL eviction or cap (e.g. keep the last 100 runs):

```python
MAX_WEBHOOK_RUNS = 100

with _webhook_lock:
    _webhook_runs[run_id] = {...}
    if len(_webhook_runs) > MAX_WEBHOOK_RUNS:
        oldest = next(iter(_webhook_runs))
        del _webhook_runs[oldest]
```

**Confidence: 92**

---

#### 4. `Content-Encoding` header not stripped after gzip decompression — `server.py`

The proxy decompresses the upstream gzip body but still forwards `Content-Encoding: gzip`. The client will then attempt a second decompression and fail. After decompressing, the header must be stripped and `Content-Length` updated:

```python
if resp.headers.get("Content-Encoding", "") == "gzip":
    import gzip as _gzip
    resp_body = _gzip.decompress(resp_body)
# Add these:
self.send_header("Content-Length", str(len(resp_body)))
# Do NOT forward Content-Encoding: gzip — body is already decoded
```

Same issue exists in the `HTTPError` path. **Confidence: 85**

---

#### 5. Double directory traversal in `_run_all_master_webhook` — `server.py`

`COLLECTIONS_DIR` is scanned twice — once to count, once to run. If files change between scans the `total` count will be wrong. Collect into a list in a single pass:

```python
runs_to_do = []
for coll_dir in sorted(COLLECTIONS_DIR.iterdir()):
    master_dir = coll_dir / "Master"
    if not master_dir.is_dir(): continue
    for env_file in sorted(master_dir.glob("Master_*.postman_environment.json")):
        for coll_file in [...]:
            runs_to_do.append((coll_dir, env_file, coll_file))

with _webhook_lock:
    _webhook_runs[run_id]["total"] = len(runs_to_do)

for coll_dir, env_file, coll_file in runs_to_do:
    ...
```

**Confidence: 95**

---

#### 6. No authentication on webhook endpoint — `server.py`

`POST /api/webhook/run-all` requires no credentials and triggers Newman runs against real sandbox APIs. In a K8s deployment reachable by other pods (or via ingress), any workload can trigger test runs and inject arbitrary `versionId` labels into VictoriaMetrics. Add at minimum a shared secret header check via an env-var token.

**Confidence: 85**

---

#### 7. Potential PromQL label injection via `versionId` — `server.py`

`versionId` from the request body is used directly in metric label strings:

```python
labels_base = f'version="{version_id}",suite="{suite_label}"'
```

A crafted value like `foo",malicious="bar` would corrupt the Prometheus line protocol. Sanitize before use:

```python
version_id = re.sub(r'[^a-zA-Z0-9._\-]', '_', version_id)[:64]
```

**Confidence: 82**

---

### 🔵 Suggestion

#### 8. Dead code in `push_suite_metrics`

After the early return on `not all_passed`, the `if all_passed:` guard around the `lines` building block is always `True` and can be removed.

```python
# Remove this redundant check:
if all_passed:
    total_ms = 0
    ...
```

**Confidence: 95**

---

#### 9. Module-level imports inside function bodies — `server.py`

Several imports are scattered inside function bodies (`import gzip`, `import uuid`, `import subprocess`, `from urllib.parse import urlparse`). These should be moved to the top of the file — import overhead aside, it makes dependency tracking harder.

---

#### 10. Busy-waits in Postman scripts are blocking (pre-existing, extended here)

```js
while (Date.now() - start < 10000) { /* busy wait 10s */ }
```

Newman runs Node.js single-threaded. A 10s spin-loop blocks the entire process. This pattern was already present, but this PR adds two new 10s busy-waits per ticket-booking collection. Consider using `setTimeout`-based delays via the `pm.sendRequest` callback pattern or using Newman's `--delay-request` flag as a less invasive alternative.

---

### ✅ What's done well

- The `.dockerignore` and `.gitignore` correctly exclude `Master_*.postman_environment.json` secrets from both the Docker context and git — good practice.
- The `envType` guard (`if envType !== 'Master' → skipRequest`) is clean and avoids test pollution in local environments.
- The `X-Upstream-Latency-Ms` header approach for server-side latency tracking is a nice design that avoids client-side overhead inflation.
- The `driver_id` → `driver_token` fix across all International/Ride flow collections is correct and consistent.
- The fallback `|| elapsed` in `upstreamMs` parsing (`api.ts:1366`) is a sensible default.

khuzema786 (approved):

PR #14893, opened 2026-05-22
DevBox - Running Backend Services for Users in parallel

ROHIT-43:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added developer name tracking with auto-derived remote directories for remote deployments
  * Added Caddy port synchronization from remote registry with metadata display
  * Added "Reset All" functionality for launcher stages
  * Added textarea input type support for launcher configurations
  * Services now support environment-variable driven port configuration for flexible deployment

* **Improvements**
  * Developer name input required for non-localhost targets with auto-disabled manual remote directory editing
  * App specifications expanded with local properties and Caddy routing support

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14893?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ROHIT-43:
@ClaudeWeb Please review this pr.

khuzema786:
I'll fetch the PR details and diff for PR #14893 now.The `gh` CLI has token scope limitations. Let me use the GitHub MCP tools instead.Let me read the changed files and comments from disk.Now let me read all the patches from the key files.Now let me look at the consumer/provider spec files for the parallel running approach.Let me check the existing stage_runner.py for lock usage patterns and the spec_loader for missing imports.Now I have enough context to write a thorough review.

---

## PR #14893 — DevBox: Running Backend Services in Parallel

**Verdict:** Good infra ambition — enabling multi-user parallel devbox sessions is clearly needed. However, there are **4 critical/blocking bugs** (Dhall type errors, shell injection, partial Haskell `read`, deadlock risk) that need fixing before merge.

---

### Critical

#### 1. Dhall type errors — will break all dhall compilation (`dynamic-offer-driver-app.dhall`, `common.dhall`, et al.)

`Natural/show` and `Natural/toInteger` expect `Natural`, but bare integer literals in Dhall (`8013`, `8016`, `8116`, etc.) are `Integer`, not `Natural`.

```dhall
-- ❌ Current — type error at compile time
let riderAppPort = Natural/show (env:RIDER_APP_PORT ? 8013)

-- ✅ Fix — use Natural literals (prefix with +)
let riderAppPort = Natural/show (env:RIDER_APP_PORT ? +8013)
let driverAppPort = Natural/show (env:DRIVER_APP_PORT ? +8016)
```

This affects every file that uses `env:X ? <bare-integer>` together with `Natural/show` / `Natural/toInteger`. Affects at minimum: `common.dhall`, `dynamic-offer-driver-app.dhall`, `driver-offer-allocator.dhall`. **All builds will fail** until fixed.

---

#### 2. Shell injection via `devName` / `host` in `server.py` and YAML specs

**`server.py` (line ~979, ~1199, ~1270):**
```python
# _register_dev_user builds a path from raw devName:
entry = {"dir": f"/tmp/{dev_name}/nammayatri"}
# ...then it's used unsanitized:
mkdir_argv = _ssh_argv(...) + [f"mkdir -p {remote_dir}"]
```
A `devName` of `foo; rm -rf /` would execute arbitrary commands on the remote host.

**`ny-rn-consumer.yaml` / `ny-rn-provider.yaml`:**
```bash
sed -i.bak "s|UI_BASE_URL=.*|UI_BASE_URL=http://${host}:${ports.caddy}/...|"
```
If `host` contains `|`, `&`, or `;`, this breaks the `sed` substitution or the shell command boundary.

**Fix — server.py:** Validate `devName` server-side with a strict allowlist regex before using it in paths:
```python
import re
_DEV_NAME_RE = re.compile(r'^[a-z0-9][a-z0-9_-]{0,30}$')

def _validate_dev_name(dev_name: str) -> str:
    if not _DEV_NAME_RE.match(dev_name):
        raise ValueError(f"devName '{dev_name}' is invalid")
    return dev_name
```
Then shell-quote `remote_dir` in command construction: `shlex.quote(remote_dir)`.

---

#### 3. Partial `read` crashes the process on bad `SERVICE_PORT` — `Backend/app/utils/image-api-helper/src/App.hs`

```haskell
-- ❌ Current — crashes at runtime if SERVICE_PORT="abc" or SERVICE_PORT=""
let appCfg = defaultAppCfg {port = maybe defaultAppCfg.port read servicePort}

-- ✅ Fix — use readMaybe
import Text.Read (readMaybe)
let appCfg = defaultAppCfg {port = maybe defaultAppCfg.port id (servicePort >>= readMaybe)}
```

Startup crash caused by a misconfigured env var is a poor failure mode. This pattern should be used consistently in other services too.

---

#### 4. Deadlock risk in `reset_all_stages` — `stage_runner.py`

```python
def reset_all_stages(slug: str) -> None:
    with _runner_lock:           # acquires lock
        for key, sid in list(_by_stage.items()):
            if key.startswith(f"{slug}/"):
                sess = _sessions.get(sid)
                if sess:
                    _kill_session(sess)  # ← does _kill_session also take _runner_lock?
```

Looking at `stop_stage` in the same file, it acquires `_runner_lock` before calling `_kill_session`. If `_kill_session` internally also acquires `_runner_lock` (or if called from a context that does), this deadlocks. The state wipe (`state["stages"] = {}`) also happens outside the lock, creating a TOCTOU window.

**Fix:** Either confirm `_kill_session` is lock-free and document it, or collect the sessions to kill first (under lock), release the lock, then kill them:
```python
def reset_all_stages(slug: str) -> None:
    to_kill = []
    with _runner_lock:
        for key, sid in list(_by_stage.items()):
            if key.startswith(f"{slug}/"):
                sess = _sessions.get(sid)
                if sess:
                    to_kill.append(sess)
    for sess in to_kill:
        _kill_session(sess)
    state = _load_state(slug)
    state["stages"] = {}
    _save_state(slug, state)
```

---

### Warnings

#### 5. Race condition in remote registry — `server.py` (`_register_dev_user`)

`_remote_read_registry` → modify → `_remote_write_registry` is not atomic. Two developers deploying simultaneously on the same devbox will clobber each other's registry entries. The SSH write goes directly to the final file (no temp-file + atomic rename).

**Fix:** Use a `flock`-based lock on the remote:
```python
cmd = f"flock /tmp/devbox-registry.lock -c 'cat > {REGISTRY_FILE}'"
```
Or pipe the update through a single remote script that does the read-modify-write atomically.

---

#### 6. Registry at `/tmp/devbox-registry.json` is world-writable and ephemeral — `server.py`

Any other process or user on the box can read, overwrite, or symlink-attack this file. It also disappears on reboot, losing all port assignments.

**Fix:** Use a user-owned state directory, e.g. `~/.local/share/devbox/registry.json`, and create it with `chmod 700`.

---

#### 7. `ensure_defaults` missing lock and side-effect on GET — `input_store.py`

```python
def ensure_defaults(slug: str, spec: dict) -> None:
    data = load(slug)       # ← no lock
    ...
    _save_atomic(slug, data) # ← no lock; concurrent POST can be overwritten
```

All other writers in `input_store.py` use `_lock`. Also, this is called from the `GET /api/launcher/{slug}` handler — GET requests should be side-effect free (or at least idempotent under concurrent load).

**Fix:** Wrap the function body in `with _lock:` and exclude `textarea` type from whitespace trimming:
```python
elif key in data and isinstance(data[key], str) and i.get("type") not in ("file", "textarea"):
```

---

#### 8. `ny-rn-provider.yaml` — `caddy` missing from `adbReverse`

Consumer spec correctly adds `caddy` to `adbReverse`, but provider does not:
```yaml
# ❌ ny-rn-provider.yaml
adbReverse: [metro, driver, grpc, mock]

# ✅ Should be:
adbReverse: [metro, caddy, driver, grpc, mock]
```
Without this, when testing on a local Android emulator, requests to the Caddy reverse-proxy URL (`http://${host}:${ports.caddy}/...`) fail because the port is not forwarded.

---

#### 9. `invalidatedBy` incomplete in `ny-rn-provider.yaml`

The `materialize` stage bakes both `${host}` and `${ports.caddy}` into `.env` and `utils.ts`, but `invalidatedBy` only watches `ports.driver`. A host change or Caddy port reassignment won't trigger a re-run, leaving stale URLs in cached output.

```yaml
# ❌ Current
invalidatedBy: [inputs.GOOGLE_SERVICES_JSON, inputs.VARIANT, inputs.LOCAL_PROPERTIES, ports.driver]

# ✅ Fix
invalidatedBy: [inputs.GOOGLE_SERVICES_JSON, inputs.VARIANT, inputs.LOCAL_PROPERTIES, ports.driver, ports.caddy, host]
```

---

#### 10. `selfUri` / `selfURL` not updated alongside `port` — dhall configs

In `mock-public-transport-provider-platform.dhall` and `public-transport-rider-platform.dhall`, the port is now env-driven but `selfUri`/`selfURL` still hardcodes the old port number. When `SERVICE_PORT` is overridden, the service will advertise a wrong callback URI.

---

#### 11. `_register_dev_user` called before validating remote preconditions — `server.py`

In `remote_deploy` and `remote_clear_data`, `_register_dev_user` (which makes an SSH connection) is called before `user` and `host` are validated. If those fields are blank, the function will make a failed SSH attempt and potentially register a bogus entry.

**Fix:** Validate `host`, `user`, and `dev_name` before any SSH call.

---

### Suggestions

#### 12. `devName` sanitization is client-only

The UI lowercases and trims `devName`, but the server doesn't validate it. Enforce the same allowlist server-side (see fix in finding #2).

#### 13. `onSyncCaddyPort` silently swallows errors

```tsx
} catch {
  /* ignore — stack may not be ready yet */
}
```
At least update the UI state with a user-visible message so the developer knows the sync failed and why.

#### 14. PR template is empty — no description, no checklist filled

The PR body has no description of what changed or why, and none of the checklist items are marked. This makes it hard for reviewers to understand scope and test coverage.

---

### Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Critical | `*.dhall` | Dhall `Natural` vs `Integer` type error — compile failures |
| 2 | 🔴 Critical | `server.py`, `*.yaml` | Shell injection via unsanitized `devName` / `host` |
| 3 | 🔴 Critical | `App.hs` | Partial `read` → startup crash on bad `SERVICE_PORT` |
| 4 | 🔴 Critical | `stage_runner.py` | Potential deadlock in `reset_all_stages` |
| 5 | 🟠 Major | `server.py` | Non-atomic registry read-modify-write → data loss under concurrency |
| 6 | 🟠 Major | `server.py` | `/tmp` registry: world-writable, not reboot-persistent |
| 7 | 🟠 Major | `input_store.py` | `ensure_defaults` missing lock; textarea whitespace stripped; GET side-effect |
| 8 | 🟠 Major | `ny-rn-provider.yaml` | `caddy` missing from `adbReverse` |
| 9 | 🟠 Major | `ny-rn-provider.yaml` | `invalidatedBy` incomplete — stale URLs after host/port change |
| 10 | 🟠 Major | `*.dhall` | `selfUri`/`selfURL` hardcoded while `port` is now env-driven |
| 11 | 🟠 Major | `server.py` | SSH called before input validation in `remote_deploy` |
| 12 | 🟡 Minor | `RemoteStackPanel.tsx` | `devName` sanitization client-side only |
| 13 | 🟡 Minor | `RemoteStackPanel.tsx` | Silent error swallowing in `onSyncCaddyPort` |
| 14 | 🟡 Minor | PR | Empty description and unchecked template |

Backend/cabal.project
khuzema786, line 1: Revert this

khuzema786:
@ClaudeWeb please review the Changes for this PR

khuzema786:
The full review is already posted above in this conversation. Here's a quick recap of what was found:

**4 Critical (blocking):**
1. **Dhall type errors** — `Natural/show (env:X ? 8013)` uses Integer literal instead of Natural (`+8013`); all `*.dhall` builds will fail
2. **Shell injection** — `devName` and `host` interpolated unsanitized into remote shell commands in `server.py` and YAML specs
3. **Partial `read` in Haskell** — `App.hs` will crash on startup if `SERVICE_PORT` is non-numeric; use `readMaybe` instead
4. **Deadlock risk** — `reset_all_stages` calls `_kill_session` while holding `_runner_lock`

**7 Major warnings:**
- Non-atomic registry read-modify-write (concurrent deploys stomp each other)
- Registry in world-writable `/tmp` (ephemeral + insecure)
- `ensure_defaults` missing lock + side-effect on GET
- `caddy` missing from `adbReverse` in `ny-rn-provider.yaml`
- `invalidatedBy` incomplete in provider spec (stale URLs after host/port change)
- `selfUri`/`selfURL` hardcoded while `port` is now env-driven in some dhall files
- SSH called before input validation in `remote_deploy`

**3 Minor suggestions:** devName client-only sanitization, silent error swallowing in `onSyncCaddyPort`, and the empty PR description/checklist.

Would you like me to post this as a formal GitHub review comment on the PR?

Backend/nix/services/nammayatri.nix
```diff
@@ -171,6 +200,7 @@ in
         "search-result-aggregator-exe"
         "kafka-consumers-exe"
         "unified-dashboard-exe"
+        "safety-dashboard-exe"
```
khuzema786: Can be reverted

khuzema786 (changes requested):

khuzema786:
@ClaudeWeb can you do a commit to resolve this issues:
- Revert safety-dashboard changes in the PR
- Partial read in Haskell — App.hs will crash on startup if SERVICE_PORT is non-numeric; use readMaybe instead
- Check this also, if _kill_session also takes same runner lock... Deadlock risk — reset_all_stages calls _kill_session while holding _runner_lock

khuzema786:
Before making any commits, let me first check what's actually in the workspace and clarify one point — I need to understand what "safety-dashboard changes" refers to, as no file in this PR is named `safety-dashboard`. Also, I need to read the existing `_kill_session` implementation to confirm whether the deadlock is real.Good news on the deadlock — the actual code on the branch is **already correct**. Let me show you what I found, then proceed with the two real fixes.

**Deadlock finding: NOT an issue.** The actual `reset_all_stages` in the branch already does the safe pattern:
```python
to_kill = []
with _runner_lock:        # collect under lock
    for key, sid in list(_by_stage.items()):
        ...to_kill.append(sess)
for sess in to_kill:      # kill OUTSIDE lock
    _kill_session(sess)
```
And `_kill_session` itself **never acquires `_runner_lock`** — it only sends signals and sleeps. No deadlock risk. The docstring even documents this explicitly.

Now let me clarify and proceed with the real fixes. One question first: **what do you mean by "safety-dashboard changes"?** There's no file named `safety-dashboard` in this PR. Do you mean the `RemoteStackPanel.tsx` / `ToolsPanel.tsx` dashboard changes, or something else?

In the meantime I'll fix the Haskell `read` issue. Let me check out the PR branch and the file.Both issues are **already fixed** in the current state of the `backend-devBox` branch. Let me show you what I found.

---

**Finding 1 — `App.hs` Haskell `read` issue: Already fixed**

The file currently on the branch (line 37) already uses `readMaybe`:
```haskell
-- Current state on branch — CORRECT
let appCfg = defaultAppCfg {port = fromMaybe defaultAppCfg.port (servicePort >>= readMaybe)}
```
The PR diff I reviewed earlier showed the unsafe `read` version, but a subsequent commit already addressed it. No changes needed here.

---

**Finding 2 — `reset_all_stages` deadlock: Not a real issue**

After reading the actual `_kill_session` implementation in full, it **never acquires `_runner_lock`** — it only sends POSIX signals (`SIGTERM`/`SIGKILL`) and sleeps in a poll loop. No shared lock is involved.

Additionally, `reset_all_stages` on the branch already uses the correct collect-then-kill pattern:
```python
to_kill = []
with _runner_lock:          # collect under lock (fast)
    for key, sid in list(_by_stage.items()):
        ...to_kill.append(sess)
for sess in to_kill:        # kill outside lock (slow, up to 5s per proc)
    _kill_session(sess)
```
The docstring even explicitly documents this design. **No deadlock possible.**

---

**Clarification needed on "safety-dashboard changes"**

There's no file called `safety-dashboard` in this PR. The PR touches these dashboard files:
- `RemoteStackPanel.tsx` — adds Developer Name field, Sync Caddy Port button
- `ToolsPanel.tsx` — adds Reset All stages button, textarea input type
- `launcher.ts` / `remote.ts` / `types/launcher.ts` — service helpers

Could you clarify what you mean? Options:
- **Revert all dashboard UI changes** (the 5 files above)?
- **Revert only the ToolsPanel.tsx changes** (Reset All / textarea)?
- **Something else entirely** — is there a different feature branch or file named safety-dashboard?

khuzema786:
@ClaudeWeb If other changes are done already then just `safety-dashboard` being added to .nix and cabal project files needs to be reverted as that is not needed to run locally or in CI/CD builds for now

khuzema786:
Got it! Let me find all "safety-dashboard" references in the nix and cabal files on this branch.Now let me check which of these safety-dashboard references were actually added by this PR (vs pre-existing on main).I have a clear picture now. Three files need surgical edits. Let me read them and make the changes.Now let me find the exact location of the `safety-dashboard-exe` block in `nammayatri.nix`.I now have the full picture. Three targeted edits needed — no other safety-dashboard lines were added by this PR. Let me make them all.Now let me verify the final state of all three files and confirm no safety-dashboard references remain in the changed files.The remaining `safety-dashboard-seed.sql` reference at line 1249 was **already in `main` before this PR** — it's not part of these changes, so it stays. Now let me commit.Good — my three edits show as 12 deletions. Let me do a quick sanity check on the edited files before committing.Diff looks exactly right. Now committing.Committed. Would you like me to push this to the `backend-devBox` branch on GitHub to update the PR?

Here's a summary of what was done:

**Committed: `DevBox: remove safety-dashboard from nix and cabal project`**

| File | Change |
|------|--------|
| `Backend/cabal.project` | Removed `app/safety-dashboard` from packages list |
| `Backend/nix/services/ports.nix` | Removed `safety-dashboard = 8025` and `safety-dashboard-metrics = 9993` |
| `Backend/nix/services/nammayatri.nix` | Removed `"safety-dashboard-exe"` from `cabalExecutables` list and its entire environment block |

**Left untouched** (pre-existing in `main`, not added by this PR):
- `safety-dashboard-seed.sql` reference in the DB seed list
- `Backend/default.nix` and `Backend/nix/scripts.nix` safety-dashboard refs (were already there before this branch)

**Other issues confirmed clear:**
- `App.hs` already uses `readMaybe` — no fix needed
- `reset_all_stages` deadlock — not a real issue; `_kill_session` never acquires `_runner_lock`

khuzema786:
@ClaudeWeb remove this service from nix too

 cause we add
safety-dashboard-exe = {
            environment = {
              SERVICE_PORT = toString ports.safety-dashboard;
              METRICS_PORT = toString ports.safety-dashboard-metrics;
              DRIVER_APP_PORT = toString ports.dynamic-offer-driver-app;
            };
          };
in nammayatri.nix so safety-dashboard.dhall get update SERVICE_PORT, METRIX_PORT and DRIVER_APP_PORT not hardcoded ports, and if we write a service in nammayatri.nix then we have to add it in cabalExecutables, other wise error will come from nix

khuzema786:
Understood — since the `safety-dashboard-exe` nix service block (which set those env vars) is now gone, the `safety-dashboard.dhall` changes that relied on `SERVICE_PORT`, `METRICS_PORT`, and `DRIVER_APP_PORT` also need to be reverted back to hardcoded values. Let me check what changed there.Clear — since the nix service that sets those env vars is removed, `safety-dashboard.dhall` should be reverted to its hardcoded values as well. Let me restore it to the `main` baseline.File matches `main` exactly. Now commit and push.Done. Commit `a9cc193` is pushed. Here's the full picture of what the two new commits on the PR now do:

| Commit | What it removes |
|--------|----------------|
| `1866dd0` | `app/safety-dashboard` from `cabal.project`; `safety-dashboard` ports from `ports.nix`; `safety-dashboard-exe` entry + env block from `nammayatri.nix` |
| `a9cc193` | Reverts `safety-dashboard.dhall` — restores hardcoded ports (`+8025`, `+9993`, `8016`, etc.) since the nix service that would have set those env vars is gone |

PR #14947, opened 2026-05-27
enh: update the upsert dashboard api for vehicleSeatLayoutMapping to …

MohamedHamdanA:
…handle multiple entries at once

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Batch upserts for vehicle seat layout mappings (submit multiple entries per request).
  * List responses now include an optional seat selection type for mappings.

* **Refactor**
  * Payloads and backend processing updated to accept per-entry seat selection, deduplicate entries, and update or create mappings in a single request.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
@@ -6,6 +6,8 @@ module Domain.Action.Dashboard.AppManagement.VehicleSeatLayoutMapping
 where
 
 import qualified API.Types.Dashboard.AppManagement.VehicleSeatLayoutMapping as API
+import Data.List (nub)
+import qualified Data.Map.Strict as Map
 import qualified Domain.Types.Merchant
 import qualified Domain.Types.VehicleSeatLayoutMapping
 import qualified Environment
@@ -36,7 +38,8 @@ listVehicleSeatLayoutMapping _merchantShortId _opCity limit offset gtfsId = do
         { id = m.id,
           vehicleNo = m.vehicleNo,
           gtfsId = m.gtfsId,
-          seatLayoutId = m.seatLayoutId
+          seatLayoutId = m.seatLayoutId,
+          seatSelectionType = m.seatSelectionType
         }
 
 upsertVehicleSeatLayoutMapping ::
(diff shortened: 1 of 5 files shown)


khuzema786:
@ClaudeWeb Can you review this PR ?

khuzema786:
I'll fetch the PR details, diff, and existing comments to do a thorough review.Let me try using the GitHub MCP tool and the fetch tool instead.Now let me fetch the PR details, files, and comments in parallel.Now let me read the full domain action file in the workspace to verify context around the changes.I have all the context I need. Let me also quickly fetch the full domain action file from GitHub for complete analysis.Now I have everything needed for a complete analysis. Let me write the review.

---

## PR Review: `enh/vehicleSeatLayoutMapping` — Batch upsert for vehicle seat layout mappings

**Verdict: Request Changes** — The core logic is sound and the three CodeRabbit issues were correctly addressed, but there are two bugs that need attention before merge.

---

### ✅ What's Done Well

The three issues CodeRabbit flagged in the first iteration were properly fixed:
- Correctly separated `UpsertReq` (wrapper) from `UpsertEntry` (input row) — `VehicleSeatLayoutMappingItem` (which has an `id`) is no longer reused as input.
- `foldM` accumulator pattern correctly handles duplicate `(gtfsId, vehicleNo)` pairs within the same request — the map is updated on create so the second occurrence goes down the `Just` (update) path.
- `seatSelectionType` is now persisted in both the create and update paths.

---

### 🔴 Critical

#### 1. Breaking API contract — unchecked in PR description
**Confidence: 95** | `VehicleSeatLayoutMapping.yaml`, `Endpoints/VehicleSeatLayoutMapping.hs`

The request body shape changed completely. Any existing caller sending:
```json
{ "vehicleNo": "MH12AB1234", "gtfsId": "route-1", "seatLayoutId": "abc" }
```
will now receive a deserialization error because the endpoint expects:
```json
{ "entries": [{ "vehicleNo": "MH12AB1234", "gtfsId": "route-1", "seatLayoutId": "abc" }] }
```

The PR description has the **"This PR contains API breaking changes"** checkbox **unchecked**. Even if this is a dashboard/internal API, existing integrations (scripts, tooling) will break silently. Please check the box and document the migration path.

---

### 🟡 Warning

#### 2. No transaction boundary — partial upsert on failure
**Confidence: 85** | `Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs` (lines 52–92)

The `foldM` loop processes each entry with independent DB calls. If the operation fails midway (e.g., a DB error on entry 7 of 20), entries 1–6 are permanently committed while 7–20 are not. There is no rollback.

```haskell
-- Current: each create/update is a separate, unguarded transaction
void $
  foldM
    (\mappingMap entry -> do
        ...
        Queries.create mapping   -- committed immediately, no rollback if next entry fails
    )
    initialMap
    req.entries
```

For a dashboard admin API this may be acceptable in practice, but callers have no way to distinguish a complete success from a partial one since the endpoint always returns `Success`. Consider wrapping in a transaction, or at minimum returning which entries succeeded/failed.

#### 3. Cache not invalidated for newly-created entries
**Confidence: 72** | `Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs`

The update branch correctly calls `invalidateCache`, but the create branch does not:

```haskell
Nothing -> do
  ...
  Queries.create mapping
  pure $ Map.insert key mapping mappingMap
  -- ☝ missing: CQVehicleSeatLayoutMapping.invalidateCache entry.vehicleNo entry.gtfsId
```

If `CQVehicleSeatLayoutMapping` uses negative caching (i.e., it caches `Nothing` results for absent keys), a prior cache miss will remain stale after the row is created, causing subsequent reads to return `Nothing` until TTL expiry. Check whether `VehicleSeatLayoutMappingExtra` caches negative lookups and add the `invalidateCache` call if so.

#### 4. N+1 queries in the prefetch step
**Confidence: 75** | `Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs` (lines 57–58)

```haskell
let uniqueGtfsIds = nub $ map (.gtfsId) req.entries
existingMappings <- concatMapM (\gid -> Queries.findAllByGtfsId Nothing Nothing gid) uniqueGtfsIds
```

This fires **one DB round-trip per unique gtfsId**. For a batch with 30 unique gtfsIds, that's 30 queries before the loop even starts. A single query with an `IN (...)` clause would be more efficient. The current `findAllByGtfsId` signature doesn't support that, but a new query function could:

```haskell
-- Suggested addition to Storage.Queries.VehicleSeatLayoutMapping
findAllByGtfsIds :: [Text] -> m [VehicleSeatLayoutMapping]
findAllByGtfsIds gtfsIds =
  findAllWithKV [Se.Is Beam.gtfsId $ Se.In gtfsIds]
```

---

### 💡 Suggestion

#### 5. `Data.List.nub` is O(n²)
**Confidence: 65** (minor, but worth fixing since you're already importing `Data.Map.Strict`)

```haskell
-- Current — O(n²)
let uniqueGtfsIds = nub $ map (.gtfsId) req.entries

-- Better — O(n log n)
import qualified Data.Set as Set
let uniqueGtfsIds = Set.toList . Set.fromList $ map (.gtfsId) req.entries
```

For small batches this doesn't matter, but it's a trivial fix.

---

### Summary

| # | Severity | Issue | File |
|---|----------|-------|------|
| 1 | 🔴 Critical | Breaking API change not acknowledged | YAML spec / Endpoints.hs |
| 2 | 🟡 Warning | No transaction — partial upsert possible | Domain Action |
| 3 | 🟡 Warning | No cache invalidation on create | Domain Action |
| 4 | 🟡 Warning | N+1 prefetch queries | Domain Action |
| 5 | 💡 Suggestion | `nub` is O(n²) | Domain Action |

Items 2–5 are all in `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs`. The generated files (`src-read-only/`) look consistent with the YAML spec changes — no issues there.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
```diff
@@ -49,27 +52,41 @@ upsertVehicleSeatLayoutMapping merchantShortId opCity req = do
...
-                seatSelectionType = req.seatSelectionType,
-                merchantId = merchant.id,
-                merchantOperatingCityId = merchantOperatingCity.id,
-                createdAt = now,
-                updatedAt = now
-              }
-      Queries.create mapping
+
+  let uniqueGtfsIds = nub $ map (.gtfsId) req.entries
+  existingMappings <- concatMapM (\gid -> Queries.findAllByGtfsId Nothing Nothing gid) uniqueGtfsIds
```
Jayanth-Parthsarathy: instead of doing a db query for each gtfs_id we can use Se.In with an array of Ids

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/AppManagement/VehicleSeatLayoutMapping.hs
```diff
@@ -49,27 +52,41 @@ upsertVehicleSeatLayoutMapping merchantShortId opCity req = do
...
+  let initialMap = Map.fromList [((m.gtfsId, m.vehicleNo), m) | m <- existingMappings]
+
+  void $
+    foldM
+      ( \mappingMap entry -> do
+          now <- getCurrentTime
+          let key = (entry.gtfsId, entry.vehicleNo)
+          case Map.lookup key mappingMap of
+            Just _ -> do
+              Queries.updateSeatLayoutIdByVehicleNoAndGtfsId entry.seatLayoutId entry.seatSelectionType entry.vehicleNo entry.gtfsId
```
Jayanth-Parthsarathy: Are we protecting against non null values (what if entry.seatSelectionType is Nothing (null).

Jayanth-Parthsarathy (approved):

PR #14954, opened 2026-05-27
backend/enh/Move-Config-Pilot-To-Lib

H-C-21:
Moved ConfigPilot to Common Lib
Added Fallback and disable option for config-pilot
Added some driver config tables to ConfigPilot (getters).
Namma-DSL for auto generating ConfigPilot Files

## Release Notes

**Chores**
  * Updated configuration management infrastructure to use ConfigPilot for consistent handling of app settings across driver and rider platforms.
  * Refactored configuration retrieval mechanisms to improve system modularity and maintainability.


piyushKumar-1 (dismissed):

H-C-21:
WIP, Do not Merge.

piyushKumar-1:
@ClaudeWeb can you verify if the lib migration of configpilot files backward compatible, doesn't change the behaviour of how it was when it was just in rider-app and is properly reusable across driver and rider app

khuzema786:
I'll fetch the PR diff, description, and existing comments in parallel, then dig into the source files for context.Let me try using the GitHub MCP tools instead.Good, the GitHub MCP tools are available. Let me fetch the PR details, changed files, and comments in parallel.Good, I have the PR details and two existing CodeRabbit comments. Now let me deeply read the actual source files to verify behavior and backward compatibility.The workspace is on main (pre-PR). I need to read the actual PR diff to see what's being added. Let me fetch the diff and the key existing files simultaneously.The workspace doesn't have the source code. Let me read the persisted diff output and fetch subsequent pages of the diff.Let me fetch the rest of the diff to see the new library files being added.Let me jump to the end of the diff where the new lib files are likely added.I can see driver-app config calls being replaced with `getOneConfig`. Let me jump further to find the new lib files.Let me search for the new lib files directly from GitHub.Let me search for the correct file paths in the PR branch.Let me read the fetched content and try different paths for the lib.Let me read the persisted PR files list and search for the lib path.Let me fetch the key new lib files and the auto-generated driver-app ConfigPilot files directly from the PR branch.Excellent! Now let me fetch the rider-app's updated ConfigPilot files and the driver-app's auto-generated ones.Now I have everything I need. Let me compile the full analysis.I now have the full picture. Let me write the review.

---

## PR Review: `backend/enh/Move-Config-Pilot-To-Lib` (#14954)

**Verdict: Request Changes.** The lib extraction is structurally sound and reusable. However, two silent behavioral regressions and one latent crash path need to be addressed before merge.

---

### 🔴 Critical

#### 1. `LogicDomain` enum values changed for two existing rider-app configs — breaks active Yudhishthira rollouts

**File:** `Backend/app/rider-platform/rider-app/Main/src/Storage/ConfigPilot/Config/PayoutConfig.hs` and `RideRelatedNotificationConfig.hs`

`getConfigImpl` queries `AppDynamicLogicRollout` by `logicDomain`. The enum values used changed:

| Config | Old value | New value |
|---|---|---|
| `PayoutConfig` | `RIDER_CONFIG PayoutConfig` | `RIDER_CONFIG PayoutConfigRider` |
| `RideRelatedNotificationConfig` | `RIDER_CONFIG RideRelatedNotificationConfig` | `RIDER_CONFIG RideRelatedNotificationConfigRider` |

Any DB rows in `app_dynamic_logic_rollout` keyed on the old domain strings will **silently stop applying**. Callers don't error — they just get base config with no dynamic overrides. This is a silent correctness regression for any merchant with active rollouts on these tables.

**Action required:** Before merging, either:
- Confirm no merchants have active rollout entries for the old domain values, **or**
- Run a migration to rename those DB rows to the new domain strings.

The `RiderConfig` ConfigType correctly keeps its value unchanged (`RiderConfig`). Only the two above changed. The in-mem cache key also changes (it embeds `getConfigType`), so existing cache entries auto-expire harmlessly.

---

### 🟠 Warning

#### 2. Cache-collision: Maybe-dimension filters are inside the cache boundary but excluded from the cache key (pre-existing, now standardised in lib)

**File:** `Backend/lib/config-pilot/src/Lib/ConfigPilot/Interface/Getter.hs:67`

`configPilotInMemKey` uses `dimensionsCacheKey` which — by design — excludes `Maybe` fields. But `resolveConfigList` caches the **already-filtered** results:

```haskell
resolveConfigList dims logicDomain mocId fetch matchers fallback =
  IM.withInMemCache (configPilotInMemKey dims) 3600 $ do   -- key = ["ConfigPilot","PayoutConfigRider","<mocId>"]
    cfgs <- fetch
    let filtered = applyMatchers ...   -- ← depends on vehicleCategory, isPayoutEnabled, payoutEntity
    ...
```

Two calls with the same `merchantOperatingCityId` but different `vehicleCategory` (e.g. `Just CAR` vs `Just BIKE`) share the same cache key. The second caller gets the first caller's filtered results for the full 3600 s TTL.

This existed in the old per-config implementations too. But now it's baked into the shared `resolveConfigList` and will affect every new config that adds `DimMatcher`s with Maybe dims.

**Fix (CodeRabbit's suggestion is correct):** Cache only the unfiltered wrapper list; apply filtering and `getConfigImpl` outside the cache boundary:

```haskell
resolveConfigList dims logicDomain mocId fetch matchers fallback = do
  wrappers <- IM.withInMemCache (configPilotInMemKey dims) 3600 $ do
    cfgs <- fetch
    let applyMatchers ms = filter (\c -> all (matchesDim dims c) ms) cfgs
    let filtered = case (applyMatchers matchers, fallback) of
          ([], Just fb) -> applyMatchers fb
          (results, _) -> results
    pure $ map (\c -> LYT.Config {config = c, extraDimensions = Nothing, identifier = 0}) filtered
  mapM (\w -> getConfigImpl dims w logicDomain mocId) wrappers
```

This also fixes CodeRabbit's second issue: `getConfigImpl` (which makes txn-sticky rollout decisions and random percentage tosses) is no longer inside the 3600 s cache — so rollout decisions are re-evaluated per call as intended.

#### 3. `DISABLE_CONFIG_PILOT=true` with no fallback throws hard `InternalError`

**File:** `Backend/lib/config-pilot/src/Lib/ConfigPilot/Interface/Types.hs` (the `getConfig` default implementation)

```haskell
if disableConfigPilot
  then fromMaybe (throwM $ InternalError $ "No Fallback configured for table: " <> tableName) mbFallback
```

Rider-app callers that use `Nothing` as fallback (`RiderConfig`, `PayoutConfig`, `RideRelatedNotificationConfig`, etc.) will crash with a hard `InternalError` when the env var is set. This is a bad developer/ops experience: toggling a kill-switch should degrade gracefully, not crash the pod.

**Suggested fix:** If `DISABLE_CONFIG_PILOT=true` and no fallback is provided, fall through to `getConfigList` directly (skipping Yudhishthira resolution) rather than erroring:

```haskell
if disableConfigPilot
  then case mbFallback of
    Just fb -> fb
    Nothing -> filterByDimensions dims <$> getConfigList dims  -- bare DB path
```

Or at minimum, document clearly that callers *must* provide a fallback for kill-switch compatibility.

---

### 🟡 Potential issues

#### 4. `lookupEnv "DISABLE_CONFIG_PILOT"` is called on every config fetch

**File:** `Types.hs` — `getConfig` default impl

```haskell
disableConfigPilot <- liftIO $ maybe False read <$> lookupEnv "DISABLE_CONFIG_PILOT"
```

This is a `read` syscall on every hot-path config access. Env vars don't change at runtime; this should be read once at app startup and injected through the `r` environment record.

#### 5. `ScheduledPayoutConfig` and `ReminderConfig` use direct-query modules, not cached queries

**File:** `spec/Storage/Merchant.yaml`, `spec/Storage/ReminderConfig.yaml`

```yaml
# ScheduledPayoutConfig
queryModule: Storage.Queries.ScheduledPayoutConfig    # ← direct DB, no Redis

# ReminderConfig
queryModule: Storage.Queries.ReminderConfig           # ← direct DB, no Redis
```

Every in-mem cache miss for these configs will hit the DB directly. All other configs use `Storage.CachedQueries.*`. Confirm this is intentional (e.g. `CachedQueries` doesn't exist yet for these tables), otherwise it should be `Storage.CachedQueries.ScheduledPayoutConfig`.

---

### ⚪ Observations / Info

#### 6. `MorthVerification.documentNumber` encryption removed — unrelated to ConfigPilot

**File:** `spec/Storage/DriverOnboarding.yaml`

```yaml
-  beamFields:
-    documentNumber:
-      documentNumberEncrypted: Text
-      documentNumberHash: DbHash
-  fromTType: ...
-  toTType: ...
```

This removes PII encryption/decryption for a document number field. It appears bundled unintentionally with this PR. This change should be reviewed separately with a security/compliance lens.

#### 7. `resolveConfigList` doc-comment example is missing the 6th argument

**File:** `Getter.hs:43-48`

The `-- >` usage example shows only 5 arguments but the actual function takes 6 (the `Maybe [DimMatcher...]` fallback). Will mislead implementors.

```haskell
-- >     []         -- matchers
-- >     Nothing    -- ← missing from the example
```

---

### ✅ What looks good

- **Backward-compatible reuse:** Driver-app `getOneConfig` calls always provide a `Just fallback` pointing to the old direct-DB function, so pre-ConfigPilot behavior is fully preserved on failure or when no rollout is configured.
- **`matchesDim` semantics:** Correctly replicate the old per-config `filterByDimensions` logic. `vehicleCategory` (Maybe→Maybe), `isPayoutEnabled` (Maybe→`Just . bool`), `payoutEntity` (Maybe→`Just . value`) are all semantically equivalent to the old explicit `filterByDimensions` implementations.
- **`RiderConfig` cache key is preserved:** `RiderConfig` kept its ConfigType value; in-mem cache keys are unchanged for callers that were already live.
- **`invalidateConfigInMem` and `TxnIdKey`/`PersonIdKey` are correctly re-exported** from the lib so rider-app callers don't need to change import paths for those.
- **DSL-driven code generation** for driver-app ConfigPilot files reduces boilerplate and ensures consistency — good pattern.

piyushKumar-1 (approved):

PR #14966, opened 2026-05-29
backend/fix: stripe payout without vpa

roman-bodavskiy:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Release Notes

* **Improvements**
  * Updated scheduled payout eligibility criteria to conditionally require virtual payment address based on merchant's payment service configuration, allowing drivers and fleet owners to be eligible for payouts based on their specific service setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nammayatri/nammayatri/pull/14966?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/Payout/ScheduledBatchPayout.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/Payout/ScheduledBatchPayout.hs
@@ -23,12 +23,15 @@ import Domain.Action.UI.DriverWallet
   )
 import Domain.Action.UI.Ride.EndRide.Internal (makeWalletRunningBalanceLockKey)
 import qualified Domain.Types.DriverInformation as DI
+import Domain.Types.Extra.Plan
 import qualified Domain.Types.FleetOwnerInformation as DFOI
 import qualified Domain.Types.Merchant as DM
 import qualified Domain.Types.MerchantOperatingCity as DMOC
+import qualified Domain.Types.MerchantServiceConfig as DEMSC
 import qualified Domain.Types.Person as DP
 import qualified Domain.Types.ScheduledPayoutConfig as DSPC
 import qualified Domain.Types.TransporterConfig as DTConf
+import qualified Kernel.External.Payout.Interface as Payout
 import Kernel.External.Types (SchedulerFlow)
 import Kernel.Prelude
 import Kernel.Storage.Esqueleto.Config (EsqDBReplicaFlow)
@@ -51,6 +54,7 @@ import qualified Storage.Queries.DriverInformationExtra as QDIE
 import qualified Storage.Queries.FleetOwnerInformationExtra as QFOIE
 import qualified Storage.Queries.Person as QPerson
 import qualified Storage.Queries.ScheduledPayoutConfig as QSPC
+import qualified Tools.Payout as TPayout
 
 --------------------------------------------------------------------------------
 -- Job entry point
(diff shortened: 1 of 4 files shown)


roman-bodavskiy:
@ClaudeWeb please review

khuzema786:
I'll gather the full PR context before reviewing.Let me try the diff directly and look for MCP GitHub tools.Good, I have the diff. Let me now read the surrounding source files for deeper context.Let me fetch the relevant source files via GitHub to understand the full context.Now I have enough context. Let me analyze the changes thoroughly.Now I have full context. Here's my review:

---

## PR Review: `backend/fix: stripe payout without vpa`

**Verdict: Approve with minor comments.** The fix is logically correct — Stripe doesn't use VPA/UPI addresses, so the hard-coded `payoutVpa IS NOT NULL` filter was wrongly blocking all Stripe-based payouts. The approach is sound and consistent with how `initiateWalletPayout` already determines the payout service.

---

### ✅ What's correct

**Consistency with `initiateWalletPayout`:** The new code uses `SubscriptionConfigOption PREPAID_SUBSCRIPTION` to determine VPA requirement, which is exactly what `DriverWallet.hs:initiateWalletPayout` uses when it calls `getCreatePayoutServiceFlow`. This ensures the eligibility filter and the actual payout call agree on which service is in use.

**Query change is minimal and correct** (`DriverInformationExtra.hs`, `FleetOwnerInformationExtra.hs`): Conditional filter appended via `<>` list composition — no structural change to existing filters, clean pattern.

**Safe fallback:** In `getPayoutServiceFlowForMerchant`, when `subscriptionConfig.payoutServiceName` is `Nothing`, it defaults to `PT.Juspay`, which means `isPayoutVpaRequired = True`. Conservative default — correct.

---

### ⚠️ Warning — Stripe drivers without a bank account will generate per-driver error logs at payout time

**File:** `ScheduledBatchPayout.hs` / `DriverWallet.hs:initiateWalletPayout`
**Confidence: 72%**

When `isPayoutVpaRequired = False`, the query now includes Stripe-enrolled drivers **regardless of whether they have a `DriverBankAccount` row**. When `initiateWalletPayout` later calls `getCreatePayoutServiceFlow`, the Stripe branch does:

```haskell
-- Tools/Payout.hs (getPayoutServiceFlow → fetchPersonBankAccount)
Payout.StripeFlow -> Just <$> (QDBA.findByPrimaryKey personId >>= fromMaybeM (InvalidRequest "Driver bank aсcount not found"))
```

For Stripe drivers with no bank account, this throws, which gets caught by the `try` in `processOneWalletPayout` and silently logged. The job won't crash, but a merchant onboarding Stripe for the first time with incomplete driver profiles will see a flood of per-driver error logs that look like failures but are actually just incomplete data.

**Consider noting this in the PR description**, or adding a metric/counter rather than only a `logError`. No code change required for correctness — the graceful failure path already exists.

---

### 💡 Suggestion — `getPayoutServiceFlowForMerchant` duplicates service name resolution logic

**File:** `Tools/Payout.hs:133-160`
**Confidence: 65%**

The `payoutServiceNameRaw` resolution block in the new function is identical to the one inside `getPayoutServiceFlow`. If the resolution logic ever changes (e.g. a new `payoutServiceNameOption` variant is added), it would need updating in two places.

A small refactor would extract a shared helper:

```haskell
resolvePayoutServiceName ::
  ServiceFlow m r =>
  (DMSUC.MerchantServiceUsageConfig -> Payout.PayoutService) ->
  PayoutServiceNameOption ->
  (PT.PayoutService -> DMSC.ServiceName) ->
  Id DMOC.MerchantOperatingCity ->
  m DMSC.ServiceName
resolvePayoutServiceName getCfg payoutServiceNameOption serviceType merchantOperatingCityId = ...
```

Then both `getPayoutServiceFlowForMerchant` and `getPayoutServiceFlow` delegate to it. Not required for this fix, but worth tracking as a follow-up.

---

### ✅ No issues with

- Module export list (`getPayoutServiceFlowForMerchant` correctly exported)
- Unqualified `import Domain.Types.Extra.Plan` bringing `PREPAID_SUBSCRIPTION` into scope — correct
- Cursor-based pagination — unaffected by the change
- The `_ -> throwError $ InternalError "Unknown Service Name"` fallback in the new function — consistent with the existing `getPayoutServiceFlow` pattern

khuzema786 (approved):

PR #14970, opened 2026-05-29
Added driver mobile number support for fetch

RitoBose:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Added driver mobile number support for fetch API of HubRequests 

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Operator dashboard: filter hub requests by driver mobile number; existing creator-mobile filtering remains supported.

* **Tests**
  * Added end-to-end integration tests for driver-mobile filtering, negative cases, combined filters, and response validations.

* **Chores**
  * Added local Postman environment, SQL seed for Operation Hub, and a new test-run command to execute the Operation Hub test suite.


RitoBose:

khuzema786:
I'll fetch the PR diff, description, and comments in parallel to do a thorough review.Let me try the MCP GitHub tools or use the API endpoint directly.Now let me get the full diff and existing comments.Now let me read the full diff output and key source files.Let me look at the existing storage query file for full context on the join pattern.Let me fetch the full storage query file from GitHub for context.Now I have everything I need. Let me write the full review.

---

## PR Review: Added driver mobile number support for fetch (#14970)

**Verdict: Request Changes.** The feature logic is sound and the parameter propagation is correct across all layers, but there are two concrete bugs — one in the integration tests that will cause all status-filtered cases to return 400, and a missed opportunity to eliminate a N+1 query that the new JOIN already joins the data for.

---

### 🔴 Bugs

#### 1. Postman `mbStatus` value is double-quoted — all filtered tests will return 400
**File:** `Backend/dev/integration-tests/collections/OperationHubFlow/01-FetchHubRequestsByDriverMobile.json`  
**Confidence:** 88

Steps 07–11 all define the status filter like this:

```json
{ "key": "mbStatus", "value": "\"PENDING\"" }
```

In JSON, `"\"PENDING\""` decodes to the 9-character string `"PENDING"` (with surrounding double-quote characters). Postman URL-encodes this to `mbStatus=%22PENDING%22`. Servant's `fromHttpApiData` for `RequestStatus` expects the bare string `PENDING`, so it fails to parse and returns **400** — failing the `pm.response.to.have.status(200)` assertion in every step that uses this parameter.

**Fix:** Remove the extra backslash-quotes in every occurrence:

```json
{ "key": "mbStatus", "value": "PENDING" }
```

And in the raw URLs:
```
?mbStatus=PENDING   (not ?mbStatus="PENDING")
```

---

#### 2. Joined `driverPerson` is discarded — N+1 query still fires per result
**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/OperationHubRequestsExtra.hs` (line ~184)  
**Confidence:** 82

The new LEFT JOIN correctly fetches `driverPerson` for the filter, but the result tuple's fourth element is immediately thrown away:

```haskell
-- PR code
finalRes <- forM res' $ \(operationHubRequests, person, operationHub, _) -> runMaybeT $ do
```

Then `castHubRequests` in `Domain/Action/Dashboard/Operator/Driver.hs` issues a **separate DB round-trip per row** to fetch the driver's phone number:

```haskell
driverPhoneNo <- case hubReq.driverId of
  Just driverId -> do
    driver <- QPerson.findById driverId >>= fromMaybeM (PersonNotFound driverId.getId)   -- N+1
    mapM decrypt driver.mobileNumber
  Nothing -> pure Nothing
```

The PR already joins the driver person in the storage layer — but only keeps it long enough to filter, then discards it. This means the query still does N extra lookups.

**Fix:** Propagate the joined driver person through. Change the storage function return type to include an optional driver:

```haskell
-- Storage return type
m [(OperationHubRequests, Person, DOH.OperationHub, Maybe Person)]

-- Result mapping (don't discard the 4th element)
finalRes <- forM res' $ \(operationHubRequests, person, operationHub, mbDriver) -> runMaybeT $ do
  ohr <- MaybeT $ fromTType' operationHubRequests
  p   <- MaybeT $ fromTType' person
  oh  <- MaybeT $ fromTType' operationHub
  drv <- lift $ mapM fromTType' mbDriver
  pure (ohr, p, oh, join drv)
```

Then `castHubRequests` can accept a `Maybe Person` instead of re-querying:

```haskell
castHubRequests (hubReq, creator, hub, mbDriver) = do
  creatorPhoneNo <- mapM decrypt creator.mobileNumber
  driverPhoneNo  <- mapM (mapM decrypt . (.mobileNumber)) mbDriver >>= pure . join
  ...
```

> Note: this N+1 was pre-existing. The PR is the first time the joined driver data is available, making it the natural place to fix it. Since the query already pays for the LEFT JOIN, not using the result is a pure waste.

---

### 🟡 Warnings

#### 3. LEFT JOIN + IS NOT NULL filter — `\_ ->` lambda hides the filtering intent
**File:** `Storage/Queries/OperationHubRequestsExtra.hs`  
**Confidence:** 68

```haskell
-- Current approach
driverPerson <- B.leftJoin_' ... (\dp -> maybe (B.sqlBool_ $ B.val_ False)
    (\hash -> ... B.&&?. BeamP.mobileNumberHash dp B.==?. B.val_ (Just hash))
    mbDriverMobileNumberHash)
...
B.&&?. maybe (B.sqlBool_ $ B.val_ True)
    (\_ -> B.sqlBool_ $ B.not_ $ B.isNothing_ driverPerson.id)  -- ← ignores the lambda arg
    mbDriverMobileNumberHash
```

The approach is logically correct: the hash check is embedded in the JOIN condition, and the WHERE clause checks that the join matched. But `\_ -> ...` hiding the hash value is a readability trap — a future reader seeing the WHERE clause alone cannot tell that the hash is already encoded in the join. When `mbDriverMobileNumberHash = Nothing`, the join condition is `False`, forcing driverPerson to always be NULL, which is handled correctly since the WHERE filter is `True` in that case.

**Suggestion:** A brief comment over the join block would prevent future confusion:

```haskell
-- LEFT JOIN on driverPerson with the hash embedded in the ON clause; 
-- the WHERE check (isNothing_ driverPerson.id) then acts as an INNER JOIN filter
-- only when mbDriverMobileNumberHash is provided.
```

#### 4. Both auth flows share a single `login_otp` variable
**File:** `Backend/dev/integration-tests/collections/OperationHubFlow/Local/Local_NY_Bangalore.postman_environment.json`  
**Confidence:** 65

Steps 01–02 (fleet owner auth) and 03–04 (driver auth) both use `{{login_otp}}` = `7891`. In a local mock environment this is fine. In any environment where the OTP service issues different codes per user, both auth flows would use the wrong OTP for at least one of the two users. Consider either capturing each OTP response separately or documenting this assumption explicitly.

---

### ✅ What looks good

- Parameter propagation is mechanically correct through all layers: YAML spec → Endpoints type → provider-dashboard handler/domain → driver-app handler/domain → storage query. No mismatches.
- The `OperationHub.hs` UI callsite correctly passes `Nothing` for the new argument (it doesn't apply driver mobile filtering in the user-facing flow).
- `getDbHash` is correctly used to hash the incoming plain-text mobile number before comparing with the DB-stored hash — same pattern as `mbMobileNumber`.
- The rename from `driver` → `creator` in the join variable is a clear correctness improvement for readability.
- Integration test covers the important cases: match, no-match, backward compat with creator mobile, both filters combined.

Backend/dev/integration-tests/collections/OperationHubFlow/setup-local-operation-hub.sql
```diff
@@ -0,0 +1,28 @@
...
+  'HSR Layout, Bangalore',
+  12.9141,
+  77.6446,
+  '9999900000',
+  moc.merchant_id,
+  moc.id,
+  now(),
+  now()
+FROM atlas_driver_offer_bpp.merchant_operating_city moc
+WHERE moc.merchant_short_id = 'NAMMA_YATRI_PARTNER'
```
0utkarsh, line 21: Hmm, okay

0utkarsh (approved):

PR #14998, opened 2026-06-01
backend/enhc: stripe refunds flow

TejaswiniT167:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Admins can initiate refund requests by ride from the dashboard; internal refund-ledger endpoint added.

* **Improvements**
  * Rider payment view now returns a list of refund requests per ride; refresh query removed.
  * Option to deduct refund amounts from driver earnings available across refund flows.
  * Cumulative refund invoice generation and improved ledger handling; invoice types and numbering include refunds.

* **Bug Fixes**
  * Better idempotency, retry, status-sync and history tracking for refund attempts.


Backend/dev/ddl-migrations/dynamic-offer-driver-app/0840-rename-finance-invoice-paymentOrderId-to-entityReferenceId.sql
```diff
@@ -0,0 +1,15 @@
...
+-- payment_order_id, so drop the empty column and RENAME the old one across it —
+-- pure DDL, preserves data, no row rewrite. Guarded: no-op once payment_order_id is gone.
+DO $$
+BEGIN
+  IF EXISTS (
+    SELECT 1 FROM information_schema.columns
+    WHERE table_schema = 'atlas_driver_offer_bpp' AND table_name = 'finance_invoice'
+      AND column_name = 'payment_order_id'
+  ) THEN
+    ALTER TABLE atlas_driver_offer_bpp.finance_invoice DROP COLUMN IF EXISTS entity_reference_id;
```
khuzema786: @TejaswiniT167 We can't do this on existing table during the release, why do we want to rename `payment_order_id` to `entity_reference_id` and drop existing `entity_reference_id` ?

We can use entity_reference_id only and stop adding anything in payment_order_id. This `payment_order_id` we can remove from code then if it is currently not being used anywhere and is having all null values only in Prod

khuzema786:
@ClaudeWeb Please Review this PR

ClaudeWeb:
I'll fetch the PR diff, description, and comments in parallel now.Let me read the full diff and explore the workspace structure.Let me read more of the diff, especially the critical `RefundLedger.hs` domain logic.Let me read the remaining sections of the diff.Let me read the remaining portions of the diff covering the RidePayment domain logic.Let me read the remaining diff sections.Let me read the final portion of the diff to check remaining files and migrations.I now have a thorough understanding of all the changes. Here is my complete review.

---

## PR #14998 — `backend/enhc: stripe refunds flow` · Review

This is a substantial, well-structured rework of the Stripe refund flow: flat `approvedAmount` is replaced by a per-component (RIDE_FARE/TOLL/PARKING) model with full BAP↔BPP ledger synchronization. The core design is sound and the integration tests are a genuine strength. However there are several correctness and safety issues that need fixing before merge.

---

### 🔴 Critical

#### 1. BAP ledger write failure silently swallowed before Stripe call

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs`  
**Line:** `processRefundRaised` function

```haskell
void $ RidePaymentFinance.createRefundRaisedLedger ctx refundRequest.id splits
-- ... continues to submit to Stripe regardless
```

`createRefundRaisedLedger` returns `m (Either FinanceError [...])`. Using `void $` silently discards a `Left err` result and proceeds to call Stripe. If the BAP ledger write fails, money moves at Stripe with no BAP ledger record — a financial inconsistency.

**Fix:**
```haskell
result <- RidePaymentFinance.createRefundRaisedLedger ctx refundRequest.id splits
case result of
  Left err -> throwError $ InternalError $ "Refund ledger write failed: " <> show err
  Right _  -> pure ()
```

---

#### 2. Missing DDL migration for `refund_request` table

The schema adds 3 new columns to `refund_request`:
- `requested_refund_components` (json)
- `approved_refunded_components` (json)
- `deduct_from_driver` (boolean)

The PR includes DDL for `finance_ledger_entry` (migrations `0839`/`1550`) and `finance_invoice` (`0840`/`1551`), but there is no corresponding migration for `refund_request` itself. Without it, the app will throw `ColumnNotFound`/`BeamRowReadError` at runtime on the first refund request read or write.

**Fix:** Add a guarded migration (e.g. `1552-refund-request-add-component-and-deduct-columns.sql`):
```sql
DO $$ BEGIN
  IF NOT EXISTS (SELECT 1 FROM information_schema.columns
    WHERE table_schema = 'atlas_app' AND table_name = 'refund_request'
      AND column_name = 'requested_refund_components') THEN
    ALTER TABLE atlas_app.refund_request
      ADD COLUMN requested_refund_components json,
      ADD COLUMN approved_refunded_components json,
      ADD COLUMN deduct_from_driver boolean;
  END IF;
END $$;
```

---

### 🟡 Warning

#### 3. GET endpoint acquires a 60-second write lock — UX regression

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs`  
**Function:** `getPaymentRefundRequest`

```haskell
eResult <- Redis.whenWithLockRedisAndReturnValue (refundRequestProccessingKey orderId) 60 $ do
  ...
Left () -> throwError (InvalidRequest "Order refund locked")
```

This is a read-only endpoint (customer viewing refund status). It now returns a hard error if an admin is processing the refund (holding the lock for up to 60 s). The old implementation did a plain DB read. Remove the lock from this GET handler entirely — there is no mutation to protect.

---

#### 4. `validateRefundComponents` silently skips validation for legacy rides

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs`  
**Function:** `validateRefundComponents`

```haskell
whenJust mbLi $ \li ->
  forM_ comps $ \c -> do      -- skipped entirely when mbLi = Nothing
    let cap = componentTotal li c.component
    ...
```

If `buildLedgerInfoFromBreakups` returns `Nothing` (a legacy ride with no structured breakup), all component cap checks are silently bypassed. A customer on a legacy ride can request a refund exceeding their payment.

**Fix:** Either throw explicitly when `mbLi = Nothing` with components present, or apply a fallback cap of `transactionAmount` total:
```haskell
li <- mbLi & fromMaybeM (InvalidRequest "Structured fare breakup required for component-wise refund")
forM_ comps $ \c -> do ...
```

---

#### 5. N+1 queries in `validateRefundComponents`

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs`

```haskell
forM_ comps $ \c -> do
  already <- RidePaymentFinance.getSettledRefundByComponent rideId.getId c.component
  -- getSettledRefundByComponent calls getRefundLegEntries (6 queries) on each iteration
```

With 3 components, this issues 3 × 6 = 18 DB round-trips. Fetch `getRefundLegEntries` once before the loop and pass the result in:

```haskell
entries <- RidePaymentFinance.getRefundLegEntries rideId.getId
forM_ comps $ \c -> do
  let (baseRef, vatRef) = refundRefTypesForComponent c.component
      already = sum [e.amount | e <- entries, e.status == LE.SETTLED, e.referenceType `elem` [baseRef, vatRef]]
  when (already + c.amount.amount > componentTotal li c.component) $ throwError ...
```

---

#### 6. `deductFromDriver` defaults to `True` (clawback) on BPP

**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/RefundLedger.hs`  
**Line:** `~431`

```haskell
let deductFromDriver = fromMaybe True req.deductFromDriver -- default: clawback
```

Silently defaulting to `True` (clawback from driver) is a money-sensitive default. If the BAP sends `deductFromDriver = Nothing` in any edge case (e.g. a serialization mismatch), the BPP will punish the driver without an explicit instruction. This should default to `False` (platform absorbs) to fail safe, with the BAP being explicit about clawback intent.

---

#### 7. Multiple OPEN refund requests no longer blocked

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs`  
**Function:** `createPaymentRefundRequest`

Old code: blocked if **any** existing request exists.  
New code: only blocks if an `APPROVED` request exists.

```haskell
when (any (\r -> r.status == DRefundRequest.APPROVED) existingRequests) $
  throwError ...
```

A customer can now submit multiple `OPEN` requests for the same order. If two admins approve different OPEN requests simultaneously (the Redis lock is per-order, so the second approve will catch the first APPROVED — but only inside the lock), there's a timing window. More importantly, the intent should be documented: is multiple partial-refund via multiple OPEN rows the intended model?

---

### 🔵 Suggestion

#### 8. Backward-incompatible API changes need a migration note

Three endpoints break existing clients without a version bump:

| Endpoint | Change |
|---|---|
| `GET /payment/{rideId}/refundRequest` | Response: `RefundRequestResp` → `RefundRequestListResp` |
| `POST /payment/refundRequest/create` | Body: `requestedAmount` removed, `refundComponents` required |
| `GET/POST /payment/refundRequest/{orderId}/…` | `orderId` → `refundRequestId` (dashboard) |

These are intentional but will silently break existing mobile clients. Consider a versioned endpoint (`/v2/`) or at minimum a client-compatibility note in the PR.

---

### ✅ What's good

- The `RefundRequestCreateHandler` / `processRefundResult` abstraction cleanly unifies three entry points (customer create, admin respond, admin initiate). Good factoring.
- Idempotency design in both BAP (`refundSucceededAlreadyRecorded`) and BPP (`entryMatchesRefundStatus` dedup) is solid.
- The `getRideIdForOrder` utility with its comment explicitly calling out the GUID cast anti-pattern is a good guard.
- DDL migrations are guarded (`IF EXISTS`) — safe for re-run.
- Integration test collection (00–07) covering the full flow from fixture ride to invoice PDF is excellent.
- `processRefundFailed` correctly sends `refundComponents = Nothing` to BPP since FAILED just voids pending legs.

Backend/dev/ddl-migrations/rider-app/1552-finance-ledger-entry-metadata-json-type-fix.sql
khuzema786, line 1: is this needed ? Can we remove this ?

Backend/dev/ddl-migrations/dynamic-offer-driver-app/0841-finance-ledger-entry-metadata-json-type-fix.sql
khuzema786, line 1: is this needed ? Can we remove this ?

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/RidePayment.hs
```diff
@@ -562,28 +599,91 @@ getPaymentRefundRequest ::
...
+    Left () -> do
+      logError $ "Order refund locked: " <> orderId.getId
+      throwError (InvalidRequest "Order refund locked")
+    Right result -> return result
+
+-- | Per-component fare+tax breakdown for a ride — the refundable options feeding
+--   refundRequest/create. Reuses 'buildLedgerInfoFromBreakups' with appFee=0, so RideFare is the
+--   FULL customer-paid fare (commission embedded); 'components = Nothing' for legacy rides
+--   whose breakup has no structured tags.
+getPaymentFareBreakup ::
```
khuzema786, line 628: These are Customer App UI Apis, we need are sending this things to UI as well ?

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Finance/RidePayment.hs
```diff
@@ -912,11 +977,130 @@ createTipLedger ctx tipAmount = do
...
+-- 5. Refund ledger entries
+-- ---------------------------------------------------------------------------
+--
+-- Per refund_request (per-component refTypes RideFareRefund/TollRefund/ParkingRefund + VAT,
+-- ref = rideId, {refundRequestId,status} metadata):
+-- APPROVED → pending OwnerAsset→BuyerAsset legs; REFUNDED → settle them; FAILED → void them.
+-- VOIDED legs are ignored by dedup/guards (so a retry re-raises); "done" = a SETTLED leg.
+
+mkRefundMetadata :: Id DRefundRequest.RefundRequest -> DRefundRequest.RefundRequestStatus -> Aeson.Value
+mkRefundMetadata refundRequestId refundRequestStatus =
```
khuzema786: @ClaudeWeb can you suggest a better approach to make it Typed as this Metadata can differ for different domain scenarios in finance ledger.

khuzema786: Also refund request status let's not have here, that single table can be the source, only it's id we can keep here in metadata

Backend/dev/ddl-migrations/dynamic-offer-driver-app/0840-rename-finance-invoice-paymentOrderId-to-entityReferenceId.sql
TejaswiniT167: Agreed — dropped the rename migrations entirely. The new `entity_reference_id` column already arrives via the generated migration (`migrations-read-only/*/finance_invoice.sql`), and all code is on the new field only, so nothing writes `payment_order_id` after this PR — it stays as a dead nullable column that old pods can still write during rollout.

TejaswiniT167:
Claude Web comments resolution: 

**1. Ledger write failure swallowed before Stripe:** Fixed — the raise-time ledger write now aborts the approve on failure, before the BPP call and the Stripe request.

**2. Missing `refund_request` migration:** The columns are covered by the generated migration included in this PR — `migrations-read-only/rider-app/refund_request.sql`. Simple column adds are emitted automatically from the storage spec; hand-written ddl-migrations are only needed for type changes and renames.

**3. GET holding a 60s lock:** Removed the lock from this GET — it was a leftover from the old implementation, where the same endpoint could refresh refund status from the gateway (a write). The V2 version is a pure list read. The mutating endpoints (respond/create/initiate/webhook resolution, and the admin info endpoint which retains refreshRefunds) all keep the shared per-order lock.

**4. Cap validation skipped for legacy rides:** Fixed — create now rejects component refunds on rides without a structured fare breakup (400; no refund_request row is created), using the same error the approve path already threw. Previously such requests entered OPEN and could only fail later at approve.

**5. Repeated queries in the cap loop:** Fixed — the cap check was re-fetching the ride's full refund-ledger entries once per requested component; it now fetches them once before the loop and computes each component's already-refunded total from that in-memory list. The summing logic was moved into a pure helper shared with the original single-component function, so the calculation has a single definition.

**6. `deductFromDriver` defaulting to True:** Made this merchant-configurable per product decision: `transporter_config.defaultRefundDeductFromDriver` decides who bears a refund when the approval doesn't specify. Resolution order: explicit flag on the approval, else the merchant config, else False (platform absorbs — a driver's wallet is never debited without an explicit instruction or configured policy).

**7. Multiple OPEN requests:** Product decision: only one refund request may be in process per order — creation is now blocked while any OPEN or APPROVED request exists (customer create and admin initiate share the same guard); the error message indicates a request is already in process. Caps are still checked at approve time against settled ledger legs before any money moves. Suite changes: 07 "Second OPEN Create Allowed" is now "Second Create While One In Process Blocked 400" (asserts the 400 + message); 07's list contract asserts exactly one OPEN row; two 06 cap-rejection steps and 07's evidence-rejection step were reordered to run without an OPEN request present, and accepted test requests are now explicitly closed (admin reject) once they've served their purpose — so every rejection test hits its own validation rather than the new guard.

**8. Breaking API changes:** The refundRequest endpoints had no UI callers on any surface (customer apps or dashboards) prior to this PR — they were backend-only, so there are no deployed clients to break or version against.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/RefundLedger.hs
```diff
@@ -0,0 +1,380 @@
...
+  [RefundLedgerComponent] -> -- per-component split (always present; caller guarantees it)
+  Flow ()
+createRefundInvoice ctx booking refundsRequestId comps = do
+  let referenceId = ctx.referenceId -- booking.id
+      lineItems = concatMap mkRefundInvoiceLineItems comps
+      anyVat = any (\c -> c.vatAmount > 0) comps
+  -- Parent = the ride's CUSTOMER invoice (keyed by bookingId); Nothing if none found.
+  mbParentRideInvoice <-
+    listToMaybe . filter (\inv -> inv.issuedToType == CUSTOMER)
+      <$> QFinanceInvoiceExtra.findByReferenceIdWithOptions referenceId (Just BeckInvoice.Ride) [FInvoice.Draft, FInvoice.Issued, FInvoice.Paid] (Just 10) (Just 0)
```
khuzema786: can be one query, issuedToType

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Internal/RefundLedger.hs
```diff
@@ -0,0 +1,380 @@
...
+            then Just InvoiceI.InvoiceLineItem {description = desc, descriptionType = Just dtype, quantity = 1, unitPrice = negate amt, lineTotal = negate amt, isExternalCharge = False, groupId = Just "g-commission-refund", itemType = Just typ}
+            else Nothing
+        commissionRefundLines =
+          catMaybes
+            [ mkNegLine "Commission Refund" InvoiceI.CommissionRefund commissionSliceBase InvoiceI.Fare,
+              mkNegLine "Commission VAT Refund" InvoiceI.CommissionRefundTax commissionSliceVat InvoiceI.Tax
+            ]
+    -- Parent = the per-ride Commission invoice (keyed by bookingId; refund-commissions are keyed
+    -- by refundRequestId, so they don't collide here); Nothing if none found.
+    mbParentCommissionInvoice <-
```
khuzema786, line 305: limit 1

Backend/dev/ddl-migrations/rider-app/1552-finance-ledger-entry-metadata-json-type-fix.sql
TejaswiniT167: entityReferenceType and entityReferenceId

khuzema786 (dismissed):

khuzema786 (dismissed):

PR #15023, opened 2026-06-02
backend/fix: vendor json-logic-hs and fix deepMerge array/null semantics

sidsethu:
## Summary

Fixes `driver_pool_config_dynamic_logic_failure` alert caused by two bugs in `json-logic-hs`'s `deepMerge` function that made dynamic pool config logic silently fail for Pickup-area rides (airports, malls).

### Root cause

The `cat` operator in `json-logic-hs` uses `deepMerge` to fold config overrides into a base `DriverPoolConfig`. Two cases were broken:

**Bug 1 — Array concatenation** (`deepMerge (Array a) (Array b) = Array (a <> b)`)

When a rule used `{"var":"config"}` as an `if` else-branch, `cat` called `deepMerge(baseConfig, baseConfig)`. Any array-typed field was doubled:
- After `ae912982e3` changed `Pickup` from 1 field to 2 fields, `area.contents` became `["uuid", null]` (array) instead of `"uuid"` (string)
- `deepMerge(Array ["uuid",null], Array ["uuid",null])` = `["uuid",null,"uuid",null]` — 4 elements
- `fromJSON` failed: **"parsing Area(Pickup) failed, expected Array of length 2, encountered Array of length 4"**

**Bug 2 — Null propagation** (`deepMerge _ Null = Null`)

Rules with a 2-arg `if` (no explicit else) return `Null` when the condition is false. `deepMerge(baseConfig, Null) = Null` wiped the entire config:
- **"parsing DriverPoolConfig(DriverPoolConfig) failed, expected Object, encountered Null"**

Both errors incremented `system_configs_failed_counter "driver_pool_config_dynamic_logic_failure"` and fell back to the DB config, silently skipping dynamic radius/batching optimisations for affected rides.

**Side effect — `{}` else-branch also broken**

Using `{}` (empty object) as else-branch caused `head []` in `jsonLogicValues`, which eagerly evaluates all `if` arguments. This threw an unhandled Haskell exception returning `INTERNAL_ERROR` with null message from the verify API.

### Fix

Two lines changed in `deepMerge`:

```haskell
-- BEFORE
deepMerge (Array a) (Array b) = Array (a <> b)   -- concatenates
deepMerge _          b        = b                  -- Null destroys base

-- AFTER
deepMerge (Array _) (Array b) = Array b            -- last wins
deepMerge a         Null      = a                  -- Null = no-op
deepMerge _         b        = b
```

With this fix:
- `{"var":"config"}` as else → `deepMerge(Array a)(Array a) = a` — no doubling, Bug 1 fixed
- 2-arg `if` returning `Null` → `deepMerge(base, Null) = base` — no wipe, Bug 2 fixed
- `{}` as else → returns `{}`, `deepMerge(Object base, Object {}) = base` — no exception, safe

The original config rules work correctly without any rule changes needed.

### Changes

- **`Backend/lib/json-logic-hs/`** — vendored from `github:nammayatri/json-logic-hs` with the 2-line fix applied to `app/JsonLogic.hs`
- **`Backend/cabal.project`** — added `lib/json-logic-hs` to packages list (local takes precedence over nix input)
- **`Backend/default.nix`** — removed `json-logic-hs.source = inputs.json-logic-hs` (nix module auto-discovers the local package from `cabal.project`)

### Test plan

- [ ] `cabal build yudhishthira` passes (verified locally — 77/77 modules compiled)
- [ ] Deploy to staging and confirm `system_configs_failed_counter` for `driver_pool_config_dynamic_logic_failure` drops to 0
- [ ] Verify the `/nammaTag/appDynamicLogic/verify` API accepts rules with `{}` or 2-arg `if` else-branches without returning `INTERNAL_ERROR`

### Related

- Alert: `system_configs_failed_counter > 15/min` for `driver_pool_config_dynamic_logic_failure` fired 2026-05-31 16:00 UTC
- Root commit that activated the bug: `ae912982e3` (added `gateId` to `Area.Pickup`, changing `contents` from string to array)
- RCA: `Backend/docs-test/RCA_driver_pool_config_dynamic_logic.md`
- Fixed config rules (15 versions): `Backend/docs-test/fixes/`

## Summary by CodeRabbit

* **Chores**
  * Updated a core dependency to a newer pinned revision (includes a bugfix branch addressing deep-merge/dispatch behavior) to bring stability and correctness improvements.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1777,11 +1777,11 @@
     "json-logic-hs": {
       "flake": false,
       "locked": {
-        "lastModified": 1779290761,
-        "narHash": "sha256-LkUlCIdGoER9wofLSG48d/PAnO/ERVZdDR6RBqBPqck=",
+        "lastModified": 1780415936,
+        "narHash": "sha256-I5UaWmMOdMAjyDv1Tr1wIIi+SwF5FsnrJAoqVIddkgA=",
         "owner": "nammayatri",
         "repo": "json-logic-hs",
-        "rev": "47b310e539e5e0c63b0c1d447b6db4dd4d78f68c",
+        "rev": "90393a4a87dc252f4f245c3c215acd6e4c4abe90",
         "type": "github"
       },
       "original": {


sidsethu:
@ClaudeWeb PR Review

khuzema786:
I'll gather all the context needed to review this PR.Let me try the GitHub MCP tools instead.Now let me read the key files in the workspace for deeper context.Now let me look at the relevant parts of the existing codebase for more context on how json-logic-hs is used.I now have enough context. Let me write the review.

---

## PR #15023 Review — `backend/fix: vendor json-logic-hs and fix deepMerge array/null semantics`

**Verdict: Request changes.** The two core `deepMerge` fixes are correct and the vendoring approach is sound, but there is one latent bug the PR claims to fix that it actually does not, and one orphaned flake input that needs cleanup.

---

### Warning — `head []` on empty-object rules is NOT fixed (Confidence: 90%)

**File:** `Backend/lib/json-logic-hs/app/JsonLogic.hs`, lines 47–50

The PR description states:

> `{}` as else → returns `{}`, `deepMerge(Object base, Object {}) = base` — no exception, safe

This reasoning is incorrect. The crash does not happen inside `deepMerge` — it happens earlier, during eager argument evaluation. The `jsonLogic` dispatcher is:

```haskell
A.Object dict ->
  let operator = fst (head (AKM.toList dict))   -- 💥 crashes if dict is empty
      values   = snd (head (AKM.toList dict))
   in applyOperation' operator values
```

For any logic operator (`if`, `cat`, etc.), `applyOperation'` calls `jsonLogicValues` which eagerly evaluates **all** arguments via `mapM (\v -> jsonLogic v data_)`. So for a rule like:

```json
{"if": [condition, override, {}]}
```

The call chain is:

1. `applyOperation' "if" (Array [cond, override, {}])` — `if` is a logic op → evaluate args
2. `jsonLogicValues (Array [cond, override, {}]) data_` → `mapM (jsonLogic _ data_) [...]`
3. `jsonLogic {} data_` → matches `A.Object empty_dict` → `head (AKM.toList {})` → `head []` → **uncaught Haskell exception → INTERNAL_ERROR**

The `deepMerge` change only matters **after** `if` returns its result and `cat` tries to merge it. It can't help when `{}` is itself an argument being evaluated.

**Fix:**

```haskell
A.Object dict ->
  case AKM.toList dict of
    []                     -> pure A.Null   -- {} as a rule → identity/no-op
    ((operator, values):_) -> applyOperation' operator values
```

The test plan item — *"Verify the `/nammaTag/appDynamicLogic/verify` API accepts rules with `{}` or 2-arg `if` else-branches"* — will pass only if the test never actually passes `{}` as a direct argument to an eagerly-evaluated operator. Please confirm which scenario was tested before marking this done.

---

### Warning — Orphaned flake input (Confidence: 90%)

**File:** `flake.nix` (not modified in this PR)

`default.nix` removes the `json-logic-hs.source = inputs.json-logic-hs;` override, but `flake.nix` still declares the input:

```nix
json-logic-hs.url = "github:nammayatri/json-logic-hs";
json-logic-hs.flake = false;
```

After this PR, nothing consumes `inputs.json-logic-hs`. This means:

- The nix fetcher still downloads the upstream source on every clean build — wasted bandwidth
- `flake.lock` keeps a stale pin to the upstream commit
- Future maintainers will be confused about why this input exists

**Fix:** Remove the two `json-logic-hs.*` lines from `flake.nix` and run `nix flake lock --update-input` (or just delete the `json-logic-hs` node from `flake.lock`) to drop it from the lock file.

---

### Suggestion — `Debug.Trace.traceShowId` is the live production `"log"` implementation (Confidence: 95%)

**File:** `Backend/lib/json-logic-hs/app/JsonLogic.hs`, line 196

```haskell
logValue :: Value -> Value
logValue = traceShowId        -- writes raw JSON to stderr, no log level, no context
```

This is used as the `"log"` operator. Any rule containing `{"log": ...}` will write unfiltered to stderr in production. Now that this code is permanently vendored and owned by NammaYatri, consider replacing it with a structured log call (e.g., `logInfo` / `logDebug` in the surrounding monad), or at minimum add a note that `{"log": ...}` is a debug-only operator.

---

### Suggestion — Missing `aeson-qq` in `build-depends` (Confidence: 85%)

**File:** `Backend/lib/json-logic-hs/json-logic-hs.cabal`

`JsonLogic.hs` imports `Data.Aeson.QQ.Simple` (used in `_poolingTest` and `_test`), but the library's `build-depends` doesn't include `aeson-qq`. This happens to compile in the monorepo because `aeson-qq` is an available transitive dependency, but the cabal metadata is incorrect and the package would fail to build standalone.

Since `_poolingTest` and `_test` are development/smoke-test helpers, consider moving them into a separate `test-suite` or `executable` stanza (removing the `aeson-qq` need from the library entirely).

---

### The core fix is correct ✓

The two-line `deepMerge` change is well-reasoned and addresses the two production bugs precisely:

```haskell
-- Bug 1 fix: array last-wins instead of concat — prevents doubling on deepMerge(config, config)
deepMerge (Array _) (Array b) = Array b

-- Bug 2 fix: Null is a no-op instead of wiping base — 2-arg `if` returning Null is now safe
deepMerge a Null = a
```

The vendoring mechanism (local `cabal.project` entry + removing the `default.nix` source override) is the right pattern for the project's nix + haskell-flake setup.

khuzema786 (approved):

PR #15031, opened 2026-06-02
backend/fix/fix-pkey-violation-for-vendor-fee

dushyant-2002:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Summary

Cancellation penalty share was silently dropped from `VendorFee` in the EOD `CalculateDriverFees` job, so the cancellation amount that should have gone to the vendor stayed with the marketplace at split settlement.

## Root cause

`makeVendorFeeForCancellationPenalty` called `QVF.create` blindly. When ride-end had already written a `VendorFee` row for the same `(driverFeeId, vendorId)`:

- KV `SETEX` overwrote the ride-end row with one holding only the cancellation amount.
- Postgres drainer's `INSERT ... ON CONFLICT DO NOTHING` swallowed the conflict.

Net: `VendorFee.amount` ended up = regular ride share only, missing the cancellation portion.

**Evidence:** `DriverFee 567cbf46...` had `cancellation_penalty_amount = ₹10`, but its `NAMMAYATRI` `VendorFee.amount = ₹8.02` (per-ride share only). Expected ≈ ₹18.02.

## Fix

Replace blind `QVF.create` with an upsert keyed on `(driverFeeId, vendorId)`:

1. Read existing `VendorFee` via `runInMasterDbAndRedis` (master read; the EOD job can race a recent ride-end write).
2. If exists → `QVF.updateAmount` summing onto existing amount.
3. If not → `QVF.create` a fresh row.

## Scope

- MANUAL parent flow only. AUTOPAY uses a freshly-generated child `driverFeeId`, so no collision is possible.
- Ride-end hot path untouched.

## Files

- `SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs` — rewrite `makeVendorFeeForCancellationPenalty`.
- `Storage/Queries/VendorFeeExtra.hs` — add `updateAmount :: Id DriverFee -> Text -> HighPrecMoney -> m ()`.

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Cancellation-penalty handling corrected so fees now accumulate per driver and fee instance.
  * Added an idempotency guard to prevent duplicate charges on retries (temporary guard with expiry).
  * Existing vendor-fee records are reliably updated (amounts and timestamps) instead of creating duplicates.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
@@ -44,6 +44,7 @@ import Domain.Types.Plan (BasedOnEntity (..), PaymentMode (AUTOPAY, MANUAL), Pla
 import Domain.Types.SubscriptionConfig
 import Domain.Types.TransporterConfig (TransporterConfig)
 import qualified Domain.Types.VendorFee as DVF
+import Kernel.Beam.Functions (runInMasterDbAndRedis)
 import Kernel.External.Encryption
 import qualified Kernel.External.Payment.Interface as PaymentInterface
 import qualified Kernel.External.Payment.Interface.Types as Payment
@@ -935,15 +936,27 @@ makeVendorFeeForCancellationPenalty ::
   TransporterConfig ->
   m ()
 makeVendorFeeForCancellationPenalty driverFee subscriptionConfig transporterConfig = when (fromMaybe False subscriptionConfig.isVendorSplitEnabled && isJust transporterConfig.cancellationFeeVendor && fromMaybe 0 driverFee.cancellationPenaltyAmount > 0) $ do
-  let vendorFee =
-        DVF.VendorFee
-          { driverFeeId = driverFee.id,
-            vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor,
-            amount = fromMaybe 0 driverFee.cancellationPenaltyAmount,
-            createdAt = driverFee.createdAt,
-            updatedAt = driverFee.updatedAt
-          }
-  QVF.create vendorFee
+  let vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor
+      cancellationAmount = fromMaybe 0 driverFee.cancellationPenaltyAmount
+      cancellationVendorFeeGuardKey = "VendorFee:CancellationApplied:" <> driverFee.id.getId
+  -- Guard against scheduler retries double-adding. Set after write so a crash re-applies, not skips.
+  alreadyApplied <- Hedis.get cancellationVendorFeeGuardKey
+  case (alreadyApplied :: Maybe Bool) of
+    Just True -> logError $ "makeVendorFeeForCancellationPenalty: cancellation already applied for driverFee " <> driverFee.id.getId <> ", skipping"
+    _ -> do
+      mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId vendorId driverFee.id
+      case mbExisting of
+        Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
+        Nothing ->
+          QVF.create
+            DVF.VendorFee
+              { driverFeeId = driverFee.id,
+                vendorId = vendorId,
+                amount = cancellationAmount,
+                createdAt = driverFee.createdAt,
+                updatedAt = driverFee.updatedAt
+              }
+      Hedis.setExp cancellationVendorFeeGuardKey True (3600 * 24)
 
 updateCancellationPenaltyAccumulationFees :: (MonadFlow m, CacheFlow m r, EsqDBFlow m r, EncFlow m r, HasKafkaProducer r) => ServiceNames -> TransporterConfig -> Id Merchant -> Id MerchantOperatingCity -> m ()
 updateCancellationPenaltyAccumulationFees serviceName transporterConfig merchantId merchantOperatingCityId = do
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/VendorFeeExtra.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/VendorFeeExtra.hs
@@ -172,3 +172,16 @@ updateVendorId driverFeeId oldVendorId newVendorId = do
           Se.Is Beam.vendorId $ Se.Eq oldVendorId
         ]
     ]
+
+updateAmount :: (EsqDBFlow m r, MonadFlow m, CacheFlow m r) => Id DriverFee -> Text -> HighPrecMoney -> m ()
+updateAmount driverFeeId vendorId newAmount = do
+  now <- getCurrentTime
+  updateWithKV
+    [ Se.Set Beam.amount (roundToTwoDecimalPlaces newAmount),
+      Se.Set Beam.updatedAt now
+    ]
+    [ Se.And
+        [ Se.Is Beam.driverFeeId $ Se.Eq driverFeeId.getId,
+          Se.Is Beam.vendorId $ Se.Eq vendorId
+        ]
+    ]


dushyant-2002:

khuzema786:
I'll gather all the context for PR #15031 before reviewing.Let me try using the MCP GitHub tools instead.Now let me fetch the changed files and look at the relevant source code.Now let me read the full source files for deeper context.Files aren't in the local workspace. Let me fetch them directly via GitHub.Now let me look at the `findByVendorAndDriverFeeId` function and the broader calling context of `makeVendorFeeForCancellationPenalty`.I have everything I need to write the review. Here it is:

---

## PR Review: `backend/fix/fix-pkey-violation-for-vendor-fee`

**Verdict:** The fix correctly identifies and addresses the root cause. Two issues worth discussing before merge — one real concern around idempotency, one residual race — plus a minor observation. The overall approach is sound.

---

### Warning

#### 1. Non-idempotent on EOD job retry — double-accumulation of `cancellationAmount` (Confidence: 82%)

**File:** `DriverFee.hs`, `makeVendorFeeForCancellationPenalty`

The `Just existing` branch computes the new total at the **call site** by reading the current stored amount and adding `cancellationAmount`:

```haskell
Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
```

If the `CalculateDriverFees` EOD job is retried (e.g., a scheduler timeout, a transient DB error that doesn't mark the job as failed), this function runs again for the same `driverFee`. The `findByVendorAndDriverFeeId` now returns the row that already contains `existing.amount + cancellationAmount` from the first run, and the update writes `existing.amount + 2 * cancellationAmount`. **Each retry doubles the cancellation portion.**

The original blind `QVF.create` was paradoxically idempotent in the retry case (the `INSERT ... ON CONFLICT DO NOTHING` drainer meant subsequent runs were no-ops). This fix trades one bug for another under retry.

**Suggested mitigations (pick one):**

Option A — Store the expected amount and compare before updating:
```haskell
Just existing ->
  when (existing.amount < existingRideShareAmount + cancellationAmount) $
    QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
```
…though this requires knowing `existingRideShareAmount`, which complicates the call.

Option B — Use a DB-side atomic increment (`SET amount = amount + ?`) instead of a read-then-write, and track whether the cancellation was already applied on the `DriverFee` row itself (e.g., a `cancellationVendorFeeApplied :: Bool` flag). This is the cleanest if you want true idempotency.

Option C — Document explicitly that the parent job guarantees single-execution per `driverFeeId` (e.g., status transitions prevent reprocessing). If that's the case, a code comment here would close the review concern.

---

#### 2. TOCTOU race between `findByVendorAndDriverFeeId` and `QVF.create` (Confidence: 62%)

**File:** `DriverFee.hs`

The `runInMasterDbAndRedis` correctly prevents reading stale replica data, but there's still a window between the master read returning `Nothing` and `QVF.create` executing. If ride-end writes a `VendorFee` row in that window, `QVF.create` will hit the same KV-overwrite / PK-conflict that the PR is trying to fix.

The window is narrow and the EOD job runs off-peak, so this is a **potential issue** rather than a blocker. A DB-level upsert (`INSERT ... ON CONFLICT (driver_fee_id, vendor_id) DO UPDATE SET amount = amount + excluded.amount`) would eliminate it entirely, but that's a larger change. Worth at least calling out in a comment near the `runInMasterDbAndRedis` line.

---

### Suggestions

#### 3. `updateAmount` duplicates logic already in `updateVendorFee`

**File:** `VendorFeeExtra.hs`

The existing `updateVendorFee` already does a find-then-update-or-create. The new `updateAmount` is a lower-level version that skips the `getLocalCurrentTime` lookup (uses `getCurrentTime` instead) and requires the caller to pre-compute the total. Consider whether the existing function can be reused by passing a `merchantOpCityId` through, or at least share the `updateWithKV` block.

#### 4. `fromMaybe "CANCELLATION_PENALTY_VENDOR"` fallback is unreachable

**File:** `DriverFee.hs`, line ~940

```haskell
when (... && isJust transporterConfig.cancellationFeeVendor ...) $ do
  let vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor
```

The `when` guard already asserts the field is `Just`, so the default string is dead code. Not a bug, but `fromJust` (or `fromMaybe` with a clearer intent signal) would make this accurate. Pre-existing, but worth cleaning up since this line was touched.

---

### What's Good

- **`runInMasterDbAndRedis`** is the right call here — it correctly forces a master read to avoid the KV/replica staleness that caused the original overwrite.
- **Scope is correct** — AUTOPAY is explicitly excluded since it uses freshly-generated child `driverFeeId`s with no collision risk.
- **`roundToTwoDecimalPlaces` applied inside `updateAmount`** — rounding happens after addition, which is correct.
- The root-cause analysis in the PR description is clear and accurate.

dushyant-2002:
@ClaudeWeb review now

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
```diff
@@ -935,15 +936,20 @@ makeVendorFeeForCancellationPenalty ::
...
 makeVendorFeeForCancellationPenalty driverFee subscriptionConfig transporterConfig = when (fromMaybe False subscriptionConfig.isVendorSplitEnabled && isJust transporterConfig.cancellationFeeVendor && fromMaybe 0 driverFee.cancellationPenaltyAmount > 0) $ do
-  let vendorFee =
+  let vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor
+      cancellationAmount = fromMaybe 0 driverFee.cancellationPenaltyAmount
+  mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId vendorId driverFee.id
+  case mbExisting of
+    Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
+    Nothing ->
+      QVF.create
         DVF.VendorFee
```
khuzema786: **Non-idempotent on retry.** If the EOD job is retried for the same `driverFee`, this branch reads the already-updated amount and adds `cancellationAmount` again, resulting in double-counting.

Either confirm the parent job's status-transition logic prevents re-entering this function for the same fee (and add a comment), or guard explicitly:
```haskell
-- Only accumulate if cancellation hasn't already been applied to this vendor fee.
-- Check that existing.amount equals only the ride-share portion before adding.
Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
```
Consider tracking `cancellationVendorFeeApplied` on the `DriverFee` row itself to make this idempotent regardless of retry.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/VendorFeeExtra.hs
```diff
@@ -172,3 +172,16 @@ updateVendorId driverFeeId oldVendorId newVendorId = do
...
+
+updateAmount :: (EsqDBFlow m r, MonadFlow m, CacheFlow m r) => Id DriverFee -> Text -> HighPrecMoney -> m ()
+updateAmount driverFeeId vendorId newAmount = do
+  now <- getCurrentTime
+  updateWithKV
+    [ Se.Set Beam.amount (roundToTwoDecimalPlaces newAmount),
+      Se.Set Beam.updatedAt now
+    ]
+    [ Se.And
+        [ Se.Is Beam.driverFeeId $ Se.Eq driverFeeId.getId,
```
khuzema786, line 184: `updateAmount` takes a pre-computed total and overwrites the DB value. The name suggests it _updates_ the amount, but the semantics are _set_ (caller is responsible for adding the delta before calling). Consider naming it `setAmount` for clarity, or accept the delta here and do the addition inside to make the function harder to misuse.

khuzema786 (changes requested):
## Review: `backend/fix/fix-pkey-violation-for-vendor-fee`

The fix correctly identifies and addresses the root cause — the blind `QVF.create` was silently losing the cancellation portion of the vendor fee. The use of `runInMasterDbAndRedis` and the read-then-upsert approach are the right instincts. However there is one real concern around idempotency that should be resolved before merge, plus a narrower residual race worth acknowledging.

---

### ⚠️ Warning — Non-idempotent on EOD job retry: double-accumulation of `cancellationAmount` (Confidence: 82%)

**File:** `DriverFee.hs` — `makeVendorFeeForCancellationPenalty`

The `Just existing` branch computes the new total at the **call site**:

```haskell
Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
```

If the `CalculateDriverFees` EOD job is retried (scheduler timeout, transient DB error that doesn't mark the job as failed, etc.), this function runs again for the same `driverFee`. The lookup now returns the row that already contains `existing.amount + cancellationAmount` from the first run, and the update writes `existing.amount + 2 × cancellationAmount`. **Each retry double-counts the cancellation portion.**

The original blind `QVF.create` was paradoxically idempotent on retry (`INSERT ... ON CONFLICT DO NOTHING` was a no-op on the second run). This fix trades one correctness bug for another under retry.

**Suggested mitigations — pick one:**

**Option A (simplest):** Add a comment + confirm the parent job gates on `DriverFee` status transitions that prevent reprocessing the same fee. If that guarantee exists, a code comment here closes the concern.

**Option B (robust):** Track whether the cancellation vendor fee was already applied on the `DriverFee` row itself (e.g., a `cancellationVendorFeeApplied :: Bool` field), and guard on it:

```haskell
unless driverFee.cancellationVendorFeeApplied $ do
  mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId vendorId driverFee.id
  ...
  QDF.markCancellationVendorFeeApplied driverFee.id  -- set after successful write
```

**Option C (strongest):** Use a DB-level atomic increment instead of read-then-write, so the operation is naturally idempotent regardless of retries.

---

### ⚠️ Potential Issue — TOCTOU race between `findByVendorAndDriverFeeId` and `QVF.create` (Confidence: 62%)

`runInMasterDbAndRedis` correctly prevents reading stale replica data, but there is still a window between the master read returning `Nothing` and `QVF.create` executing. If ride-end writes a `VendorFee` row in that window, `QVF.create` hits the same KV-overwrite / PK-conflict the PR is fixing. The window is narrow and the EOD job runs off-peak, so this is not a blocker, but worth a comment near the `runInMasterDbAndRedis` line to explain the residual risk and why it is acceptable. A DB-level `INSERT ... ON CONFLICT (driver_fee_id, vendor_id) DO UPDATE SET amount = amount + excluded.amount` would eliminate it entirely.

---

### 💡 Suggestion — `updateAmount` overlaps with existing `updateVendorFee`

**File:** `VendorFeeExtra.hs`

The existing `updateVendorFee` already implements find-then-update-or-create. The new `updateAmount` is a leaner version (skips `getLocalCurrentTime`, caller pre-computes the total). The main difference is the caller here wraps the find in `runInMasterDbAndRedis` while `updateVendorFee` does not. Consider whether `updateVendorFee` can be extended to accept a master-read flag, or at least add a comment on `updateAmount` explaining why a new function was needed over the existing one.

---

### 💡 Minor — `fromMaybe "CANCELLATION_PENALTY_VENDOR"` fallback is dead code

**File:** `DriverFee.hs`

```haskell
when (... && isJust transporterConfig.cancellationFeeVendor ...) $ do
  let vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor
```

The `when` guard already asserts `isJust`, so the default string is unreachable. Pre-existing behaviour, but since this line was touched, `fromJust` (with a matching `fromJust` error) or a pattern match would make the intent clearer.

---

### ✅ What's good

- `runInMasterDbAndRedis` is the correct call — it forces a master read to avoid KV/replica staleness that caused the original overwrite.
- Scope is correctly limited to the MANUAL flow. AUTOPAY exclusion is well-reasoned (freshly-generated child `driverFeeId` → no collision).
- `roundToTwoDecimalPlaces` is applied inside `updateAmount` after addition — correct ordering.
- Root-cause analysis in the PR description is precise and actionable.

The idempotency question (Finding 1) is the one that needs a concrete answer before this lands.

khuzema786:
Review posted successfully to [PR #15031](https://github.com/nammayatri/nammayatri/pull/15031#pullrequestreview-4410408235) with **Request Changes**.

**Summary of what was flagged:**

1. **Warning (82%)** — Non-idempotent on EOD job retry: the `Just existing` branch accumulates `cancellationAmount` onto whatever is currently stored, so any job retry double-counts it. The author needs to either confirm the parent job's status-gating prevents reprocessing (and add a comment), or make the operation explicitly idempotent.

2. **Potential Issue (62%)** — Residual TOCTOU race between the master read and `QVF.create` in the `Nothing` branch. Narrow window but not zero; worth a comment or a DB-level upsert if stricter correctness is needed.

3. **Suggestion** — `updateAmount` duplicates logic already in `updateVendorFee`; the naming also implies "delta" but it takes a total. Inline comment flagged both points.

4. **Minor** — The `fromMaybe "CANCELLATION_PENALTY_VENDOR"` fallback is unreachable given the surrounding `isJust` guard.

dushyant-2002:
@ClaudeWeb i have added this to prevent double addition 
let cancellationVendorFeeGuardKey = "VendorFee:CancellationApplied:" <> driverFee.id.getId
  acquired <- Hedis.setNxExpire cancellationVendorFeeGuardKey (3600 * 6) True

khuzema786:
Good direction — Redis `SETNX` as an idempotency guard is a standard pattern here. Two concrete concerns before this lands:

---

### 1. Set the key *after* the DB write, not before (important)

The current flow is likely:
```haskell
acquired <- Hedis.setNxExpire cancellationVendorFeeGuardKey (3600 * 6) True
when acquired $ do
  mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId ...
  -- DB write happens here
```

If the **DB write throws** (network hiccup, drainer error, etc.) after `setNxExpire` has already succeeded, the key is set but the fee was never applied. The next retry sees `acquired = False` and silently skips — the cancellation vendor fee is **permanently lost**, which is the exact bug you're fixing in a different form.

Safer pattern — set the guard **on success**:

```haskell
mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId vendorId driverFee.id
case mbExisting of
  Just existing -> do
    QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
    void $ Hedis.setNxExpire cancellationVendorFeeGuardKey (3600 * 48) True
  Nothing -> do
    QVF.create DVF.VendorFee { ... }
    void $ Hedis.setNxExpire cancellationVendorFeeGuardKey (3600 * 48) True
```

And at the top, check first and short-circuit:
```haskell
alreadyApplied <- isJust <$> Hedis.get cancellationVendorFeeGuardKey
unless alreadyApplied $ do
  -- full upsert logic + set key on success
```

---

### 2. TTL of 6 hours is too short

The EOD job runs daily. A 6-hour TTL means if the job starts at 23:00 and a retry is triggered at 05:30 (scheduler picks it up after a long delay or failed node), the guard has expired and you're vulnerable again. Use **48 hours** (`3600 * 48`) to cover the full billing cycle with margin.

---

### Also — can you share the full updated snippet?

Want to make sure the `acquired` value is actually gating the upsert (i.e., `when acquired $ do ...`). If you can paste the full revised `makeVendorFeeForCancellationPenalty`, I can confirm it's wired correctly before approving.

dushyant-2002:
@ClaudeWeb  review now .
12 hours is enough for redis Key . as job will run mostly for 4 hours only

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
```diff
@@ -935,15 +936,27 @@ makeVendorFeeForCancellationPenalty ::
...
-  QVF.create vendorFee
+  let vendorId = fromMaybe "CANCELLATION_PENALTY_VENDOR" transporterConfig.cancellationFeeVendor
+      cancellationAmount = fromMaybe 0 driverFee.cancellationPenaltyAmount
+      cancellationVendorFeeGuardKey = "VendorFee:CancellationApplied:" <> driverFee.id.getId
+  -- Guard against scheduler retries double-adding. Set after write so a crash re-applies, not skips.
+  alreadyApplied <- Hedis.get cancellationVendorFeeGuardKey
+  case (alreadyApplied :: Maybe Bool) of
+    Just True -> logError $ "makeVendorFeeForCancellationPenalty: cancellation already applied for driverFee " <> driverFee.id.getId <> ", skipping"
+    _ -> do
+      mbExisting <- runInMasterDbAndRedis $ QVF.findByVendorAndDriverFeeId vendorId driverFee.id
```
khuzema786, line 947: `logError` here will fire production alerts / on-call pages for an expected retry-guard skip. This is normal, handled behaviour — change to `logInfo`:
```haskell
Just True -> logInfo $ "makeVendorFeeForCancellationPenalty: cancellation already applied for driverFee " <> driverFee.id.getId <> ", skipping (retry guard)"
```

khuzema786 (dismissed):
## Review (updated) — Approving

The idempotency concern from the previous review is correctly addressed. The guard key is read before the write and **set only after** the DB write succeeds — so a crash between the two means the next retry correctly re-applies rather than silently skipping. Good.

---

### One nit before merge: `logError` should be `logInfo`

**File:** `DriverFee.hs`, `Just True ->` branch

```haskell
Just True -> logError $ "makeVendorFeeForCancellationPenalty: cancellation already applied for driverFee " <> ...
```

A retry being detected and correctly skipped is **expected, handled behaviour** — not an error. Logging it as `logError` will fire alerts or trigger unnecessary on-call investigations in production. Use `logInfo` (or `logDebug` if this is too noisy):

```haskell
Just True -> logInfo $ "makeVendorFeeForCancellationPenalty: cancellation already applied for driverFee " <> driverFee.id.getId <> ", skipping (retry guard)"
```

Reserve `logError` for states that require human action. A guarded skip is not one of them.

---

### TTL reasoning — accepted

12-hour TTL is fine given the job's max 4-hour runtime. The guard only needs to outlive any retry window within the same job run. Next day's run will process different `driverFeeId`s (distinct per billing cycle), so expiry by then is correct.

---

### What's solid in this revision

- Guard is set **post-write** — crash-safe in the right direction (re-apply on crash, not skip).
- `_ ->` catch-all correctly handles both `Nothing` and any unexpected stored value.
- `runInMasterDbAndRedis` + the guard together cover both the cross-replica staleness and the retry double-count problems.
- Scope remains correctly limited to MANUAL flow.

LGTM modulo the `logError` → `logInfo` change.

khuzema786:
I have the full updated diff. The implementation correctly addresses my prior concerns — guard is read first, the DB write happens, and `setExp` is called only after success. Let me post the review.Review approved and posted to [PR #15031](https://github.com/nammayatri/nammayatri/pull/15031#pullrequestreview-4410556361).

**What changed between the two reviews:**

| Concern | Status |
|---|---|
| Idempotency / double-accumulation | ✅ Fixed — guard key set post-write |
| Guard key set before write (crash risk) | ✅ Fixed — `Hedis.setExp` called only after DB write succeeds |
| TTL too short | ✅ Accepted — 12h covers the 4h job window |

**One remaining nit (non-blocking):** `logError` on the retry-guard skip path should be `logInfo` — a correctly skipped retry is expected behavior, not an error, and will generate false alerts in production.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
```diff
@@ -935,15 +936,27 @@ makeVendorFeeForCancellationPenalty ::
...
+      case mbExisting of
+        Just existing -> QVF.updateAmount driverFee.id vendorId (existing.amount + cancellationAmount)
```
0utkarsh, line 949: Sequential write issue in case of crash can still be there as this is not idempotent

0utkarsh: Also, concurrent writes issues as well. What if a ride also ends at this time? That will also try to update the amount, can we handle the race condition?

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
dushyant-2002: there's a lock in start of job to prevent concurrent execution .

0utkarsh (approved):

PR #15056, opened 2026-06-04
backend/feat: add driver_identity_info table for nominee+address (backward-compatible fallback)

bytedex:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Drivers can add/update nominee DOB and address state; profile responses surface nominee DOB and computed nominee age.
  * Nominee/address data moved to a dedicated identity cohort with read/merge and upsert semantics; insurance and profile reads prefer cohort data.

* **Bug Fixes / Validation**
  * Validation for nominee relationship and age (18–99).
  * Legacy nominee/address fields marked deprecated.

* **Tests**
  * End-to-end integration tests for the nominee/profile flow added.

* **Chores**
  * New DB table and migration for driver identity cohort.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/DriverIdentityInfo.yaml
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/DriverIdentityInfo.yaml
@@ -1,0 +1,33 @@
+imports:
+  Merchant: Domain.Types.Merchant
+  MerchantOperatingCity: Domain.Types.MerchantOperatingCity
+  Person: Domain.Types.Person
+  AddressDocumentType: Domain.Types.DriverInformation
+  Day: Data.Time
+  IndianState: Kernel.Types.Beckn.Context
+
+DriverIdentityInfo:
+  tableName: driver_identity_info
+
+  fields:
+    driverId: Id Person
+    nomineeName: Maybe Text
+    nomineeRelationship: Maybe Text
+    nomineeDob: Maybe Day
+    address: Maybe Text
+    addressDocumentType: Maybe AddressDocumentType
+    addressState: Maybe IndianState
+    merchantId: Id Merchant
+    merchantOperatingCityId: Id MerchantOperatingCity
+    createdAt: UTCTime
+    updatedAt: UTCTime
+
+  constraints:
+    driverId: PrimaryKey
+
+  queries:
+    findByDriverId:
+      kvFunction: findOneWithKV
+      where: driverId
+
+  excludedFields: [merchantId, merchantOperatingCityId, createdAt, updatedAt]
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
@@ -192,6 +192,7 @@ import SharedLogic.AnalyticsExtra as AnalyticsExtra
 import qualified SharedLogic.Booking as SBooking
 import qualified SharedLogic.DriverFleetOperatorAssociation as SA
 import qualified SharedLogic.DriverFlowStatus as SDF
+import qualified SharedLogic.DriverIdentityInfo as DIInfo
 import SharedLogic.DriverOnboarding
 import qualified SharedLogic.DriverOnboarding.Status as SStatus
 import qualified SharedLogic.External.LocationTrackingService.Flow as LF
@@ -228,6 +229,7 @@ import qualified Storage.Queries.AlertRequest as QAR
 import qualified Storage.Queries.Booking as QRB
 import qualified Storage.Queries.DriverBankAccount as QDBA
 import qualified Storage.Queries.DriverGstinExtra as QDGExtra
+import qualified Storage.Queries.DriverIdentityInfo as QDII
 import qualified Storage.Queries.DriverInformation as QDriverInfo
 import qualified Storage.Queries.DriverLicense as QDriverLicense
 import qualified Storage.Queries.DriverOperatorAssociation as DOV
@@ -4530,9 +4532,10 @@ postDriverFleetDriverUpdate merchantShortId opCity driverId requestorId req = do
         DP.DRIVER -> do
           driverInfo <- QDriverInfo.findById personId >>= fromMaybeM DriverInfoNotFound
           let dob = fmap (\d -> UTCTime d 0) req.dob <|> driverInfo.driverDob
-              address = req.address <|> driverInfo.address
-              addressDocumentType = castAddressDocumentType <$> req.addressDocumentType <|> driverInfo.addressDocumentType
-          QDriverInfo.updateDriverDobAndAddress dob address addressDocumentType personId
+          Redis.withLockRedis (DIInfo.driverIdentityInfoLockKey personId) 10 $ do
+            mbExisting <- QDII.findByDriverId personId
+            void $ DIInfo.upsertDriverIdentityInfo mbExisting personId driver.merchantId driver.merchantOperatingCityId driverInfo Nothing Nothing Nothing req.address (castAddressDocumentType <$> req.addressDocumentType) Nothing
+          QDriverInfo.updateDriverDobAndAddress dob Nothing Nothing personId
         role | DCommon.checkFleetOwnerRole role -> do
           fleetOwnerInfo <- B.runInReplica (FOI.findByPrimaryKey personId) >>= fromMaybeM (InvalidRequest "Fleet owner information does not exist")
           reqStripeIdNumber <- forM req.stripeIdNumber encrypt
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
@@ -257,6 +257,7 @@ import SharedLogic.CallBAP (sendDriverOffer, sendRideAssignedUpdateToBAP)
 import qualified SharedLogic.CallInternalMLPricing as ML
 import qualified SharedLogic.DeleteDriver as DeleteDriverOnCheck
 import qualified SharedLogic.DriverFee as SLDriverFee
+import qualified SharedLogic.DriverIdentityInfo as DIInfo
 import SharedLogic.DriverOnboarding
 import SharedLogic.DriverPool as DP
 import qualified SharedLogic.EventTracking as ET
@@ -300,6 +301,7 @@ import qualified Storage.Queries.DriverFee as QDF
 import qualified Storage.Queries.DriverGoHomeRequest as QDGR
 import qualified Storage.Queries.DriverGstinExtra as QDGExtra
 import qualified Storage.Queries.DriverHomeLocation as QDHL
+import qualified Storage.Queries.DriverIdentityInfo as QDII
 import qualified Storage.Queries.DriverInformation as QDriverInformation
 import qualified Storage.Queries.DriverOperatorAssociationExtra as QDOA
 import qualified Storage.Queries.DriverPanCard as QPanCard
@@ -476,7 +478,8 @@ data DriverInformationRes = DriverInformationRes
     membershipId :: Maybe (Id DStclMembership.StclMembership),
     createdAt :: UTCTime,
     role :: SP.Role,
-    operatorBadgeToken :: Maybe Text
+    operatorBadgeToken :: Maybe Text,
+    nomineeDob :: Maybe Day
   }
   deriving (Generic, ToJSON, FromJSON, ToSchema)
 
@@ -564,7 +567,8 @@ data DriverEntityRes = DriverEntityRes
     vehicleImageUploadedAt :: Maybe UTCTime,
     subscriptionCreditBalance :: Maybe HighPrecMoney,
     role :: SP.Role,
-    operatorBadgeToken :: Maybe Text
+    operatorBadgeToken :: Maybe Text,
+    nomineeDob :: Maybe Day
   }
   deriving (Show, Generic, FromJSON, ToJSON, ToSchema)
 
@@ -601,7 +605,9 @@ data UpdateDriverReq = UpdateDriverReq
     address :: Maybe Text,
     addressDocumentType :: Maybe DriverInfo.AddressDocumentType,
     nomineeName :: Maybe Text,
-    nomineeRelationship :: Maybe Text
+    nomineeRelationship :: Maybe Text,
+    nomineeDob :: Maybe Day,
+    addressState :: Maybe Context.IndianState
   }
   deriving (Generic, ToJSON, FromJSON, ToSchema, Show)
 
@@ -906,7 +912,8 @@ getInformation (personId, merchantId, merchantOpCityId) mbClientId toss tnant' c
   merchant <-
     CQM.findById merchantId
       >>= fromMaybeM (MerchantNotFound merchantId.getId)
-  driverEntity <- buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId) merchant serviceName
+  mbIdentity <- QDII.findByDriverId person.id
+  driverEntity <- buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId, DIInfo.getIdentityInfo mbIdentity driverInfo) merchant serviceName
   dues <- QDF.findAllFeeByTypeServiceStatusAndDriver serviceName driverId [DDF.RECURRING_INVOICE, DDF.RECURRING_EXECUTION_INVOICE] [DDF.PAYMENT_PENDING, DDF.PAYMENT_OVERDUE]
   let currentDues = sum $ map (\dueInvoice -> SLDriverFee.roundToHalf dueInvoice.currency (dueInvoice.govtCharges + dueInvoice.platformFee.fee + dueInvoice.platformFee.cgst + dueInvoice.platformFee.sgst)) dues
   let manualDues = sum $ map (\dueInvoice -> SLDriverFee.roundToHalf dueInvoice.currency (dueInvoice.govtCharges + dueInvoice.platformFee.fee + dueInvoice.platformFee.cgst + dueInvoice.platformFee.sgst)) $ filter (\due -> due.status == DDF.PAYMENT_OVERDUE) dues
@@ -1150,11 +1157,11 @@ buildDriverEntityRes ::
     Redis.HedisLTSFlowEnv r,
     EsqDBFlow m r
   ) =>
-  (SP.Person, DriverInformation, DStats.DriverStats, Id DMOC.MerchantOperatingCity) ->
+  (SP.Person, DriverInformation, DStats.DriverStats, Id DMOC.MerchantOperatingCity, DIInfo.IdentityInfo) ->
   DM.Merchant ->
   Plan.ServiceNames ->
   m DriverEntityRes
-buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId) merchant serviceName = do
+buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId, identityInfo) merchant serviceName = do
   transporterConfig <- getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = person.merchantOperatingCityId.getId}) (Just (SCTC.findByMerchantOpCityId person.merchantOperatingCityId Nothing)) >>= fromMaybeM (TransporterConfigNotFound person.merchantOperatingCityId.getId)
   vehicleMB <- QVehicle.findById person.id
   DriverSpecificSubscriptionData {mbDriverPlan = driverPlan, ..} <- getDriverSpecificSubscriptionDataWithSubsConfig (person.id, transporterConfig.merchantId, merchantOpCityId) transporterConfig driverInfo vehicleMB serviceName
@@ -1309,8 +1316,9 @@ buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId) merchan
         isTTSEnabled = driverInfo.isTTSEnabled,
         isHighAccuracyLocationEnabled = driverInfo.isHighAccuracyLocationEnabled,
         rideRequestVolumeEnabled = driverInfo.rideRequestVolumeEnabled,
-        nomineeName = driverInfo.nomineeName,
-        nomineeRelationship = driverInfo.nomineeRelationship,
+        nomineeName = identityInfo.nomineeName,
+        nomineeRelationship = identityInfo.nomineeRelationship,
+        nomineeDob = identityInfo.nomineeDob,
         subscriptionCreditBalance = subsCreditBalance,
         role = person.role,
         operatorBadgeToken = if person.role `elem` [SP.BUS_CONDUCTOR, SP.BUS_DRIVER] then person.operatorBadgeToken else Nothing,
@@ -1403,10 +1411,6 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
       isHighAccuracyLocationEnabled = req.isHighAccuracyLocationEnabled <|> driverInfo.isHighAccuracyLocationEnabled
       rideRequestVolumeEnabled = req.rideRequestVolumeEnabled <|> driverInfo.rideRequestVolumeEnabled
       onboardingAs = req.onboardingAs <|> driverInfo.onboardingAs
-      address = req.address <|> driverInfo.address
-      addressDocumentType = req.addressDocumentType <|> driverInfo.addressDocumentType
-      nomineeName = req.nomineeName <|> driverInfo.nomineeName
-      nomineeRelationship = req.nomineeRelationship <|> driverInfo.nomineeRelationship
   -- Compute vehicle-related fields (use existing values if no vehicle or no request)
   (canDowngradeToSedan, canDowngradeToHatchback, canDowngradeToTaxi, canSwitchToRental, canSwitchToInterCity, canSwitchToIntraCity, availableUpiApps) <-
     case mVehicle of
@@ -1464,9 +1468,20 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
               DV.AUTO_LITE -> [DVST.AUTO_LITE]
               DV.PINK_AUTO -> [DVST.PINK_AUTO]
       QVehicle.updateSelectedServiceTiers selectedServiceTiers person.id
+  let nomineeOrAddressChanged = isJust req.nomineeName || isJust req.nomineeRelationship || isJust req.nomineeDob || isJust req.address || isJust req.addressDocumentType || isJust req.addressState
+  mbIdentityInfo <-
+    if nomineeOrAddressChanged
+      then Redis.withLockRedisAndReturnValue (DIInfo.driverIdentityInfoLockKey personId) 10 $ do
+        mbExisting <- QDII.findByDriverId personId
+        Just <$> DIInfo.upsertDriverIdentityInfo mbExisting personId person.merchantId merchantOpCityId driverInfo req.nomineeName req.nomineeRelationship req.nomineeDob req.address req.addressDocumentType req.addressState
+      else QDII.findByDriverId personId
+  let (legacyAddress, legacyAddressDocumentType, legacyNomineeName, legacyNomineeRelationship) =
+        if nomineeOrAddressChanged
+          then (Nothing, Nothing, Nothing, Nothing)
+          else (driverInfo.address, driverInfo.addressDocumentType, driverInfo.nomineeName, driverInfo.nomineeRelationship)
   -- Update driver information (works for both cases: with or without vehicle)
   when (isJust req.canDowngradeToSedan || isJust req.canDowngradeToHatchback || isJust req.canDowngradeToTaxi || isJust req.canSwitchToRental || isJust req.canSwitchToInterCity || isJust req.availableUpiApps || isJust req.isPetModeEnabled || isJust req.tripDistanceMaxThreshold || isJust req.tripDistanceMinThreshold || isJust req.maxPickupRadius || isJust req.isSilentModeEnabled || isJust req.rideRequestVolume || isJust req.isTTSEnabled || isJust req.isHighAccuracyLocationEnabled || isJust req.rideRequestVolumeEnabled || isJust req.onboardingAs || isJust req.address || isJust req.addressDocumentType || isJust req.nomineeName || isJust req.nomineeRelationship) $ do
-    QDriverInformation.updateDriverInformation canDowngradeToSedan canDowngradeToHatchback canDowngradeToTaxi canSwitchToRental canSwitchToInterCity canSwitchToIntraCity availableUpiApps isPetModeEnabled tripDistanceMaxThreshold tripDistanceMinThreshold maxPickupRadius isSilentModeEnabled rideRequestVolume isTTSEnabled isHighAccuracyLocationEnabled rideRequestVolumeEnabled onboardingAs address addressDocumentType nomineeName nomineeRelationship person.id
+    QDriverInformation.updateDriverInformation canDowngradeToSedan canDowngradeToHatchback canDowngradeToTaxi canSwitchToRental canSwitchToInterCity canSwitchToIntraCity availableUpiApps isPetModeEnabled tripDistanceMaxThreshold tripDistanceMinThreshold maxPickupRadius isSilentModeEnabled rideRequestVolume isTTSEnabled isHighAccuracyLocationEnabled rideRequestVolumeEnabled onboardingAs legacyAddress legacyAddressDocumentType legacyNomineeName legacyNomineeRelationship person.id
 
   let petTag = Yudhishthira.TagNameValue "PetDriver#\"true\""
   when (isPetModeEnabled && maybe False (Yudhishthira.elemTagNameValue petTag) person.driverTag) $
@@ -1486,8 +1501,8 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
   when (isJust req.vehicleName) $ QVehicle.updateVehicleName req.vehicleName personId
   QPerson.updatePersonRec personId updPerson
   driverStats <- runInReplica $ QDriverStats.findById (cast personId) >>= fromMaybeM DriverInfoNotFound
-  let driverEntityArg :: (SP.Person, DriverInformation, DStats.DriverStats, Id DMOC.MerchantOperatingCity)
-      driverEntityArg = (updPerson, updatedDriverInfo, driverStats, merchantOpCityId)
+  let driverEntityArg :: (SP.Person, DriverInformation, DStats.DriverStats, Id DMOC.MerchantOperatingCity, DIInfo.IdentityInfo)
+      driverEntityArg = (updPerson, updatedDriverInfo, driverStats, merchantOpCityId, DIInfo.getIdentityInfo mbIdentityInfo updatedDriverInfo)
   let merchantId = person.merchantId
   org <-
     CQM.findById merchantId
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
@@ -1,0 +1,79 @@
+module SharedLogic.DriverIdentityInfo
+  ( IdentityInfo (..),
+    getIdentityInfo,
+    upsertDriverIdentityInfo,
+    driverIdentityInfoLockKey,
+  )
+where
+
+import Control.Applicative ((<|>))
+import Data.Time (Day)
+import qualified Domain.Types.DriverIdentityInfo as DII
+import qualified Domain.Types.DriverInformation as DI
+import qualified Domain.Types.Merchant as DM
+import qualified Domain.Types.MerchantOperatingCity as DMOC
+import qualified Domain.Types.Person as DP
+import Kernel.Prelude
+import Kernel.Types.Beckn.Context (IndianState)
+import Kernel.Types.Id (Id)
+import Kernel.Utils.Common (CacheFlow, EsqDBFlow, MonadFlow, getCurrentTime)
+import qualified Storage.Queries.DriverIdentityInfo as QDII
+
+data IdentityInfo = IdentityInfo
+  { nomineeName :: Maybe Text,
+    nomineeRelationship :: Maybe Text,
+    nomineeDob :: Maybe Day,
+    address :: Maybe Text,
+    addressDocumentType :: Maybe DI.AddressDocumentType,
+    addressState :: Maybe IndianState
+  }
+
+getIdentityInfo :: Maybe DII.DriverIdentityInfo -> DI.DriverInformation -> IdentityInfo
+getIdentityInfo mbInfo driverInfo =
+  IdentityInfo
+    { nomineeName = (mbInfo >>= (.nomineeName)) <|> driverInfo.nomineeName,
+      nomineeRelationship = (mbInfo >>= (.nomineeRelationship)) <|> driverInfo.nomineeRelationship,
+      nomineeDob = mbInfo >>= (.nomineeDob),
+      address = (mbInfo >>= (.address)) <|> driverInfo.address,
+      addressDocumentType = (mbInfo >>= (.addressDocumentType)) <|> driverInfo.addressDocumentType,
+      addressState = mbInfo >>= (.addressState)
+    }
+
+driverIdentityInfoLockKey :: Id DP.Person -> Text
+driverIdentityInfoLockKey driverId = "driver_identity_info_upsert:" <> driverId.getId
+
+-- Caller must run this inside `withLockRedis (driverIdentityInfoLockKey driverId)` with mbExisting read inside that lock.
+upsertDriverIdentityInfo ::
+  (MonadFlow m, CacheFlow m r, EsqDBFlow m r) =>
+  Maybe DII.DriverIdentityInfo ->
+  Id DP.Person ->
+  Id DM.Merchant ->
+  Id DMOC.MerchantOperatingCity ->
+  DI.DriverInformation ->
+  Maybe Text ->
+  Maybe Text ->
+  Maybe Day ->
+  Maybe Text ->
+  Maybe DI.AddressDocumentType ->
+  Maybe IndianState ->
+  m DII.DriverIdentityInfo
+upsertDriverIdentityInfo mbExisting driverId merchantId mocId driverInfo pName pRel pDob pAddr pAddrDoc pState = do
+  now <- getCurrentTime
+  let row =
+        DII.DriverIdentityInfo
+          { driverId = driverId,
+            nomineeName = pName <|> (mbExisting >>= (.nomineeName)) <|> driverInfo.nomineeName,
+            nomineeRelationship = pRel <|> (mbExisting >>= (.nomineeRelationship)) <|> driverInfo.nomineeRelationship,
+            nomineeDob = pDob <|> (mbExisting >>= (.nomineeDob)),
+            address = pAddr <|> (mbExisting >>= (.address)) <|> driverInfo.address,
+            addressDocumentType = pAddrDoc <|> (mbExisting >>= (.addressDocumentType)) <|> driverInfo.addressDocumentType,
+            addressState = pState <|> (mbExisting >>= (.addressState)),
+            merchantId = merchantId,
+            merchantOperatingCityId = mocId,
+            createdAt = maybe now (.createdAt) mbExisting,
+            updatedAt = now
+          }
+  case mbExisting of
+    Just _ -> QDII.updateByPrimaryKey row
+    Nothing -> QDII.create row
+  pure row
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/IffcoTokioInsurance.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/IffcoTokioInsurance.hs
@@ -24,9 +24,11 @@ import Kernel.Types.Id
 import Kernel.Utils.Common
 import Lib.ConfigPilot.Interface.Types (getOneConfig)
 import Servant.Client.Core ()
+import qualified SharedLogic.DriverIdentityInfo as DIInfo
 import qualified Storage.Cac.TransporterConfig as SCTC
 import qualified Storage.CachedQueries.Merchant.MerchantServiceConfig as CQMSC
 import Storage.ConfigPilot.Config.TransporterConfig (TransporterConfigDimensions (..))
+import qualified Storage.Queries.DriverIdentityInfo as QDII
 import qualified Storage.Queries.DriverInformation as QDI
 import qualified Storage.Queries.DriverLicense as QDL
 import qualified Storage.Queries.IffcoTokioInsurance as QIffco
@@ -78,8 +80,10 @@ triggerIffcoTokioInsurance driverId merchantId merchantOpCityId = do
     unless alreadyInsured $ do
       person <- QPerson.findById (cast driverId) >>= fromMaybeM (PersonNotFound driverId.getId)
       driverInfo <- QDI.findByPrimaryKey (cast driverId) >>= fromMaybeM DriverInfoNotFound
-      let nomineeName = driverInfo.nomineeName
-          nomineeRelationship = driverInfo.nomineeRelationship
+      mbIdentity <- QDII.findByDriverId (cast driverId)
+      let identityInfo = DIInfo.getIdentityInfo mbIdentity driverInfo
+          nomineeName = identityInfo.nomineeName
+          nomineeRelationship = identityInfo.nomineeRelationship
       if isNothing nomineeName || isNothing nomineeRelationship
         then logInfo $ "IffcoTokio: skipping insurance for driver=" <> driverId.getId <> " nomineeName or nomineeRelationship is missing"
         else do
--- a/Backend/dev/integration-tests/collections/DriverProfileNomineeFlow/01-DriverProfileNomineeFlow.json
+++ b/Backend/dev/integration-tests/collections/DriverProfileNomineeFlow/01-DriverProfileNomineeFlow.json
@@ -1,0 +1,197 @@
+{
+  "info": {
+    "name": "DriverProfileNomineeFlow",
+    "schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
+    "description": "E2E for the DriverIdentityInfo cohort: nominee+address moved off driver_information. Random driver per run (idempotent, repeatable on any device). Covers full update (incl. addressState), migrated read, and partial-merge."
+  },
+  "event": [
+    {
+      "listen": "prerequest",
+      "script": {
+        "type": "text/javascript",
+        "exec": [
+          "if (!pm.collectionVariables.get('_test_driver_number')) {",
+          "  var rNum = '9' + Math.floor(100000000 + Math.random() * 899999999);",
+          "  pm.collectionVariables.set('_test_driver_number', rNum);",
+          "}",
+          "// mock-server auto-skip (per Rules.md): skip mock-only requests on non-Local envs",
+          "var envType = pm.environment.get('envType') || 'Local';",
+          "var url = (pm.request && pm.request.url) ? pm.request.url.toString() : '';",
+          "var mockUrl = pm.environment.get('mockServerUrl') || '__no_mock__';",
+          "if (envType !== 'Local' && url.indexOf(mockUrl) !== -1 && pm.execution && pm.execution.skipRequest) {",
+          "  pm.execution.skipRequest();",
+          "}"
+        ]
+      }
+    }
+  ],
+  "variable": [
+    {
+      "key": "_test_driver_number",
+      "value": ""
+    }
+  ],
+  "item": [
+    {
+      "name": "Driver Auth",
+      "request": {
+        "method": "POST",
+        "header": [
+          {
+            "key": "Content-Type",
+            "value": "application/json"
+          }
+        ],
+        "url": "{{baseURL_namma_P}}/auth",
+        "body": {
+          "mode": "raw",
+          "raw": "{\"mobileNumber\": \"{{_test_driver_number}}\", \"mobileCountryCode\": \"+91\", \"merchantId\": \"{{driver_merchant_id}}\", \"merchantOperatingCity\": \"{{city}}\"}"
+        }
+      },
+      "event": [
+        {
+          "listen": "test",
+          "script": {
+            "type": "text/javascript",
+            "exec": [
+              "pm.test('auth 200', function(){ pm.response.to.have.status(200); });",
+              "var d = pm.response.json();",
+              "pm.expect(d.authId, 'authId').to.be.a('string');",
+              "pm.collectionVariables.set('driver_authId', d.authId);"
+            ]
+          }
+        }
+      ]
+    },
+    {
+      "name": "Driver OTP",
+      "request": {
+        "method": "POST",
+        "header": [
+          {
+            "key": "Content-Type",
+            "value": "application/json"
+          }
+        ],
+        "url": "{{baseURL_namma_P}}/auth/{{driver_authId}}/verify",
+        "body": {
+          "mode": "raw",
+          "raw": "{\"otp\": \"{{login_otp}}\", \"deviceToken\": \"test-device-nominee\"}"
+        }
+      },
+      "event": [
+        {
+          "listen": "test",
+          "script": {
+            "type": "text/javascript",
+            "exec": [
+              "pm.test('verify 200', function(){ pm.response.to.have.status(200); });",
+              "var d = pm.response.json();",
+              "pm.expect(d.token, 'token').to.be.a('string');",
+              "pm.collectionVariables.set('driver_token', d.token);"
+            ]
+          }
+        }
+      ]
+    },
+    {
+      "name": "Update Nominee And Address (Full)",
+      "request": {
+        "method": "POST",
+        "header": [
+          {
+            "key": "token",
+            "value": "{{driver_token}}"
+          },
+          {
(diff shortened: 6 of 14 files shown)


bytedex (commented):
@ClaudeWeb review this PR

khuzema786:
I'll fetch the PR diff, description, and existing comments in parallel before diving into the review.Let me read the full diff and explore the workspace files.Let me check a few more critical files in the workspace before writing the review.The workspace is likely empty. Let me use the GitHub MCP tool to fetch relevant files directly.Now let me check `buildDriverInformationRes` to see how `nomineeDob`/`nomineeAge` are populated in that path.Now I have all the context I need. Let me write the full review.

---

## PR Review: `backend/feat: add driver_identity_info table for nominee+address`

**Overall verdict: Request Changes.** The backward-compatible fallback design is sound and the integration test collection is a nice touch, but there is a **critical data-loss bug** in the main `updateDriver` path that must be fixed before merge, plus a TOCTOU race in the upsert helper.

---

### 🔴 Critical — Data Loss

#### 1. `updateDriverInformation` unconditionally nulls legacy columns even when `upsertNomineeAddress` was never called

**File:** `src/Domain/Action/UI/Driver.hs`

The PR passes `Nothing Nothing Nothing Nothing` for `address`, `addressDocumentType`, `nomineeName`, `nomineeRelationship` in every call to `updateDriverInformation`. The gate condition for that call includes **many non-nominee fields**:

```haskell
when (isJust req.canDowngradeToSedan   -- ← non-nominee
   || isJust req.isPetModeEnabled       -- ← non-nominee
   || isJust req.isTTSEnabled           -- ← non-nominee
   -- ... 12 more non-nominee fields ...
   || isJust req.nomineeName) $ do
    QDriverInformation.updateDriverInformation
      ... Nothing Nothing Nothing Nothing person.id   -- wipes address + nominee
```

The `upsertNomineeAddress` guard is a **separate, narrower** condition:

```haskell
when (isJust req.nomineeName || isJust req.nomineeRelationship
   || isJust req.nomineeDob  || isJust req.nomineeAge
   || isJust req.address     || isJust req.addressDocumentType) $
    DIInfo.upsertNomineeAddress ...
```

**Failure scenario:**

1. Existing driver has `nomineeName = "Alice"`, `nomineeRelationship = "Spouse"` in `driver_information` (legacy row; `driver_identity_info` is empty).
2. Driver sends `PATCH /driver/profile` with only `{ "canDowngradeToSedan": true }`.
3. Upsert condition → **false** → `driver_identity_info` remains empty.
4. `updateDriverInformation` condition → **true** (because of `canDowngradeToSedan`) → sets `nomineeName = NULL`, `nomineeRelationship = NULL` in `driver_information`.
5. `getIdentityInfo` now finds nothing in either table → **nominee data permanently lost**.

**Fix:** Only null the legacy columns when the upsert actually ran, or read and preserve the old values when no nominee field is in the request.

```haskell
-- Option A: conditional nulling
let shouldNullLegacyNominee = isJust req.nomineeName || isJust req.nomineeRelationship
                           || isJust req.nomineeDob   || isJust req.nomineeAge
                           || isJust req.address      || isJust req.addressDocumentType
    legacyAddress       = if shouldNullLegacyNominee then Nothing else driverInfo.address
    legacyAddrDocType   = if shouldNullLegacyNominee then Nothing else driverInfo.addressDocumentType
    legacyNomineeName   = if shouldNullLegacyNominee then Nothing else driverInfo.nomineeName
    legacyNomineeRel    = if shouldNullLegacyNominee then Nothing else driverInfo.nomineeRelationship
QDriverInformation.updateDriverInformation
  ... legacyAddress legacyAddrDocType legacyNomineeName legacyNomineeRel person.id
```

---

### 🟠 Warning — Race Condition

#### 2. TOCTOU race in `upsert`

**File:** `src/Storage/Queries/DriverIdentityInfoExtra.hs:613-627`

```haskell
upsert a@DriverIdentityInfo {..} = do
  res <- findOneWithKV [Se.Is Beam.driverId $ Se.Eq a.driverId.getId]
  now <- getCurrentTime
  if isJust res
    then updateOneWithKV ...
    else createWithKV a   -- ← another thread may have inserted between find and create
```

In a multi-instance deployment, two concurrent profile updates for the same driver can both observe `isNothing res` and both attempt `createWithKV`, hitting the primary-key constraint. Other tables in this repo (e.g. `DriverGstin`) use the same pattern but this is the only new table in this PR.

**Fix:** Push the upsert semantics to the database layer:

```sql
INSERT INTO driver_identity_info (...) VALUES (...)
ON CONFLICT (driver_id) DO UPDATE SET
  nominee_name = EXCLUDED.nominee_name,
  ...
  updated_at  = EXCLUDED.updated_at;
```

If the KV abstraction doesn't expose `ON CONFLICT DO UPDATE`, at minimum wrap the read+write in a `withTransaction` block or catch the PK violation and retry with an update.

---

#### 3. Non-atomic upsert + legacy-column nulling in `updateDriver`

**File:** `src/Domain/Action/UI/Driver.hs`

`upsertNomineeAddress` and the subsequent `updateDriverInformation` are two independent DB operations with no surrounding transaction. A crash between them means:

- Upsert succeeded, null didn't → both tables have data (harmless, read path handles it)
- Null succeeded, upsert didn't → **legacy data wiped, new table empty** (data loss, same failure mode as finding #1)

These two calls should be wrapped in a single DB transaction, or at minimum the null should be made idempotent by only running after a confirmed successful upsert.

---

#### 4. Dashboard path nulls address *before* writing to new table

**File:** `src/Domain/Action/Dashboard/Fleet/Driver.hs:4526-4529`

```haskell
QDriverInfo.updateDriverDobAndAddress dob Nothing Nothing personId  -- nulls first
DIInfo.upsertNomineeAddress ...                                      -- saves second
```

A crash between these two lines would permanently lose the address for this driver. (Note: `getIdentityInfo` uses the already-fetched in-memory `driverInfo` as fallback, so the runtime behavior is safe; the crash window is the concern.) The order should be reversed — write to the new table first, then null the old one.

---

### 🟡 Potential Issues

#### 5. `deriveState` — `isInfixOf` can produce wrong state due to list ordering

**File:** `src/SharedLogic/DriverIdentityInfo.hs:519-523`

```haskell
find (\st -> T.toLower st `T.isInfixOf` lower) indianStates
```

`find` returns the **first** match. "Uttar Pradesh" appears **before** "Uttarakhand" in `indianStates`. An address like `"Old Tehri, Uttarakhand (near Uttar Pradesh border)"` contains the substring `"uttar pradesh"` and would be misclassified as Uttar Pradesh. Similarly, any address mentioning a state name en passant ("Goa Foods Warehouse, Bangalore, Karnataka") would be misclassified.

Consider using longest-match or right-to-left scanning, or suffixing state names with commas/digits to reduce ambiguity.

---

#### 6. `merchantId`/`merchantOperatingCityId` not refreshed on UPDATE in `upsert`

**File:** `src/Storage/Queries/DriverIdentityInfoExtra.hs:616-625`

The `updateOneWithKV` branch omits `merchantId` and `merchantOperatingCityId`:

```haskell
updateOneWithKV
  [ Se.Set Beam.updatedAt now,
    Se.Set Beam.nomineeName nomineeName,
    ...
    -- merchantId and merchantOperatingCityId not refreshed
  ]
```

If a driver is ever re-enrolled under a different merchant operating city, the row will carry stale IDs. This is likely benign given current business rules but worth a comment.

---

#### 7. Double `getCurrentTime` in the upsert chain

`upsertNomineeAddress` calls `getCurrentTime` (to stamp `createdAt`/`updatedAt` on the built row), and `upsert` calls it *again* (to stamp `updatedAt` on the update SET). The row passed to `createWithKV` uses the first timestamp; the `updateOneWithKV` uses the second. The inconsistency is harmless but the second call is redundant if you rely on the caller-provided `updatedAt`. Either remove the inner `now <- getCurrentTime` and use `a.updatedAt`, or remove it from the caller.

---

### ✅ What's Good

- **Fallback logic** (`getIdentityInfo`) is clean and correct — new-table-first with `<|>` fallback ensures zero regressions for drivers with legacy data only.
- **`dobToAge` / `ageToDob` implementations** are correct for all valid inputs (validated range 18-99 prevents edge cases).
- **Validation** (relationship allowlist, age bounds) is well-placed and symmetric between `nomineeDob` and `nomineeAge` paths.
- **Integration test collection** covers the key flows (full update, partial merge, age→dob, validation errors) and is a good addition.
- The YAML spec and generated Beam/Query/OrphanInstances boilerplate look correct and consistent with the rest of the codebase.

Backend/dev/integration-tests/collections/DriverProfileNomineeFlow/01-DriverProfileNomineeFlow.json
```diff
@@ -0,0 +1,301 @@
...
+      },
+      "event": [
+        {
+          "listen": "test",
+          "script": {
+            "type": "text/javascript",
+            "exec": [
+              "pm.test('rejected', function(){ pm.expect(pm.response.code).to.be.oneOf([400,500]); });",
+              "var d = pm.response.json();",
+              "pm.test('invalid relationship error', function(){ pm.expect(d.errorCode).to.eql('INVALID_REQUEST'); pm.expect(d.errorMessage).to.include('relationship'); });"
```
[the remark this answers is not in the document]
bytedex: Not required, this is for testing both 400, 500 response

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/DriverIdentityInfo.yaml
```diff
@@ -0,0 +1,35 @@
...
+  tableName: driver_identity_info
+
+  fields:
+    driverId: Id Person
+    nomineeName: Maybe Text
+    nomineeRelationship: Maybe Text
+    nomineeDob: Maybe Day
+    address: Maybe Text
+    addressDocumentType: Maybe AddressDocumentType
+    state: Maybe Text
```
RitikaHanish: There is IndianState defined in Shared-kernel already, Please use that. 
Also check if any state is missing there, you can add.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverIdentityInfoExtra.hs
```diff
@@ -0,0 +1,37 @@
...
+        [ Se.Set Beam.updatedAt now,
+          Se.Set Beam.nomineeName nomineeName,
+          Se.Set Beam.nomineeRelationship nomineeRelationship,
+          Se.Set Beam.nomineeDob nomineeDob,
+          Se.Set Beam.address address,
+          Se.Set Beam.addressDocumentType addressDocumentType,
+          Se.Set Beam.state state
+        ]
+        [Se.Is Beam.driverId $ Se.Eq a.driverId.getId]
+    else createWithKV a
```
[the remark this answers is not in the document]
RitikaHanish: @bytedex Can you please check it ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,137 @@
...
+import Kernel.Utils.Common (CacheFlow, EsqDBFlow, MonadFlow, getCurrentTime)
+import qualified Storage.Queries.DriverIdentityInfo as QDII
+
+data IdentityInfo = IdentityInfo
+  { nomineeName :: Maybe Text,
+    nomineeRelationship :: Maybe Text,
+    nomineeDob :: Maybe Day,
+    address :: Maybe Text,
+    addressDocumentType :: Maybe DI.AddressDocumentType,
+    state :: Maybe Text
```
RitikaHanish: can be renamed as addressState

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,137 @@
...
+getIdentityInfo driverId driverInfo = do
+  mbInfo <- QDII.findByDriverId driverId
+  pure
+    IdentityInfo
+      { nomineeName = (mbInfo >>= (.nomineeName)) <|> driverInfo.nomineeName,
+        nomineeRelationship = (mbInfo >>= (.nomineeRelationship)) <|> driverInfo.nomineeRelationship,
+        nomineeDob = mbInfo >>= (.nomineeDob),
+        address = (mbInfo >>= (.address)) <|> driverInfo.address,
+        addressDocumentType = (mbInfo >>= (.addressDocumentType)) <|> driverInfo.addressDocumentType,
+        state = mbInfo >>= (.state)
```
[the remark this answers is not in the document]
RitikaHanish: No need of this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,137 @@
...
+   in fromInteger (ty - by) - (if hadBirthdayThisYear then 0 else 1)
+
+deriveState :: Maybe Text -> Maybe Text
+deriveState mbAddr = do
+  addr <- mbAddr
+  let lower = T.toLower addr
+      ranked = [(T.length (fst (T.breakOnEnd (T.toLower st) lower)), st) | st <- indianStates, T.toLower st `T.isInfixOf` lower]
+  if null ranked then Nothing else Just (snd (foldr1 (\x y -> if fst x >= fst y then x else y) ranked))
+
+indianStates :: [Text]
```
RitikaHanish: remove it, already there in Shared-Kernel.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
RitikaHanish: In the getIdentityInfo itself, fallback can be added, to add it in indentity info.

here itself upsert for identityInfo can be called.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -4522,9 +4523,8 @@ postDriverFleetDriverUpdate merchantShortId opCity driverId requestorId req = do
         DP.DRIVER -> do
           driverInfo <- QDriverInfo.findById personId >>= fromMaybeM DriverInfoNotFound
           let dob = fmap (\d -> UTCTime d 0) req.dob <|> driverInfo.driverDob
-              address = req.address <|> driverInfo.address
-              addressDocumentType = castAddressDocumentType <$> req.addressDocumentType <|> driverInfo.addressDocumentType
-          QDriverInfo.updateDriverDobAndAddress dob address addressDocumentType personId
+          DIInfo.upsertNomineeAddress personId driver.merchantId driver.merchantOperatingCityId driverInfo Nothing Nothing Nothing req.address (castAddressDocumentType <$> req.addressDocumentType)
+          QDriverInfo.updateDriverDobAndAddress dob Nothing Nothing personId
```
RitikaHanish, line 4538: This we can remove.

bytedex: @RitikaHanish No we are not moving dob now 👍

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
bytedex: Will add a dashboard API for backfilling as discussed

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
bytedex: @RitikaHanish Will have to add State conversion from Text -> IndianStates

RitikaHanish:
@ClaudeWeb Can you review the PR by making sure performance, system design and efficiency ?

khuzema786:
Let me pull fresh context on the KV layer, query patterns, and how `buildDriverEntityRes` is wired up across call sites.Now I have a complete picture of the call graph, query counts, and data flow. Let me write the focused performance/system-design review.

---

## PR #15056 — Performance, System Design & Efficiency Review

**Verdict: Several efficiency issues worth addressing, one medium-severity design gap (no backfill), and one structural redundancy in the write path. None are blocking on their own, but together they add measurable latency to every profile fetch.**

---

### 🔴 Performance — Double KV Read Per Write (Redundant I/O)

**Files:** `SharedLogic/DriverIdentityInfo.hs:482`, `Storage/Queries/DriverIdentityInfoExtra.hs:613`

`upsertNomineeAddress` calls `getIdentityInfo` (KV read #1 to merge old values), then immediately delegates to `upsert`, which calls `findOneWithKV` (KV read #2 to decide insert-vs-update) on the **same key**:

```
upsertNomineeAddress
  └─ getIdentityInfo → findByDriverId         ← KV read #1
  └─ upsert
       └─ findOneWithKV                        ← KV read #2 (same driverId key)
       └─ updateOneWithKV / createWithKV       ← KV write
```

Read #2 is fully redundant — the result of read #1 already tells you whether the row exists. Two Redis round-trips where one suffices, on the hottest write path.

**Fix:** Pass the `Maybe DII.DriverIdentityInfo` that `getIdentityInfo` already has into `upsert`:

```haskell
-- DriverIdentityInfoExtra.hs
upsert :: (MonadFlow m, CacheFlow m r, EsqDBFlow m r)
       => Maybe DriverIdentityInfo   -- already-fetched current row
       -> DriverIdentityInfo         -- new row to write
       -> m ()
upsert mbExisting new = do
  now <- getCurrentTime
  case mbExisting of
    Just _  -> updateOneWithKV [ Se.Set Beam.updatedAt now, ... ] [Se.Is Beam.driverId ...]
    Nothing -> createWithKV new
```

This removes the second KV round-trip entirely.

---

### 🟠 Performance — Unconditional KV Read Added to Every Profile Fetch

**File:** `src/Domain/Action/UI/Driver.hs`, `buildDriverEntityRes` (line ~1152)

```haskell
buildDriverEntityRes (person, driverInfo, ...) merchant serviceName = do
  identityInfo <- DIInfo.getIdentityInfo person.id driverInfo   -- NEW: always fires
  transporterConfig <- ...
  vehicleMB <- ...
  -- 15+ more queries follow
```

`buildDriverEntityRes` is called from **both** `getInformation` (GET /driver/profile) and `updateDriver` (PATCH /driver/profile), and transitively from `getInformationV2`. Every profile fetch now pays an extra KV round-trip even for the vast majority of drivers who have no row in `driver_identity_info` yet.

The KV hit is fast, but at scale (thousands of concurrent profile fetches) it adds aggregate latency and Redis connection pressure. This is unavoidable for the fallback to work, but the impact window is large. Consider:

- **Short-term:** Add a structured log or metric (`identityInfoCacheMiss`) to track the miss rate post-deploy, so you know when the fleet has fully migrated.
- **Medium-term:** Backfill existing rows (see system design point below) to convert KV misses to hits.

---

### 🟠 System Design — No Backfill Migration Means Persistent KV Misses at Deploy

**File:** `dev/migrations-read-only/dynamic-offer-driver-app/driver_identity_info.sql`

The SQL migration creates the table empty. Every existing driver hits:

```
getIdentityInfo → findByDriverId
  → KV miss
  → Postgres SELECT (PK lookup: fast, but still a round-trip)
  → returns Nothing
  → falls back to driverInfo.nomineeName / driverInfo.address
```

For a fleet with N active drivers, the first deploy triggers **N extra Postgres reads** (spread across the first few minutes of traffic). The KV layer will gradually warm up as each driver's profile is fetched, but the `state` field (new in `driver_identity_info`, derived from `address`) will **never be backfilled** for existing drivers — it will remain `Nothing` forever unless the driver re-saves their profile.

**Recommended fix:** Add a one-time data migration:

```sql
INSERT INTO atlas_driver_offer_bpp.driver_identity_info
  (driver_id, nominee_name, nominee_relationship, address, address_document_type,
   merchant_id, merchant_operating_city_id, created_at, updated_at)
SELECT
  driver_id, nominee_name, nominee_relationship, address, address_document_type,
  merchant_id, merchant_operating_city_id, now(), now()
FROM atlas_driver_offer_bpp.driver_information
WHERE nominee_name IS NOT NULL
   OR address IS NOT NULL
ON CONFLICT (driver_id) DO NOTHING;
```

This pre-populates the new table and eliminates both the cold-start Postgres hits and the gap in `state` data for existing drivers.

---

### 🟠 System Design — `updateDriver` Fires 3 Reads + 2 Writes for a Single Nominee Update

**File:** `src/Domain/Action/UI/Driver.hs`, `updateDriver`

Full I/O sequence for one `PATCH /driver/profile` with `nomineeName`:

| Step | Operation | Notes |
|---|---|---|
| 1 | `upsertNomineeAddress → getIdentityInfo → findByDriverId` | KV read (merge base) |
| 2 | `upsert → findOneWithKV` | KV read (redundant, see above) |
| 3 | `upsert → createWithKV / updateOneWithKV` | KV write + Postgres |
| 4 | `updateDriverInformation → updateOneWithKV` | Postgres write (nulls legacy cols) |
| 5 | `updatedDriverInfo ← findById` | KV read (re-fetch for response) |
| 6 | `buildDriverEntityRes → getIdentityInfo → findByDriverId` | KV read (should be cache hit after step 3) |

Steps 1 + 2 are two reads on the same key. Fixing the double-read (as above) plus confirming the KV write in step 3 properly warms the cache for step 6 would reduce this to **2 KV reads + 2 writes**, which is the practical minimum.

---

### 🟡 Design — `state` Derivation Is Order-Dependent and Fragile

**File:** `src/SharedLogic/DriverIdentityInfo.hs:519-523`

```haskell
find (\st -> T.toLower st `T.isInfixOf` lower) indianStates
```

`find` returns the **first** match in `indianStates`. The current list has "Uttar Pradesh" before "Uttarakhand". An address like `"Tehri, Uttarakhand near Uttar Pradesh"` is tested against "uttar pradesh" *first* — and since "uttar pradesh" is a substring of that string, the driver is classified as being in Uttar Pradesh.

Beyond ordering: the `isInfixOf` approach matches any substring. "Nagaland Road, Bangalore, Karnataka" would be classified as Nagaland.

Since `state` is used downstream (potentially for routing, analytics, or insurance eligibility), false positives here have real consequences.

**Better approach:** Match only at word boundaries, or after stripping punctuation, or require the state name to follow a pin-code/city pattern. At minimum, test against the full `indianStates` list and pick the **longest** match rather than the first:

```haskell
deriveState mbAddr = do
  addr <- mbAddr
  let lower = T.toLower addr
  maximumByMaybe (comparing T.length)
    [ st | st <- indianStates, T.toLower st `T.isInfixOf` lower ]
```

---

### 🟡 Design — `merchantId` / `merchantOperatingCityId` Are Write-Once but Not Enforced

**File:** `src/Storage/Queries/DriverIdentityInfoExtra.hs:616-625`

The `updateOneWithKV` branch in `upsert` does not update `merchantId` or `merchantOperatingCityId`:

```haskell
updateOneWithKV
  [ Se.Set Beam.updatedAt now,
    Se.Set Beam.nomineeName nomineeName,
    -- merchantId, merchantOperatingCityId absent
  ]
```

This means the column values are frozen at row-creation time. If the intent is "write-once" semantics (driver never changes merchant/city), that assumption should be documented. If a driver can transfer cities, add a `Se.Set Beam.merchantOperatingCityId` to the update set.

---

### 🟡 Design — `ageToDob` Always Produces January 1st (Lossy Conversion Creates Clustering)

**File:** `src/SharedLogic/DriverIdentityInfo.hs:504-507`

```haskell
ageToDob today age =
  let (y, _, _) = toGregorian today
   in fromGregorian (y - toInteger age) 1 1   -- always Jan 1
```

Every client submitting `nomineeAge` gets their nominee's DOB stored as `YYYY-01-01`. Consequences:

1. **Analytics distortion:** Any age/DOB distribution query will show an artificial spike on Jan 1 for every birth year.
2. **Re-fetch mismatch:** A client posts `nomineeAge=30` today (Jun 11). The stored DOB is `1996-01-01`. If they GET the profile tomorrow, they see `nomineeAge=30` (same year, so it matches). But on Jan 1 next year, `dobToAge(2027-01-01, 1996-01-01) = 31`. The client's nominee "aged" overnight without any update. This is expected from a DOB perspective but surprising from an age perspective.
3. **Insurance/eligibility edge:** If `state` or nominee data is used for insurance triggers, the Jan-1 clustering could confuse downstream analytics.

**Recommendation:** If age-only input is supported, either store `nomineeAge` as-is in a separate column (not convert to DOB), or document the Jan-1 convention explicitly in the API schema description.

---

### ✅ What's Efficient

- **PK-only index** is correct for the access pattern — all lookups are by `driver_id`, no scan needed.
- **`indianStates` as a module-level constant** avoids repeated allocation.
- **`enableKVPG` with `[('driverId)] []`** correctly maps the PK to the KV key with no secondary index overhead.
- **`findByDriverId` == `findByPrimaryKey`** in practice — both filter on `driverId`. The auto-generated `findByPrimaryKey` (with `Se.And [ Se.Is ... ]`) is slightly more verbose than needed but functionally identical.
- **`buildDriverEntityRes` result is cached in-memory** within the request (computed once, used in `makeDriverInformationRes`) — no repeated fetch within a single request.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -4526,9 +4523,8 @@ postDriverFleetDriverUpdate merchantShortId opCity driverId requestorId req = do
         DP.DRIVER -> do
           driverInfo <- QDriverInfo.findById personId >>= fromMaybeM DriverInfoNotFound
           let dob = fmap (\d -> UTCTime d 0) req.dob <|> driverInfo.driverDob
-              address = req.address <|> driverInfo.address
-              addressDocumentType = castAddressDocumentType <$> req.addressDocumentType <|> driverInfo.addressDocumentType
-          QDriverInfo.updateDriverDobAndAddress dob address addressDocumentType personId
+          DIInfo.upsertNomineeAddress personId driver.merchantId driver.merchantOperatingCityId driverInfo Nothing Nothing Nothing req.address (castAddressDocumentType <$> req.addressDocumentType)
```
RitikaHanish: upsertNomineeAddress please rename to convey nominee and address are separate not nominee's address it;s driver's address

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -473,7 +474,9 @@ data DriverInformationRes = DriverInformationRes
     membershipId :: Maybe (Id DStclMembership.StclMembership),
     createdAt :: UTCTime,
     role :: SP.Role,
-    operatorBadgeToken :: Maybe Text
+    operatorBadgeToken :: Maybe Text,
+    nomineeDob :: Maybe Day,
+    nomineeAge :: Maybe Int
```
RitikaHanish: Age can be removed, because just from years, exact date can't be predicted.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1461,9 +1467,25 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
               DV.AUTO_LITE -> [DVST.AUTO_LITE]
               DV.PINK_AUTO -> [DVST.PINK_AUTO]
       QVehicle.updateSelectedServiceTiers selectedServiceTiers person.id
-  -- Update driver information (works for both cases: with or without vehicle)
```
RitikaHanish, line 1467: keep this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1461,9 +1467,25 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
...
+  -- Nominee + address now live in driver_identity_info. Validate, upsert there, then null the old
+  -- driver_information columns in the updateDriverInformation call below.
+  nowT <- getCurrentTime
```
RitikaHanish: it might be already there, check once

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1461,9 +1467,25 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
...
+  let mbNomineeDob = req.nomineeDob <|> (DIInfo.ageToDob (utctDay nowT) <$> req.nomineeAge)
```
RitikaHanish: remove this age to DOB

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1461,9 +1467,25 @@ updateDriver (personId, _, merchantOpCityId) mbBundleVersion mbClientVersion mbC
...
+  whenJust req.nomineeRelationship $ \rel ->
+    unless (rel `elem` ["Spouse", "Daughter", "Son", "Mother", "Father"]) $
```
RitikaHanish: remove this not required for now.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,140 @@
...
+    IdentityInfo
+      { nomineeName = (mbInfo >>= (.nomineeName)) <|> driverInfo.nomineeName,
+        nomineeRelationship = (mbInfo >>= (.nomineeRelationship)) <|> driverInfo.nomineeRelationship,
+        nomineeDob = mbInfo >>= (.nomineeDob),
+        address = (mbInfo >>= (.address)) <|> driverInfo.address,
+        addressDocumentType = (mbInfo >>= (.addressDocumentType)) <|> driverInfo.addressDocumentType,
+        addressState = mbInfo >>= (.addressState)
+      }
+
+upsertNomineeAddress ::
```
RitikaHanish: correct it name

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,140 @@
...
+   in fromGregorian (y - toInteger age) 1 1
+
+dobToAge :: Day -> Day -> Int
+dobToAge today dob =
+  let (ty, tm, td) = toGregorian today
+      (by, bm, bd) = toGregorian dob
+      hadBirthdayThisYear = (tm, td) >= (bm, bd)
+   in fromInteger (ty - by) - (if hadBirthdayThisYear then 0 else 1)
+
+deriveState :: Maybe Text -> Maybe IndianState
```
RitikaHanish: we can remove this.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,140 @@
...
+deriveState mbAddr = do
+  addr <- mbAddr
+  let lower = T.toLower addr
+      ranked = [(T.length (fst (T.breakOnEnd (T.toLower name) lower)), st) | (name, st) <- indianStateNames, T.toLower name `T.isInfixOf` lower]
+  if null ranked then Nothing else Just (snd (foldr1 (\x y -> if fst x >= fst y then x else y) ranked))
+
+indianStateNames :: [(Text, IndianState)]
```
RitikaHanish: remove it

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverIdentityInfo.hs
```diff
@@ -0,0 +1,140 @@
...
+  let mergedAddress = pAddr <|> current.address
+  let row =
+        DII.DriverIdentityInfo
+          { driverId = driverId,
+            nomineeName = pName <|> current.nomineeName,
+            nomineeRelationship = pRel <|> current.nomineeRelationship,
+            nomineeDob = pDob <|> current.nomineeDob,
+            address = mergedAddress,
+            addressDocumentType = pAddrDoc <|> current.addressDocumentType,
+            addressState = deriveState mergedAddress,
```
RitikaHanish: no need to derive, if it's there, then fine else it can be nothing

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverIdentityInfoExtra.hs
```diff
@@ -0,0 +1,35 @@
...
+import Kernel.Prelude
+import Kernel.Types.Error
+import Kernel.Utils.Common (CacheFlow, EsqDBFlow, MonadFlow, fromMaybeM, getCurrentTime)
+import qualified Sequelize as Se
+import qualified Storage.Beam.DriverIdentityInfo as Beam
+import Storage.Queries.OrphanInstances.DriverIdentityInfo
+
+-- Extra code goes here --
+
+upsert :: (MonadFlow m, CacheFlow m r, EsqDBFlow m r) => DriverIdentityInfo -> m ()
```
RitikaHanish: Please change it.

RitikaHanish:
@ClaudeWeb Can you please review for error handling also ?
Make sure all error messages are handling correctly, error codes are proper.

khuzema786:
Let me pull the diff and check the existing error-handling patterns in the codebase for comparison.Now let me look at the error-handling code in the Driver.hs context and the existing error types.Now let me extract the error types from the saved Tools.Error.hs file.Now I have everything I need. Let me write the error-handling review.I've confirmed the ordering issue from the diff. Now I have everything for the review.

---

## PR #15056 — Error Handling Review

**Verdict: Request Changes.** There are two structural correctness issues (write-before-validate ordering, and DB errors surfacing as raw 500s), four error message quality issues, and a missing validation parity gap in the dashboard path.

---

### 🔴 Critical — Validation Runs AFTER a DB Write Has Already Committed

**File:** `src/Domain/Action/UI/Driver.hs`, `updateDriver`

The order in the PR's updated `updateDriver` is:

```haskell
-- Step 1: DB WRITE — vehicle service tiers committed to DB
QVehicle.updateSelectedServiceTiers selectedServiceTiers person.id

-- Step 2: THEN validation runs (new code)
nowT <- getCurrentTime
let mbNomineeDob = req.nomineeDob <|> ...
whenJust req.nomineeRelationship $ \rel ->
  unless (rel `elem` [...]) $
    throwError (InvalidRequest "Invalid nominee relationship")   -- thrown AFTER step 1
whenJust mbNomineeDob $ \dob ->
  unless (...) $
    throwError (InvalidRequest "Nominee age must be between 18 and 99")   -- thrown AFTER step 1
```

A request that combines `canDowngradeToSedan = true` (or any vehicle preference change) **and** an invalid `nomineeRelationship = "Friend"` will:

1. Commit the service tier change to the database ✓
2. Throw 400 `INVALID_REQUEST` and roll back nothing

The client receives an error response, but the vehicle preference is already changed. The system is now in a state the client doesn't know about.

**Fix:** Move both `whenJust` validation blocks to run **before** any DB write — i.e., before the `case mVehicle of` block, immediately after `mbNomineeDob` is computed:

```haskell
-- at the TOP of the write section, before any QVehicle / QDriverInformation call
let mbNomineeDob = req.nomineeDob <|> (DIInfo.ageToDob (utctDay nowT) <$> req.nomineeAge)
-- validate first
whenJust req.nomineeRelationship $ \rel -> ...
whenJust mbNomineeDob $ \dob -> ...
-- only then proceed to writes
case mVehicle of ...
```

---

### 🔴 Critical — TOCTOU Race Produces Raw DB Exception (Potential 500 + Info Leak)

**File:** `src/Storage/Queries/DriverIdentityInfoExtra.hs:613-627`

```haskell
upsert a@DriverIdentityInfo {..} = do
  res <- findOneWithKV [Se.Is Beam.driverId $ Se.Eq a.driverId.getId]
  now <- getCurrentTime
  if isJust res
    then updateOneWithKV [...] [...]
    else createWithKV a          -- ← if two threads both reach here, second gets PK violation
```

In a multi-instance deployment, two concurrent profile updates for the same driver can both see `isNothing res` and both call `createWithKV`. The second call hits a PostgreSQL primary-key constraint violation. This exception:

- Is **not caught** anywhere in `upsert`, `upsertNomineeAddress`, or `updateDriver`
- Bubbles all the way to the framework's top-level handler as an unhandled exception
- Returns HTTP **500** with a raw DB error message (e.g., `duplicate key value violates unique constraint "driver_identity_info_pkey"`)
- Leaks internal table names and constraint names to the API response

**Fix:** Catch the specific exception and either retry with an update or return a domain error:

```haskell
upsert a@DriverIdentityInfo {..} = do
  res <- findOneWithKV [Se.Is Beam.driverId $ Se.Eq a.driverId.getId]
  now <- getCurrentTime
  if isJust res
    then updateOneWithKV [...] [...]
    else createWithKV a
      `catch` \(e :: SomeException) ->
        -- Lost the race — row now exists, retry as update
        if "duplicate key" `T.isInfixOf` show e
          then updateOneWithKV [...] [...]
          else throwM e
```

Or, better: push the upsert to the database with `ON CONFLICT DO UPDATE` to eliminate the race entirely.

---

### 🟠 Warning — Dashboard Path Has No Input Validation (Asymmetry with UI Path)

**File:** `src/Domain/Action/Dashboard/Fleet/Driver.hs:4523-4529`

The UI path (`updateDriver`) validates:
- `nomineeRelationship` against an allowlist `["Spouse", "Daughter", "Son", "Mother", "Father"]`
- `nomineeDob` / `nomineeAge` within the range 18–99

The dashboard path (`postDriverFleetDriverUpdate`) calls `upsertNomineeAddress` with `req.address` and `req.addressDocumentType` but performs **none of these checks**. A dashboard admin can write any arbitrary relationship string or a DOB corresponding to age 5 into the new `driver_identity_info` table, bypassing all business rules.

**Fix:** Extract the validation into a shared helper and call it from both paths:

```haskell
-- SharedLogic/DriverIdentityInfo.hs
validateNomineeInput :: (MonadFlow m) => Maybe Text -> Maybe Day -> Day -> m ()
validateNomineeInput mbRel mbDob today = do
  whenJust mbRel $ \rel ->
    unless (rel `elem` validRelationships) $
      throwError (InvalidRequest $ "Invalid nominee relationship. Must be one of: " <> T.intercalate ", " validRelationships)
  whenJust mbDob $ \dob ->
    unless (let age = dobToAge today dob in age >= 18 && age <= 99) $
      throwError (InvalidRequest "Nominee age must be between 18 and 99")
  where
    validRelationships = ["Spouse", "Daughter", "Son", "Mother", "Father"]
```

---

### 🟠 Warning — Error Messages Don't Include Valid Options or Actual Value

**File:** `src/Domain/Action/UI/Driver.hs:1467-1474`

**Issue 1 — relationship error hides the allowlist:**
```haskell
throwError (InvalidRequest "Invalid nominee relationship")
```
The client receives `"Invalid nominee relationship"` but has no idea which values are accepted. Compare to other errors in the same file which include the dynamic context:
```haskell
InvalidRequest $ "Can't downgrade from " <> show vehicle.variant
```

**Fix:**
```haskell
throwError (InvalidRequest $
  "Invalid nominee relationship '" <> rel <> "'. Must be one of: Spouse, Daughter, Son, Mother, Father")
```

**Issue 2 — age error is ambiguous on bounds and gives wrong message for future DOBs:**
```haskell
throwError (InvalidRequest "Nominee age must be between 18 and 99")
```

- Does not say "inclusive" — is 18 valid? Is 99 valid?
- If the client submits `nomineeDob = "2099-01-01"`, the computed age is negative, but the error says "must be between 18 and 99" rather than "nominee date of birth cannot be in the future."

**Fix — add a future-date check first, then give a precise bounds message:**
```haskell
whenJust mbNomineeDob $ \dob -> do
  when (dob > utctDay nowT) $
    throwError (InvalidRequest "Nominee date of birth cannot be in the future")
  let age = DIInfo.dobToAge (utctDay nowT) dob
  unless (age >= 18 && age <= 99) $
    throwError (InvalidRequest $ "Nominee age must be between 18 and 99 (inclusive), got " <> show age)
```

---

### 🟠 Warning — Integration Test Accepts HTTP 500 for Validation Cases

**File:** `dev/integration-tests/collections/DriverProfileNomineeFlow/01-DriverProfileNomineeFlow.json:891,925`

```javascript
// Both validation test cases do this:
pm.expect(pm.response.code).to.be.oneOf([400, 500]);
```

A 500 means an unhandled server exception. Accepting it in a validation test masks exactly the scenario described in the findings above (race condition producing a raw DB exception, or a coding error turning a validation path into an internal crash). Both validation cases should assert specifically `400`:

```javascript
pm.expect(pm.response.code).to.equal(400);
```

---

### 🟡 Potential Issue — Stale Error-Handling Comment in `IffcoTokioInsurance`

**File:** `src/SharedLogic/IffcoTokioInsurance.hs` (module docstring, not changed by the PR)

The existing comment reads:
```
-- or (3) nomineeName / nomineeRelationship is missing in driver_information.
```

After the PR's change, these fields are read from `driver_identity_info` (with fallback from `driver_information`). The comment now refers to the wrong table, which could mislead the on-call engineer diagnosing a missed insurance trigger.

**Fix:** Update the comment:
```haskell
-- or (3) nomineeName / nomineeRelationship is missing from both driver_identity_info
--        and the legacy driver_information fallback.
```

---

### 🟡 Potential Issue — `deriveState` Failure Is Completely Silent

**File:** `src/SharedLogic/DriverIdentityInfo.hs:519-523`

When no Indian state is found in the address string, `deriveState` returns `Nothing` with no log, no metric, no warning. The `state` field is silently omitted. If `state` is later used for routing, eligibility, or insurance, missing state will manifest far from the upsert call and be hard to trace.

**Fix:** Add a single `logDebug` or `logInfo` so the silence is intentional and observable:

```haskell
upsertNomineeAddress ... = do
  ...
  let st = deriveState mergedAddress
  when (isJust mergedAddress && isNothing st) $
    logInfo $ "upsertNomineeAddress: could not derive state from address for driver=" <> driverId.getId
  let row = DII.DriverIdentityInfo { ..., state = st, ... }
```

---

### 🟡 Potential Issue — Legacy Invalid Values Are Never Re-validated

**File:** `src/Domain/Action/UI/Driver.hs:1468`

Validation only checks the **request** value of `nomineeRelationship`, not the stored/merged result:

```haskell
whenJust req.nomineeRelationship $ \rel -> ...  -- only fires when client sends a new value
```

Drivers who have an old value of `nomineeRelationship = "Friend"` stored in `driver_information` (written before this validation was added) will have that value migrated into `driver_identity_info` by `upsertNomineeAddress` without ever being challenged. The invalid value persists indefinitely until the driver explicitly re-sets their relationship.

This is acceptable for a backward-compatible migration, but should be documented — particularly for the insurance trigger which checks `isNothing nomineeRelationship`.

---

### ✅ What's Done Correctly

- `InvalidRequest` with HTTP 400 (`E400`) is the correct error type for these validation cases — it matches the pattern used throughout Driver.hs.
- The Postman test assertions check `errorCode == 'INVALID_REQUEST'` and specific message substrings — these match the actual thrown errors.
- `fromMaybeM DriverInfoNotFound` is correctly used for the `driverInfo` fetch in all three changed files.
- Async errors in `IffcoTokioInsurance` (`Left err`) are properly caught, logged, and the insurance record is marked `FAILED` rather than left in `PENDING`.
- `logInfo` for the "skipping insurance" case correctly identifies the driver ID — easy to grep in production logs.

bytedex (commented):
@ClaudeWeb re-review this pr, sovled the before mentioned critical

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
```diff
@@ -4522,9 +4523,8 @@ postDriverFleetDriverUpdate merchantShortId opCity driverId requestorId req = do
         DP.DRIVER -> do
           driverInfo <- QDriverInfo.findById personId >>= fromMaybeM DriverInfoNotFound
           let dob = fmap (\d -> UTCTime d 0) req.dob <|> driverInfo.driverDob
-              address = req.address <|> driverInfo.address
-              addressDocumentType = castAddressDocumentType <$> req.addressDocumentType <|> driverInfo.addressDocumentType
-          QDriverInfo.updateDriverDobAndAddress dob address addressDocumentType personId
+          DIInfo.upsertDriverIdentityInfo personId driver.merchantId driver.merchantOperatingCityId driverInfo Nothing Nothing Nothing req.address (castAddressDocumentType <$> req.addressDocumentType) Nothing
+          QDriverInfo.updateDriverDobAndAddress dob Nothing Nothing personId
```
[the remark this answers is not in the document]
bytedex: driver_information has no state column → nothing is lost by nulling legacy fields

khuzema786:
Let me fetch the latest diff and comments from the PR to see what's been fixed.The PR has only one commit (force-pushed/amended on Jun 11). Let me pull the complete current diff to diff-against-the-previous-version.Good — now I have a clear picture of what changed. Let me read the key changed sections directly.Now I have a complete picture of the changes. Let me write the re-review.

---

## PR #15056 — Re-Review (Post-Fix Pass)

**Progress: 3 of 5 criticals fixed. Two issues remain — one carry-over (TOCTOU race) and one regression introduced by this round of changes.**

---

### ✅ Confirmed Fixed

| # | Issue | Status |
|---|---|---|
| Data loss — unconditional `Nothing Nothing Nothing Nothing` wipe | **Fixed** — `legacyAddress`/`legacyNomineeName` fallback to `driverInfo.*` when `nomineeOrAddressChanged = False` |
| Dashboard path writes old table BEFORE new table | **Fixed** — `upsertDriverIdentityInfo` now runs before `updateDriverDobAndAddress` |
| `createdAt` was being reset on every update | **Fixed** — `createdAt = maybe now (.createdAt) mbExisting` correctly preserves it |

Good fixes. The data-loss correction specifically is the most important and it's clean.

---

### 🔴 Regression — Input Validation Completely Removed

**File:** `src/Domain/Action/UI/Driver.hs`, `updateDriver`

The previous version validated `nomineeRelationship` against an allowlist and `nomineeDob`/`nomineeAge` against an 18–99 age range. The updated PR **removes all of that validation** — no `throwError` for any nominee field anywhere in the new code.

Consequence: a driver can now persist `nomineeRelationship = "Dog"` or `nomineeDob = "1890-01-01"` (134 years old). The insurance trigger (`IffcoTokioInsurance`) will silently submit this data upstream to IffcoTokio with no rejection at the API boundary.

**Fix:** Restore validation, now targeting the renamed field:

```haskell
let validRelationships = ["Spouse", "Daughter", "Son", "Mother", "Father"]
whenJust req.nomineeRelationship $ \rel ->
  unless (rel `elem` validRelationships) $
    throwError (InvalidRequest $ "Invalid nominee relationship '" <> rel
      <> "'. Must be one of: " <> T.intercalate ", " validRelationships)
whenJust req.nomineeDob $ \dob -> do
  when (dob > utctDay nowT) $
    throwError (InvalidRequest "Nominee date of birth cannot be in the future")
  let age = dobToAge (utctDay nowT) dob
  unless (age >= 18 && age <= 99) $
    throwError (InvalidRequest $ "Nominee age must be between 18 and 99, got " <> show age)
```

This validation should also be added to `postDriverFleetDriverUpdate` (dashboard path still has none).

---

### 🔴 Carry-Over — TOCTOU Race Still Present

**File:** `src/SharedLogic/DriverIdentityInfo.hs`, `upsertDriverIdentityInfo`

```haskell
mbExisting <- QDII.findByDriverId driverId   -- read
...
case mbExisting of
  Just _ -> QDII.updateByPrimaryKey row      -- path A
  Nothing -> QDII.create row                 -- path B  ← two concurrent threads both land here
```

This is the exact same check-then-act race as before, just moved from `DriverIdentityInfoExtra.hs` into `SharedLogic`. Two concurrent profile updates for the same driver both observe `isNothing mbExisting`, both call `QDII.create`, the second gets a PK constraint violation, and the request dies with a raw DB exception (HTTP 500, leaks internal table names).

The fix requires either:
- A database-level `INSERT ... ON CONFLICT DO UPDATE` (if the KV layer supports it), or
- Catching the duplicate-key exception and retrying as an update, or
- Wrapping the read-modify-write in a per-driver Redis lock (pattern already used elsewhere in this codebase for similar races)

---

### 🟠 Warning — Postman Tests Reference Removed Fields

**File:** `dev/integration-tests/collections/DriverProfileNomineeFlow/01-DriverProfileNomineeFlow.json`

The tests reference two fields that no longer exist in the API:

**1. `nomineeAge` in the request body (lines ~37573, ~39660):**
```javascript
"raw": "{\"nomineeAge\": 30}"
"raw": "{\"nomineeAge\": 10}"
```
`nomineeAge` was removed from `UpdateDriverReq`. This JSON will be silently ignored by the Haskell decoder — the test sends it but the server ignores it. The "age round-trips ~30" test will check `d.nomineeAge` in the response, which will be `undefined`, causing:

**2. `nomineeAge` in response assertions (lines ~34835, ~35829):**
```javascript
pm.expect(d.nomineeAge).to.be.a('number')   // fails — field not in response
pm.expect(d.nomineeAge).to.be.within(18,99) // fails — undefined
```
`nomineeAge` was also removed from `DriverEntityRes`/`DriverInformationRes`. The integration tests will fail on any environment where they run.

**Fix:** Remove the `nomineeAge` request/response assertions or update them to use `nomineeDob` instead. If age-based input is still needed as a convenience, re-add `nomineeAge :: Maybe Int` to `UpdateDriverReq` with a server-side `ageToDob` conversion (as it was in the previous iteration).

---

### 🟠 Warning — `IndianState` Stored as `text` Needs Verified Serialization

**Files:** `src-read-only/Storage/Beam/DriverIdentityInfo.hs`, `dev/migrations-read-only/driver_identity_info.sql`

The Beam row type has:
```haskell
addressState :: B.C f (Kernel.Prelude.Maybe Kernel.Types.Beckn.Context.IndianState)
```
The SQL column is:
```sql
ALTER TABLE ... ADD COLUMN address_state text;
```

`IndianState` is a Haskell enum. For the Beam layer to correctly read/write it as `text`, it needs `FromField`/`ToField` instances for PostgreSQL, typically provided by `mkBeamInstancesForEnum ''IndianState` or similar TH. If those instances don't exist for `IndianState`, every read of a row that has a non-NULL `address_state` will crash at runtime with a deserialization error.

**Action:** Confirm `IndianState` has Beam/PostgreSQL serialization instances, or add them. Check whether `IndianState` is already stored as `text` in any other table in this codebase.

---

### 🟡 Minor — Redundant Double Fallback in `IffcoTokioInsurance`

**File:** `src/SharedLogic/IffcoTokioInsurance.hs`

```haskell
identityInfo <- DIInfo.getIdentityInfo (cast driverId) driverInfo
let nomineeName = identityInfo.nomineeName <|> driverInfo.nomineeName
    nomineeRelationship = identityInfo.nomineeRelationship <|> driverInfo.nomineeRelationship
```

`getIdentityInfo` already returns `nomineeName` as `(mbInfo >>= .nomineeName) <|> driverInfo.nomineeName` — the fallback is baked in. The `<|> driverInfo.nomineeName` here is redundant. It won't cause a bug, but it's misleading and suggests the author wasn't sure whether `getIdentityInfo` already handles the merge.

**Fix:**
```haskell
let nomineeName = identityInfo.nomineeName
    nomineeRelationship = identityInfo.nomineeRelationship
```

---

### Summary Scorecard

| Finding | This Round |
|---|---|
| Data loss (unconditional null) | ✅ Fixed |
| Dashboard path ordering | ✅ Fixed |
| `createdAt` reset on update | ✅ Fixed |
| TOCTOU race condition | ❌ Still present |
| Input validation (relationship, age) | ❌ Regression — was working, now removed |
| Integration tests reference removed fields | 🆕 New break |
| `IndianState` → `text` serialization | 🆕 Needs verification |
| Redundant double fallback in IffcoTokio | 🆕 Minor |

The data-loss fix is solid. Restore the input validation (was correctly implemented before) and handle the TOCTOU race — then this is ready for another pass.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1152,6 +1157,7 @@ buildDriverEntityRes ::
   Plan.ServiceNames ->
   m DriverEntityRes
 buildDriverEntityRes (person, driverInfo, driverStats, merchantOpCityId) merchant serviceName = do
+  identityInfo <- DIInfo.getIdentityInfo person.id driverInfo
```
0utkarsh: This extra table query can be a query param? We can have param like `identityInfoRequired`. If true, query and send

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/IffcoTokioInsurance.hs
```diff
@@ -76,8 +77,9 @@ triggerIffcoTokioInsurance driverId merchantId merchantOpCityId = do
     unless alreadyInsured $ do
       person <- QPerson.findById (cast driverId) >>= fromMaybeM (PersonNotFound driverId.getId)
       driverInfo <- QDI.findByPrimaryKey (cast driverId) >>= fromMaybeM DriverInfoNotFound
-      let nomineeName = driverInfo.nomineeName
-          nomineeRelationship = driverInfo.nomineeRelationship
+      identityInfo <- DIInfo.getIdentityInfo (cast driverId) driverInfo
+      let nomineeName = identityInfo.nomineeName <|> driverInfo.nomineeName
```
0utkarsh: Above query itself falls back to driverInfo so we can just write `nomineeName = identityInfo.nomineeName`

bytedex:
> The previous version validated `nomineeRelationship` against an allowlist and `nomineeDob`/`nomineeAge` against an 18–99 age range. The updated PR **removes all of that validation** — no `throwError` for any nominee field anywhere in the new code.
> 
> Consequence: a driver can now persist `nomineeRelationship = "Dog"` or `nomineeDob = "1890-01-01"` (134 years old). The insurance trigger (`IffcoTokioInsurance`) will silently submit this data upstream to IffcoTokio with no rejection at the API boundary.

@ClaudeWeb this is expected for now

khuzema786:
Understood — validation being deferred is an intentional decision. Updating the scorecard accordingly:

---

## PR #15056 — Updated Scorecard

| Finding | Status |
|---|---|
| Data loss (unconditional null) | ✅ Fixed |
| Dashboard path ordering | ✅ Fixed |
| `createdAt` reset on update | ✅ Fixed |
| Input validation (relationship, age) | ✅ Accepted — deferred by design |
| **TOCTOU race condition** | ❌ Still open |
| **Integration tests reference removed fields** | ❌ Still open (tests will fail) |
| **`IndianState` → `text` serialization** | ❌ Needs verification |
| Redundant double fallback in IffcoTokio | 🟡 Minor |

The two things still worth resolving before merge:

1. **TOCTOU in `upsertDriverIdentityInfo`** — `findByDriverId` → `create`/`updateByPrimaryKey` is non-atomic. Two concurrent updates for the same driver will cause a raw DB 500. Wrapping with a per-driver Redis lock or catching the PK violation and retrying as an update would fix it.

2. **Postman tests are broken** — `{"nomineeAge": 30}` is silently ignored (field removed from `UpdateDriverReq`) and assertions on `d.nomineeAge` in the response will fail since the field is no longer in `DriverEntityRes`. These tests need to be updated to reflect the new API shape.

3. **`IndianState` serialization** — confirm `IndianState` has Beam `FromField`/`ToField` instances for `text`. If it doesn't, any row with a non-NULL `address_state` will crash on read.

bytedex:
> ### 🔴 Carry-Over — TOCTOU Race Still Present
> **File:** `src/SharedLogic/DriverIdentityInfo.hs`, `upsertDriverIdentityInfo`
> 
> ```haskell
> mbExisting <- QDII.findByDriverId driverId   -- read
> ...
> case mbExisting of
>   Just _ -> QDII.updateByPrimaryKey row      -- path A
>   Nothing -> QDII.create row                 -- path B  ← two concurrent threads both land here
> ```
> 
> This is the exact same check-then-act race as before, just moved from `DriverIdentityInfoExtra.hs` into `SharedLogic`. Two concurrent profile updates for the same driver both observe `isNothing mbExisting`, both call `QDII.create`, the second gets a PK constraint violation, and the request dies with a raw DB exception (HTTP 500, leaks internal table names).
> 
> The fix requires either:
> 
> * A database-level `INSERT ... ON CONFLICT DO UPDATE` (if the KV layer supports it), or
> * Catching the duplicate-key exception and retrying as an update, or
> * Wrapping the read-modify-write in a per-driver Redis lock (pattern already used elsewhere in this codebase for similar races)

This is expected because we are trusting the system and we don't want multiple association in edge case

0utkarsh (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/DriverIdentityInfo.yaml
```diff
@@ -0,0 +1,36 @@
...
+    addressState: Maybe IndianState
+    merchantId: Id Merchant
+    merchantOperatingCityId: Id MerchantOperatingCity
+    createdAt: UTCTime
+    updatedAt: UTCTime
+
+  constraints:
+    driverId: PrimaryKey
+
+  extraOperations:
```
0utkarsh: Why extra file?

0utkarsh (approved):

PR #15082, opened 2026-06-08
rider-app/feat: BHIM PWA Journey partner-auth facade (verify token + fetch user)

dayanidhicse:
## What

Implements the **BHIM "PWA Journey" auth facade** — **Verify Token + Fetch User Details only** (payment stays on Juspay; no transaction/offline-status, deep-link, or partner-secret-key path).

Built as a **generic, provider-pluggable partner embedded-auth integration** (per review), with **BHIM as the first provider** and generic naming — BHIM-specific code is confined to `lib/external/src/PartnerAuth/BHIM/*`.

**Endpoint:** `POST /v2/partner/{provider}/session`
Request `{ token }` → response `{ isValid, sessionToken?, mobileNumber?, name? }`

Flow: parse `provider` → load its DB config (`MerchantServiceConfig`) → **verify token** (S2S, AES) → **fetch user details** (S2S, AES) → resolve/create the NammaYatri user by mobile number and issue an NY session **reusing the exact `/auth/signature` path**, under the **NAMMA_YATRI** merchant.

## How

- **`lib/external/src/PartnerAuth/`** — `PartnerAuthService` enum, `PartnerAuthServiceConfig` sum, a `case`-dispatch `Interface` exposing `verifyToken` / `getUserDetails` (mirrors `ChatCompletion`). BHIM provider: AES-256-CBC **random-IV** envelope (`encryptEnvelope`/`decryptEnvelope`, single home for mode/IV/padding), `partnerID`-header S2S client, and error-body (`errorCode`) handling → logged + mapped to `isValid:false`.
- **DB config** via `MerchantServiceConfig` (`service_name = PartnerAuth_BHIM`); AES key stored as `EncryptedField` (encrypted at rest), decrypted per call. Provider selected from the request path.
- **Session reuse:** new **additive** `resolveAndIssueDirectSession` in `Registration.hs` (plaintext-mobile entry point into the signature flow's person-resolution + DIRECT token issuance). `mobileSignatureAuth` and existing flows are **untouched**.
- **Never a 500:** the whole interaction is wrapped in `try @SomeException`; invalid token / unknown provider / any error → `200 {isValid:false}` + `logError`.

## Verification

- `cabal build all` passes under **`-Werror`** (whole monorepo).
- AES round-trip unit tests pass (`rider-app-test`): round-trip, random-IV differs + both decrypt, rejects non-32-byte key.

## Follow-ups before prod (flagged)

- **Real secrets** — `partnerId`, the encrypted `aesKey`, and UAT/prod base URLs in the seed migration (`dev/feature-migrations/0010-…`) are placeholders.
- **AES assumption** (AES-256-CBC, random 16-byte prepended IV, PKCS7, base64 over `IV‖ct`) is unconfirmed — `BHIM/Encryption.hs` isolates it to one file; a BHIM sample-vector interop test is stubbed (`TODO`) pending the real key.
- **Runtime smoke** (valid token → working `sessionToken` for getProfile) needs a real/mock BHIM + running stack; the reference mock (`ny-react-native` `web/mocks/bhim/server.js`) wasn't available in this repo, so the facade was built from the written spec.

## Release Notes

* **New Features**
  * Added partner authentication integration supporting BHIM provider.
  * New endpoint `POST /partner/{provider}/session` enables seamless user authentication through partner platforms.
  * Partner-authenticated users receive session tokens and direct app access without OTP verification.
  * Added AES encryption for secure partner communication.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/package.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/package.yaml
@@ -233,6 +233,7 @@ tests:
     dependencies:
       - rider-app
       - mobility-core
+      - external
       - tasty
       - tasty-hunit
       - tasty-quickcheck
--- a/Backend/app/rider-platform/rider-app/Main/package.yaml
+++ b/Backend/app/rider-platform/rider-app/Main/package.yaml
@@ -191,6 +191,7 @@ library:
     - finance-kernel
     - utils
     - external
+    - safe-exceptions
 
 executables:
   rider-app-exe:
--- a/Backend/app/rider-platform/rider-app/Main/rider-app.cabal
+++ b/Backend/app/rider-platform/rider-app/Main/rider-app.cabal
@@ -75,6 +75,7 @@ library
       API.Types.Dashboard.AppManagement.OrphanInstances.MerchantOnboarding
       API.Types.Dashboard.AppManagement.OrphanInstances.TicketDashboard
       API.Types.Dashboard.AppManagement.OrphanInstances.TransitOperator
+      API.Types.UI.PartnerAuthExtra
       API.Types.UI.PartnerBookingStatementExtra
       API.Types.UI.Payment
       API.UI
@@ -306,6 +307,7 @@ library
       Domain.Action.UI.NyRegularSubscription
       Domain.Action.UI.Offers
       Domain.Action.UI.ParkingBooking
+      Domain.Action.UI.PartnerAuth
       Domain.Action.UI.PartnerBookingStatement
       Domain.Action.UI.PartnerOrganizationFRFS
       Domain.Action.UI.Pass
@@ -782,6 +784,7 @@ library
       Tools.Metrics.BAPMetrics.Types
       Tools.MultiModal
       Tools.Notifications
+      Tools.PartnerAuth
       Tools.Payment
       Tools.Payout
       Tools.Schema
@@ -870,6 +873,7 @@ library
       API.Action.UI.NyRegularSubscription
       API.Action.UI.Offers
       API.Action.UI.ParkingBooking
+      API.Action.UI.PartnerAuth
       API.Action.UI.PartnerBookingStatement
       API.Action.UI.Pass
       API.Action.UI.PassDetails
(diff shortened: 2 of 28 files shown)


dayanidhicse:
Thanks for the review — addressed the actionable items (pushed in `5b989089a5`):

**Fixed**
- ✅ **PII in logs** (`Interface/BHIM.hs`): no longer log decrypted bodies on error paths; log only the endpoint label + `errorCode` (user-details bodies carry name/mobile).
- ✅ **Seed default city** (`0013-…sql`): the `SELECT` is now constrained to the merchant's default city (`moc.city = m.default_city`), matching the runtime lookup.
- ✅ **Seed `ON CONFLICT`**: retargeted to the city-aware PK `(service_name, merchant_operating_city_id)` per `1241-add-merchant-operating-city-in-merchant-service-config.sql`.
- ✅ **`-Wwarn=unused-imports` pragma** (`Domain/Action/UI/PartnerAuth.hs`): removed — all imports are used.
- ✅ **Hardcoded `+91`**: documented that BHIM is an India-only (UPI) app; to be made configurable when a non-India provider is onboarded.
- ➕ Renamed the migration `0010 → 0013` (the rebase pulled in an upstream `0010-invoice-template-lynx-helsinki.sql`).

**Intentionally kept (reasoning)**
- **`try @_ @SomeException`**: mirrors the established `Tools/EventTracking.hs` pattern for the same swallow-and-log purpose. Switching to `Control.Exception.Safe.try` (sync-only) means adding `safe-exceptions` to the library deps — happy to do it if you'd prefer that over consistency.
- **`getOneConfig` with `serviceName`**: the *current* `Tools/EventTracking.hs` uses the identical `getOneConfig { serviceName = Just … }` pattern, so this matches the in-repo convention (the cited learning about filtering locally doesn't reflect current code).
- **IP fraud-gate (`isIPBlocked`)**: the facade reuses the `/auth/signature` path (`mobileSignatureAuth`), which itself does **not** call `SMC.isIPBlocked` — only the OTP `auth` flow does. Adding it requires plumbing the client IP (`x-forwarded-for`) + regenerating the API spec; reasonable as a follow-up if desired.
- **`hlint` redundant `('[JSON])`**: that's in generated `src-read-only/` code (NammaDSL output) — present in every generated action and not hand-editable per project rules.

> Note: local build verification is currently blocked by transient GitHub 504s on the `shared-kernel`/nixpkgs source tarballs (infra, not code) — relying on CI for the authoritative `-Werror` build.

PR #15087, opened 2026-06-08
backend/fix/fix-pkey-violation-for-vendor-fee

dushyant-2002:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Summary

Cancellation penalty share was silently dropped from `VendorFee` in the EOD `CalculateDriverFees` job, so the cancellation amount that should have gone to the vendor stayed with the marketplace at split settlement.

## Root cause

`makeVendorFeeForCancellationPenalty` called `QVF.create` blindly. When ride-end had already written a `VendorFee` row for the same `(driverFeeId, vendorId)`:

- KV `SETEX` overwrote the ride-end row with one holding only the cancellation amount.
- Postgres drainer's `INSERT ... ON CONFLICT DO NOTHING` swallowed the conflict.

Net: `VendorFee.amount` ended up = regular ride share only, missing the cancellation portion.

**Evidence:** `DriverFee 567cbf46...` had `cancellation_penalty_amount = ₹10`, but its `NAMMAYATRI` `VendorFee.amount = ₹8.02` (per-ride share only). Expected ≈ ₹18.02.

## Fix

Replace blind `QVF.create` with an upsert keyed on `(driverFeeId, vendorId)`:

1. Read existing `VendorFee` via `runInMasterDbAndRedis` (master read; the EOD job can race a recent ride-end write).
2. If exists → `QVF.updateAmount` summing onto existing amount.
3. If not → `QVF.create` a fresh row.

## Scope

- MANUAL parent flow only. AUTOPAY uses a freshly-generated child `driverFeeId`, so no collision is possible.
- Ride-end hot path untouched.

## Files

- `SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs` — rewrite `makeVendorFeeForCancellationPenalty`.
- `Storage/Queries/VendorFeeExtra.hs` — add `updateAmount :: Id DriverFee -> Text -> HighPrecMoney -> m ()`.

### Additional Changes

## Checklist

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
@@ -44,6 +44,7 @@ import Domain.Types.Plan (BasedOnEntity (..), PaymentMode (AUTOPAY, MANUAL), Pla
 import Domain.Types.SubscriptionConfig
 import Domain.Types.TransporterConfig (TransporterConfig)
 import qualified Domain.Types.VendorFee as DVF
+import Kernel.Beam.Functions (runInMasterDbAndRedis)
 import Kernel.External.Encryption
 import qualified Kernel.External.Payment.Interface as PaymentInterface
 import qualified Kernel.External.Payment.Interface.Types as Payment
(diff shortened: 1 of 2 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs
```diff
@@ -934,15 +935,27 @@ makeVendorFeeForCancellationPenalty ::
...
+        Nothing ->
+          QVF.create
+            DVF.VendorFee
+              { driverFeeId = driverFee.id,
+                vendorId = vendorId,
+                amount = cancellationAmount,
+                createdAt = driverFee.createdAt,
+                updatedAt = driverFee.updatedAt
+              }
+      Hedis.setExp cancellationVendorFeeGuardKey True (3600 * 24)
```
piyushKumar-1, line 958: this is not done yet, two thread can enter the alreadyApplied False case ?

dushyant-2002: this job is also thread safe so for a single driverFee its not possible that at the same time two thread read the alreadyApplied key . the dfee can come  only again for a retry in next batch but before that we'll set cancellationVendorFeeGuardKey

PR #15092, opened 2026-06-08
[fix]: IffcoTokioInsurance query always fetch from DB

Yashwanth1906:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal query retrieval mechanism for insurance data lookups. No user-facing changes to functionality.


sidsethu:
@ClaudeWeb PR Review

khuzema786:
I'll gather the PR details, diff, and comments in parallel before reviewing.Let me try the GitHub MCP tool instead.Now let me fetch the PR details and comments in parallel.Now let me look at the changed file and related query helpers in the codebase.Let me look at the signatures for both query functions to understand the semantic difference.Let me look at the `findLatestRideByDriverId` parallel to understand the canonical pattern, and also find the `Kernel.Beam.Functions` module.Now I have all the context needed for a thorough review. Let me also look at the parallel `findLatestRideByDriverId` for comparison.I have everything I need. Here is the full review.

---

## PR Review: [fix] IffcoTokioInsurance query always fetch from DB

**Verdict: Request Changes** — The core intent is correct, but the replacement function drops the DB-level `LIMIT 1`, fetching all rows for the driver before discarding extras in Haskell. One warning, one suggestion.

---

### Context

The YAML spec (`IffcoTokioInsurance.yaml:35`) correctly declares `driverId: "!SecondaryKey"`, so KV lookups by driver ID should work. The bug is that `findAllWithOptionsKV` — when given explicit ordering + limit parameters — bypasses the KV (Redis) layer and always falls through to the database. Switching to `findAllWithKVAndConditionalDB` is the right fix and exactly matches the established codebase pattern at `RideExtra.hs:1187`:

```haskell
-- RideExtra.hs:1187 — canonical pattern in this codebase
findLatestRideByDriverId (Id driverId) =
  findAllWithKVAndConditionalDB [Se.Is BeamR.driverId $ Se.Eq driverId] (Just (Se.Desc BeamR.createdAt))
  <&> listToMaybe
```

---

### ⚠️ Warning — Missing DB-level `LIMIT 1` (Confidence: 85)

**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/IffcoTokioInsuranceExtra.hs:21`

| | Old | New |
|---|---|---|
| Function | `findAllWithOptionsKV` | `findAllWithKVAndConditionalDB` |
| DB limit | `(Just 1)` → SQL `LIMIT 1` | *(no limit parameter)* |
| Application | — | `listToMaybe` discards all but first |

The old code pushed `LIMIT 1` down to the database. The new `findAllWithKVAndConditionalDB` has no limit parameter, so on the **DB path** it fetches **all** insurance rows for the driver, then `listToMaybe` throws them away on the Haskell side.

For IffcoTokio insurance, a driver realistically has very few records, so this is low-impact today. But it is a correctness regression relative to the prior behaviour, and unbounded fetches are fragile.

This pattern is already present in `RideExtra.hs:1187` for rides (where a driver can have thousands), so either:
1. This is an **accepted codebase trade-off** — in which case it should be explicitly noted, or  
2. Both sites need a `findAllWithOptionsKVAndConditionalDB`-style helper that accepts ordering + limit — if one doesn't exist yet.

**Fix (if the accepted pattern is fine here, document it):**
```haskell
-- Fetches all rows for driver from KV/DB; LIMIT applied in-memory via listToMaybe.
-- Insurance records per driver are O(1) in practice, so this is acceptable.
findLatestByDriverId (Id driverId) = do
  results <- findAllWithKVAndConditionalDB [Se.Is Beam.driverId $ Se.Eq driverId] (Just (Se.Desc Beam.createdAt))
  pure $ listToMaybe results
```

**Fix (if a bounded path is preferred):** Check whether `findAllWithOptionsKVAndConditionalDB` exists in `Kernel.Beam.Functions`. If it does, use:
```haskell
results <- findAllWithOptionsKVAndConditionalDB
             [Se.Is Beam.driverId $ Se.Eq driverId]
             (Just (Se.Desc Beam.createdAt))
             (Just 1)
             Nothing
```

---

### 💡 Suggestion — Empty PR description (Confidence: N/A)

All checklist boxes are unchecked and the Description/Motivation/Testing sections are blank. At minimum the motivation should be filled in:

> **Why `findAllWithOptionsKV` always went to DB:** When ordering/limit parameters are passed, the KV layer cannot satisfy them natively and falls through to Postgres, defeating the cache. `findAllWithKVAndConditionalDB` checks KV first and conditionally reads DB only for missing entries.

This saves the next reviewer from having to reconstruct the root cause from scratch.

---

### ✅ What's correct

- `driverId` is correctly registered as `"!SecondaryKey"` (quoted) in the YAML spec — the KV secondary-key index exists and the new function will use it.
- Ordering (`Se.Desc Beam.createdAt`) is preserved, so `listToMaybe` returns the newest record.
- The change aligns with the canonical `findLatestRideByDriverId` pattern in `RideExtra.hs:1187`.
- No schema changes, no migration needed, no API-breaking changes.

sidsethu (approved):

PR #15094, opened 2026-06-08
Backend/enhc: linking vehicle to driver issues

dhruv-1010:
## Summary by CodeRabbit

* **New Features**
  * End-to-end vehicle-document discovery and verification added to driver onboarding, including per-document status, messages and image/meta handling.
* **Improvements**
  * Mandatory vehicle-document validation now runs when linking or activating a vehicle registration, earlier in the flow to produce clearer reject/pending outcomes.
  * Onboarding persistence of vehicle-document status adjusted for specific driver enablement scenarios.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
@@ -498,6 +498,7 @@ library
       SharedLogic.DriverOnboarding
       SharedLogic.DriverOnboarding.Digilocker
       SharedLogic.DriverOnboarding.Status
+      SharedLogic.DriverOnboarding.VehicleDocs
       SharedLogic.DriverPool
       SharedLogic.DriverPool.Config
       SharedLogic.DriverPool.DriverPoolData
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs
@@ -2853,10 +2853,11 @@ postDriverFleetLinkRCWithDriver merchantShortId opCity fleetOwnerId mbRequestorI
   rc <- RCQuery.findLastVehicleRCWrapper req.vehicleRegistrationNumber >>= fromMaybeM (RCNotFound req.vehicleRegistrationNumber)
   when (isNothing rc.fleetOwnerId || (isJust rc.fleetOwnerId && rc.fleetOwnerId /= Just fleetOwnerId)) $ throwError VehicleNotPartOfFleet
   unless (rc.verificationStatus == Documents.VALID) $ throwError (RcNotValid)
+  transporterConfig <- SCTC.findByMerchantOpCityId merchantOpCityId Nothing >>= fromMaybeM (TransporterConfigNotFound merchantOpCityId.getId)
+  SStatus.validateMandatoryVehicleDocsForRC transporterConfig rc
   validateFleetDriverAssociation fleetOwnerId driver.id
   isValidAssociation <- checkRCAssociationForDriver driver.id (Just rc) False
   when (not isValidAssociation) $ do
-    transporterConfig <- SCTC.findByMerchantOpCityId merchantOpCityId Nothing >>= fromMaybeM (TransporterConfigNotFound merchantOpCityId.getId)
     allLinkedRCs <- DAQuery.findAllLinkedByDriverId driver.id
     rcs <- RCQuery.findAllById (map (.rcId) allLinkedRCs)
     let validLinkedRCs = filter (\rc' -> rc'.verificationStatus /= Documents.INVALID) rcs
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs
@@ -107,6 +107,7 @@ import Kernel.Utils.Validation
 import Lib.Scheduler.JobStorageType.DB.Table (SchedulerJobT)
 import qualified SharedLogic.Analytics as Analytics
 import SharedLogic.DriverOnboarding
+import qualified SharedLogic.DriverOnboarding.VehicleDocs as SStatus
 import SharedLogic.Reminder.Helper (createReminder)
 import qualified Storage.Cac.TransporterConfig as SCTC
 import qualified Storage.CachedQueries.DocumentVerificationConfig as SCO
@@ -878,6 +879,7 @@ linkRCStatus (driverId, merchantId, merchantOpCityId) isTaxiBoothRequest req@RCS
   transporterConfig <- SCTC.findByMerchantOpCityId merchantOpCityId (Just (DriverId (cast driverId))) >>= fromMaybeM (TransporterConfigNotFound merchantOpCityId.getId)
   if req.isActivate
     then do
+      SStatus.validateMandatoryVehicleDocsForRC transporterConfig rc
       validated <- validateRCActivation isTaxiBoothRequest driverId transporterConfig rc
       when validated $ activateRC driverInfo merchantId merchantOpCityId transporterConfig now rc
     else do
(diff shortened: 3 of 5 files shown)


dhruv-1010:
@ClaudeWeb review these changes 
Block linking/activating an RC to a driver unless its mandatory vehicle documents are admin-approved — but only when transporterConfig.enableManualDocumentStatusCheck == Just True. Mandatory = fromMaybe isMandatory isMandatoryForEnabling && vehicleCategory == cat (live recompute via existing fetchVehicleDocuments, not the stored column). Applied to both link APIs.

Changes made
1. [Status.hs](vscode-webview://17pe3qenpe5dh8uo0kqoopkaf0mmlhj3ifobbi0k3jf5q232vn93/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverOnboarding/Status.hs#L330) — new exported helper validateMandatoryVehicleDocsForRC :: TransporterConfig -> RC -> Flow ():

Gated on enableManualDocumentStatusCheck.
Reuses fetchVehicleDocuments + computeAdminDocsVerificationStatus.
Exhaustive case: ADMIN_APPROVED → pass; ADMIN_REJECTED → "rejected"; ADMIN_PENDING → "not verified".
No PII in logs (uses rc.id, not registrationNo).
2. Dashboard API [Fleet/Driver.hs:2856](vscode-webview://17pe3qenpe5dh8uo0kqoopkaf0mmlhj3ifobbi0k3jf5q232vn93/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Driver.hs#L2856) (postDriverFleetLinkRCWithDriver / linkRCWithDriver) — fetch transporterConfig earlier, call the validator. Link-only handler, so always correct.

3. UI API [VehicleRegistrationCertificate.hs:881](vscode-webview://17pe3qenpe5dh8uo0kqoopkaf0mmlhj3ifobbi0k3jf5q232vn93/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs#L881) (linkRCStatus / /rc/setStatus) — added import ... Status as SStatus, call validator inside the if req.isActivate branch (so deactivation isn't blocked — drivers can't get trapped).

khuzema786:
I'll fetch the PR diff, description, and comments in parallel before diving into the code.Now let me read the relevant source files for full context.The workspace doesn't have the files locally. Let me fetch them via GitHub MCP.Let me read the saved full output and also fetch the key surrounding context from GitHub directly.Let me fetch the `fetchVehicleDocuments` signature and `computeAdminDocsVerificationStatus` to verify type correctness and the empty-list behavior.Now let me get the `computeAdminDocsVerificationStatus` definition and the `fetchVehicleDocuments` signature.Now I have enough context. Let me check one more thing — the `refreshVehicleDocsVerificationStatusForRC` context to compare with the new function's pattern.I now have all the context needed. Let me compose the review.

---

## PR Review: #15094 — Backend/enhc: linking vehicle to driver issues

**Verdict: Request changes.** The core logic is sound but there's one production-breaking bug in the new validator, one undocumented behaviour change bundled in, and a cross-city config inconsistency worth fixing before merge.

---

### 🔴 Critical

#### 1. `computeAdminDocsVerificationStatus []` returns `ADMIN_PENDING` — blocks all RC linking when no mandatory docs are configured
**Confidence: 95**
**File:** `SharedLogic/DriverOnboarding/Status.hs` (new `validateMandatoryVehicleDocsForRC`)

`computeAdminDocsVerificationStatus` has this definition:
```haskell
computeAdminDocsVerificationStatus docs
  | null docs = DDVS.ADMIN_PENDING   -- ← empty list → PENDING
  | ...
```

When `fetchVehicleDocuments` (called with `Just registrationNo` + `onlyMandatoryDocs = Just True`) returns `[]` — which happens whenever **no mandatory vehicle docs are configured** for the RC's vehicle category — the `find` branch hits `Nothing`, calls `computeAdminDocsVerificationStatus []` → `ADMIN_PENDING`, and the new function then throws:

```haskell
DDVS.ADMIN_PENDING -> throwError (InvalidRequest "Mandatory vehicle documents are not verified")
```

This permanently blocks every RC link/activate attempt for that merchant when `enableManualDocumentStatusCheck == Just True` but no mandatory doc types have been set up yet. Compare with the existing `refreshVehicleDocsVerificationStatusForRC`, which uses the **identical** `find … Nothing → computeAdminDocsVerificationStatus []` pattern but just persists the value — it never throws.

**Fix:** Treat "no mandatory docs configured" as vacuously approved — change the `Nothing` branch:
```haskell
let status =
      case find (\vehicleDoc -> vehicleDoc.registrationNo == registrationNo) vehicleDocuments of
        Just vehicleDoc -> computeAdminDocsVerificationStatus vehicleDoc.documents
        Nothing         -> DDVS.ADMIN_APPROVED  -- no mandatory docs → nothing to block on
```

---

### 🟠 Warning

#### 2. Undocumented `separateDriverVehicleEnablement` behaviour change in `statusHandler'`
**Confidence: 85**
**File:** `SharedLogic/DriverOnboarding/Status.hs` (diff lines ~601–611)

The PR description mentions only the RC-linking guard. But the diff also changes `statusHandler'`:

```diff
-    persistDocsVerificationStatuses person driverDocsForPersist vehicleDocsForPersist
+    let vehicleDocsForPersist' =
+          if transporterConfig.separateDriverVehicleEnablement == Just True && person.role == DP.DRIVER
+            then []
+            else vehicleDocsForPersist
+    persistDocsVerificationStatuses person driverDocsForPersist vehicleDocsForPersist'
```

When `separateDriverVehicleEnablement == Just True && role == DRIVER`, this passes `[]` as `vehicleDocsForPersist`, meaning `persistVehicleDocsVerificationStatuses` is called with an empty list — vehicle doc statuses are **silently never persisted** via this path. This is entirely undocumented and sits inside the `when (enableManualDocumentStatusCheck == Just True)` block, so both flags must be on for it to bite.

If intentional (e.g., "don't double-persist vehicle statuses when separate-enablement handles them independently"), it needs to be:
1. Explicitly described in the PR
2. Confirmed it doesn't leave `docsVerificationStatus` stale for vehicles

---

#### 3. `transporterConfig` passed from driver's city but doc configs fetched from RC's city
**Confidence: 80**
**File:** `SharedLogic/DriverOnboarding/Status.hs` (`validateMandatoryVehicleDocsForRC`)

The function takes a `transporterConfig` from the caller (driver's `merchantOpCityId`), but internally resolves `merchantOperatingCity` from `rc.merchantOperatingCityId` and fetches `allDocumentVerificationConfigs` from **the RC's city**:

```haskell
merchantOpCityId   <- rc.merchantOperatingCityId ...
merchantOperatingCity <- CQMOC.findById merchantOpCityId ...
allDocumentVerificationConfigs <- CQDVC.findAllByMerchantOpCityId merchantOperatingCity.id Nothing
```

In a multi-city deployment an RC registered in City A could be linked to a driver in City B — the feature flag comes from City B's config while the doc configs come from City A. This can produce either a false block (City B flag on, City A has no configs → empty list → bug #1 above) or false pass (City B flag on, City A configs say APPROVED but City B's own configs would say PENDING).

The existing `refreshVehicleDocsVerificationStatusForRC` avoids this by always deriving its `transporterConfig` from `rc.merchantOperatingCityId`. Aligning with that pattern would fix the inconsistency:

```haskell
validateMandatoryVehicleDocsForRC ::
  RC.VehicleRegistrationCertificate ->
  Flow ()
validateMandatoryVehicleDocsForRC rc = do
  merchantOpCityId <- rc.merchantOperatingCityId & fromMaybeM ...
  transporterConfig <- SCTC.findByMerchantOpCityId merchantOpCityId Nothing >>= fromMaybeM ...
  when (transporterConfig.enableManualDocumentStatusCheck == Just True) $ do
    ...
```

This would also simplify the call sites (no need to pass `transporterConfig` in).

---

### 🔵 Suggestion

#### 4. Redundant `find` after `fetchVehicleDocuments` already filters by `registrationNo`
**Confidence: 65 (potential)**
**File:** `SharedLogic/DriverOnboarding/Status.hs`

`fetchVehicleDocuments` is called with `(Just registrationNo)`, so it already returns only items for that RC. The subsequent:
```haskell
find (\vehicleDoc -> vehicleDoc.registrationNo == registrationNo) vehicleDocuments
```
is redundant — `listToMaybe vehicleDocuments` would be equivalent. This is minor but the same pattern exists in `refreshVehicleDocsVerificationStatusForRC`, so it's a copy-paste artefact. Worth cleaning up for clarity.

---

### What's Good

- Gating on `enableManualDocumentStatusCheck == Just True` is correct — proper opt-in.
- No PII (registration number) in logs, only `rc.id` — correct.
- Deactivation path in `VehicleRegistrationCertificate.hs` is correctly left ungated (drivers can't get trapped without a vehicle).
- `transporterConfig` fetch moved out of the `when (not isValidAssociation)` block in `Fleet/Driver.hs` — needed for the validator but also a correct refactor since the config is reused downstream.
- The exhaustive `case` on `ADMIN_APPROVED / ADMIN_REJECTED / ADMIN_PENDING` with no wildcard is the right style.

khuzema786 (approved):

PR #15161, opened 2026-06-11
backend/feat: cross-cloud UI request routing via merchant cloud config

vijaygupta18:
## Problem

We run two clouds (AWS + GCP) serving the same BAP/BPP apps. Clients are routed cloud-side, but a request can still land on the wrong cloud (stale client config, missed update). Today such requests are served by whichever cloud they hit; we want them served by the cloud that owns the request's merchant.

## Solution

A WAI middleware (backend safety net) that transparently forwards wrong-cloud **UI API** requests to the owning cloud and relays the response. The client sees nothing.

### Routing data
Two new **nullable** columns on `merchant` (both schemas):
- `cloud_type` — which cloud this merchant is enabled on (`AWS`/`GCP`)
- `cloud_base_url` — public base URL of this app's deployment on that cloud (path prefix included, e.g. `https://api.<domain>/dobpp`)

`NULL` (current state of every row) ⇒ middleware does nothing. **The feature is fully inert until a merchant row is explicitly configured.**

### Per-request flow (UI routes only)
1. Gate: first path segment ∈ UI tree (`v2` rider / `ui` driver, derived from the new `UIAPIPrefix` type-level constant — same symbol that mounts the Servant UI tree, so route and check cannot drift) **and** `token` header present **and** no `x-ny-forwarded-from` header.
2. Token → merchantId via the existing auth-token Redis cache (read-only; same key + `merchantIdFallback` auth uses). KV regtoken fallback on miss. No auth-cache format change.
3. merchantId → (cloudType, cloudBaseUrl) via in-mem cache (`withInMemCache`, 60s TTL; merchant-cardinality only, no token data in memory).
4. `fromMaybe podCloud merchant.cloudType /= podCloud` (pod cloud from existing `CLOUD_TYPE` env) and URL present ⇒ forward whole request (method/path/query/headers/body), relay response verbatim. Hop-by-hop headers stripped both ways; shared TLS manager; 60s timeout; no redirects.

### Safety properties
- **Loop guard**: forwarded requests carry `x-ny-forwarded-from`; a request bearing it is never forwarded again — config disagreement between clouds costs at most one extra hop, never a loop.
- **Fail-open**: any Redis/KV/config/forward error ⇒ serve locally (buffered body replayed to the local app).
- **Scope**: Beckn, internal, dashboard, webhook, partner-org traffic is structurally untouchable (path gate + token-header gate).
- **Advisory only**: real auth still runs on whichever cloud serves; stale routing data cannot create an auth bypass.
- Middleware sits inside `logRequestAndResponse'`, so forwarded requests appear in request logs; each forward also emits a log line (path, fromCloud → toCloud, status).

## Rollout
1. Deploy (no-op: columns NULL everywhere).
2. Set `cloud_type` + `cloud_base_url` for a merchant; DB replication propagates to both clouds.
3. Monitor `Forwarded` log lines — sustained volume for a merchant means client-side routing config should be fixed.

## Testing
- `cabal build rider-app dynamic-offer-driver-app` green (with `-Werror`).
- Verified against prod edge behavior: public URLs carry path prefixes (`/dobpp`, `/pilot/app`) which the gateway strips before pods — incoming prefix check matches what pods see; outgoing URL join handles prefixed `cloud_base_url` (and trailing slashes).
- Local end-to-end test (forward to a mock second cloud, loop-guard, fail-open) in progress; will update this PR with results.

## Summary by CodeRabbit

* **New Features**
  * Cross-cloud proxying: automatic routing of UI requests to the merchant’s configured cloud when applicable
  * Merchants can set optional cloud type and base URL for multi-cloud deployments

* **Infrastructure**
  * Server now uses a TLS-enabled HTTP manager for secure inter-cloud calls
  * Database migrations and storage updated to persist merchant cloud configuration settings

### DIFF
--- a/Backend/lib/utils/src/Utils/Common/CrossCloudProxy.hs
+++ b/Backend/lib/utils/src/Utils/Common/CrossCloudProxy.hs
@@ -11,8 +11,6 @@
 
  the GNU Affero General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
 -}
--- requestBody record update is the only way to re-attach a buffered body on this wai version
-{-# OPTIONS_GHC -Wno-deprecations #-}
 
 module Utils.Common.CrossCloudProxy
   ( crossCloudProxyMiddleware,
@@ -20,11 +18,11 @@ module Utils.Common.CrossCloudProxy
   )
 where
 
+import qualified Data.Aeson as A
 import qualified Data.ByteString as BS
-import qualified Data.ByteString.Lazy as BSL
-import Data.IORef (atomicModifyIORef, newIORef)
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as TE
+import qualified Data.Text.Encoding.Error as TE
 import qualified EulerHS.Runtime as R
 import Kernel.Prelude hiding (app)
 import Kernel.Types.Flow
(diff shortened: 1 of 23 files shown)


Backend/lib/utils/src/Utils/Common/CrossCloudProxy.hs
```diff
@@ -0,0 +1,118 @@
...
+      [] -> False
+
+    forward ownerCloud ownerUrl = do
+      reqBody <- Wai.strictRequestBody req
+      result <- try @_ @SomeException $ do
+        baseReq <- Http.parseRequest (T.unpack $ showBaseUrl ownerUrl)
+        let basePath = if "/" `BS.isSuffixOf` Http.path baseReq then BS.init (Http.path baseReq) else Http.path baseReq
+            fullPath = basePath <> Wai.rawPathInfo req
+            fwdHeaders =
+              filter (\(h, _) -> h `notElem` requestHopHeaders) (Wai.requestHeaders req)
```
piyushKumar-1, line 80: baki sab filter kyu krne hai ?

vijaygupta18: not needed as host needs to be changed and other are appended by network only

PR #15164, opened 2026-06-11
backend/fix: stale cache and slab-based fare-policy fix

AyushMittal42:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Fixed - 

- Stale cache during farePolicy upsert
- Slab based fare policy not coming in export

### Additional Changes

## Checklist

## Bug Fixes
* CSV export for fare policies now correctly handles multiple slab types, displaying accurate per-slab details including base fares, waiting charges, night shift charges, platform fees, and per-km rates.
* Improved cache management for fare policy operations to enhance system performance.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
@@ -2513,6 +2513,10 @@ getMerchantConfigFarePolicyExport merchantShortId opCity = do
                                 (nc, nt) = extractNight slab.nightShiftCharge
                                 (pfType, pfVal, pfCg, pfSg) = extractPlatformFee slab.platformFeeInfo
                              in (showT slab.startDistance, showT slab.baseFare, wc, wt, fw, nc, nt, pfType, pfVal, pfCg, pfSg, perKmOneWay)
+                          -- TODO: Ambulance slabs are age-banded by `vehicleAge`, but there is no CSV
+                          -- column for it, so it is not emitted here and the import side hardcodes
+                          -- vehicleAge = 0. Multi-band ambulance policies therefore do not round-trip
+                          -- faithfully. Add a vehicleAge column (export + import) to fix.
                           (_, _ : _) ->
                             let slab = ambulanceSlabsList !! min sectionIdx (length ambulanceSlabsList - 1)
                                 (wc, wt, fw) = extractWait slab.waitingChargeInfo
@@ -2936,7 +2940,6 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
               -- new record instead of caching an empty result.
               forM_ oldFareProducts CQFProduct.clearCache
               CQFProduct.clearCache fareProduct
-              CQFP.clearCacheById finalFarePolicy.id
 
               return (newErrors, newBoundedAlreadyDeletedMap)
 
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
@@ -2370,6 +2370,15 @@ getMerchantConfigFarePolicyExport merchantShortId opCity = do
                 map (\s -> (showT s.timePercentage, showT s.distancePercentage, showT s.farePercentage, showT s.includeActualTimePercentage, showT s.includeActualDistPercentage)) (NE.toList details.pricingSlabs)
               _ -> []
 
+          slabsList =
+            case farePolicy.farePolicyDetails of
+              FarePolicy.SlabsDetails details -> NE.toList details.slabs
+              _ -> []
+          ambulanceSlabsList =
+            case farePolicy.farePolicyDetails of
+              FarePolicy.AmbulanceDetails details -> NE.toList details.slabs
+              _ -> []
+
           -- InterCity details
           (perKmOneWay, perKmRoundTrip, kmPerExtraHour, perDayMaxHour, perDayMaxMins, defaultWaitDest, stateEntryPermit) =
             case farePolicy.farePolicyDetails of
@@ -2415,7 +2424,18 @@ getMerchantConfigFarePolicyExport merchantShortId opCity = do
                   length pricingSlabsList,
                   length driverExtraFeeBoundsList
                 ]
-            _ -> 1
+            FarePolicy.SlabsDetails _ ->
+              maximum
+                [ 1,
+                  length slabsList,
+                  length driverExtraFeeBoundsList
+                ]
+            FarePolicy.AmbulanceDetails _ ->
+              maximum
+                [ 1,
+                  length ambulanceSlabsList,
+                  length driverExtraFeeBoundsList
+                ]
 
           -- Build a row for a given section index
           buildRow sectionIdx =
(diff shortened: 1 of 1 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -2364,6 +2384,46 @@ getMerchantConfigFarePolicyExport merchantShortId opCity = do
...
+                          Nothing -> ("", "", "", "")
+                     in case (slabsList, ambulanceSlabsList) of
+                          (_ : _, _) ->
+                            let slab = slabsList !! min sectionIdx (length slabsList - 1)
+                                (wc, wt, fw) = extractWait slab.waitingChargeInfo
+                                (nc, nt) = extractNight slab.nightShiftCharge
+                                (pfType, pfVal, pfCg, pfSg) = extractPlatformFee slab.platformFeeInfo
+                             in (showT slab.startDistance, showT slab.baseFare, wc, wt, fw, nc, nt, pfType, pfVal, pfCg, pfSg, perKmOneWay)
+                          (_, _ : _) ->
+                            let slab = ambulanceSlabsList !! min sectionIdx (length ambulanceSlabsList - 1)
```
khuzema786, line 2521: @ClaudeWeb this PR has such issues of out of index can you find all issues review and also push a fix commit to this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -2779,7 +2839,9 @@ postMerchantConfigFarePolicyUpsert merchantShortId opCity req = do
                 foldlM
                   ( \usageCount fp -> do
                       let currentCount = Map.findWithDefault 0 fp.farePolicyId usageCount
-                      when (currentCount <= 1) $ CQFP.delete fp.farePolicyId
+                      when (currentCount <= 1) $ do
+                        CQFP.delete fp.farePolicyId
+                        CQFP.clearCacheById fp.farePolicyId
```
khuzema786: @ClaudeWeb let's remove this change else it may clear cache and no are product may exist till it is created later

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
khuzema786: I'll fetch the PR diff, description, and existing comments in parallel before diving in.Good, I have the diff. Now let me read the actual source file and the FarePolicy types to understand the full context.The file isn't in the local workspace. Let me fetch it via GitHub MCP and also get the FarePolicy type definition.Now let me fetch the full file from GitHub and the FarePolicy type definition in parallel.Now let me read the relevant section of the saved Merchant.hs file around lines 2274–2430 to get full context.Let me read more of the file, especially around the slab types and the areas before `buildRow`.Now let me get the Slabs slab type and also read more of Merchant.hs around the variable definitions.Let me look at how `buildRow` is invoked and get full context around numRows.Now I have the full picture. Let me analyze all the relevant parts of the code.

**Key findings:**

1. `buildRow` is called as `map buildRow [0 .. numRows - 1]`
2. `numRows` for `SlabsDetails` = `maximum [1, length slabsList, length driverExtraFeeBoundsList]`
3. `slabsList = NE.toList details.slabs` — NonEmpty, so always ≥ 1 element for `SlabsDetails`
4. The `_ -> 1` catch-all is replaced by two exhaustive arms for all 5 constructors of `FarePolicyDetailsD`
5. The outer variable `(baseDist, baseFareVal, waitCharge, ...)` uses `_ -> ("", "", "0", ...)` for non-Progressive/Rental/InterCity types

Here is the complete review:

---

## Review: `backend/fix: stale cache and slab-based fare-policy fix` (PR #15164)

**Verdict: No actual out-of-bounds crashes, but there are real logic correctness issues worth fixing.**

---

### Analysis of the `(!!)` Accesses — Are they OOB?

**Short answer: The `!!` accesses are runtime-safe as written**, because:

```haskell
(_ : _, _) ->
  let slab = slabsList !! min sectionIdx (length slabsList - 1)
```

- The `(_ : _, _)` pattern guarantees `slabsList` is non-empty before we ever reach `!!`
- So `length slabsList - 1 ≥ 0` always holds in this branch
- `min sectionIdx (length slabsList - 1)` clamps `sectionIdx` to a valid index

Same reasoning applies to the `ambulanceSlabsList !! ...` branch.

---

### Warning: Silent Data Repetition (Logic Bug — Confidence 85)

The `numRows` for `SlabsDetails` is:
```haskell
maximum [1, length slabsList, length driverExtraFeeBoundsList]
```

If `driverExtraFeeBoundsList` is **longer** than `slabsList` (e.g., 5 driver-fee bounds but only 2 slabs), `numRows = 5`, and `buildRow` is called for `sectionIdx ∈ [0..4]`. For `sectionIdx ∈ [2..4]`:

```haskell
slabsList !! min sectionIdx (length slabsList - 1)
-- = slabsList !! 1  (the last slab, repeated 3 times)
```

Rows 2, 3, 4 will all carry **duplicate slab data** (`startDistance`, `baseFare`, `waitingCharge`, etc.) for different `driverExtraFeeBounds` entries. This will silently produce incorrect CSV rows and could corrupt data on re-import.

The existing code for `ProgressiveDetails`, `RentalDetails`, and `InterCityDetails` has the same pattern — this is a pre-existing design choice, not something introduced here — but it's worth flagging since this is new territory for slab-based policies.

**Fix** (defensive guard, matching the style used elsewhere in `buildRow`):

```haskell
-- Instead of using the outer case-on-tuple pattern, mirror the guard style:
( rowBaseDistance, rowBaseFare, ... ) =
  case slabsList of
    [] -> case ambulanceSlabsList of
      [] -> (baseDist, baseFareVal, ...)
      _ ->
        let slab = ambulanceSlabsList !! min sectionIdx (length ambulanceSlabsList - 1)
            ...
        in (showT slab.baseDistance, ...)
    _ ->
      let slab = slabsList !! min sectionIdx (length slabsList - 1)
          ...
      in (showT slab.startDistance, ...)
```

This has identical runtime behavior but makes the guard explicit, matching the pattern used for `progressiveSections`, `rentalDistBuffersList`, and `pricingSlabsList` elsewhere in the same `buildRow`.

---

### Warning: `_ -> 1` Removal — Exhaustiveness (Confidence 90)

The removed `_ -> 1` was covering `SlabsDetails` **and** `AmbulanceDetails`. The new code adds both constructors explicitly. Since `FarePolicyDetailsD` has exactly 5 constructors (`ProgressiveDetails | SlabsDetails | RentalDetails | InterCityDetails | AmbulanceDetails`) and all are now matched, the pattern is exhaustive. ✓

However, if a new `FarePolicyDetailsD` constructor is ever added without updating this match, Haskell will produce an **incomplete-patterns warning** (caught at compile time), not a silent default. This is actually **safer** than the old `_ -> 1` catch-all. ✓

---

### Suggestion: Prefer Safe List Access (Confidence 60)

Haskell's `(!!)` is partial and will throw on negative indices. While the current guards make it safe, consider `Data.List.atMay` (from `safe`) or a local helper to make the intent clearer and prevent future regressions if the guard logic changes:

```haskell
safeIndex :: [a] -> Int -> Maybe a
safeIndex xs i = if i < 0 || i >= length xs then Nothing else Just (xs !! i)
```

---

### Cache Fix — Correct (Confidence 95)

```haskell
when (currentCount <= 1) $ do
  CQFP.delete fp.farePolicyId
  CQFP.clearCacheById fp.farePolicyId   -- ← new
CQFProduct.delete fp.id

-- ...
CQFP.clearCacheById finalFarePolicy.id  -- ← new
```

Both additions are correct:
1. Clearing cache **on delete** prevents stale reads after a fare policy is removed.
2. Clearing cache **after upsert** ensures the updated fare policy is reflected immediately.

No issues here. ✓

---

### Summary

| Finding | Severity | Confidence |
|---|---|---|
| Silent data repetition when `driverExtraFeeBounds` > slabs | Warning (logic) | 85 |
| No actual runtime OOB crashes — `min` guard is safe | N/A | 99 |
| `_ -> 1` removal is correct and more exhaustive-safe | N/A | 90 |
| Cache fix is correct | N/A | 95 |

The PR does not introduce a crash-level OOB bug. The `min sectionIdx (length list - 1)` guard is sound given the non-empty list guarantees. The data repetition issue is a pre-existing design pattern in this export function, but worth documenting or addressing for slab-based exports since slab count and driver-fee-bound count are unlikely to be intentionally aligned.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
khuzema786: As an inline review reply on this thread:

---

Agreed — this line should be removed:

```haskell
-- REMOVE this:
CQFP.clearCacheById finalFarePolicy.id
```

**Reason:** At this point in the flow, `finalFarePolicy` has been written to the DB but the associated `FareProduct` record may not yet be fully committed/visible. Evicting the cache here opens a window where any concurrent request that triggers a cache-miss will find no `FareProduct` and could either 404 or cache an empty result — the exact problem the comment just above this line warns against:

```haskell
-- new record instead of caching an empty result.
forM_ oldFareProducts CQFProduct.clearCache
CQFProduct.clearCache fareProduct
```

The `FareProduct` cache is already being cleared on the lines immediately above. The fare policy cache (`CQFP`) **does not need to be evicted here** because:

1. The fare policy was just **upserted** — it's either brand new (no stale cache entry exists) or updated (the caller should be invalidating on write, not here).
2. If a cache miss occurs between this eviction and the next read, the system may reconstruct a stale or empty state before the new `FareProduct` is associated.

The cache clear on **delete** (added earlier in the same PR) is correct and should stay:

```haskell
when (currentCount <= 1) $ do
  CQFP.delete fp.farePolicyId
  CQFP.clearCacheById fp.farePolicyId  -- ✓ safe: policy is being removed
```

But the post-upsert `clearCacheById` is premature and should be dropped.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
H-C-21, line 2850: AmbulanceDetails will go to this case, can this cause issue?

H-C-21: extra rows slabs are discarded (no merge/deduplicate logic like in InterCityDetails above)

H-C-21 (dismissed):

swami-patil-2025:
@H-C-21 I resolved conflicts of this pre-verified branch by going with the main branch's approach. 
Can you approve again?

H-C-21 (approved):

PR #15183, opened 2026-06-12
backend/fix: non-lynx invoice logic

TejaswiniT167:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added configurable, itemized ride-receipt invoice rendering with detailed fare line-items (including discounts/deductions).
  * Invoices can now display ride details: ride ID, driver name, vehicle number, pickup/drop addresses, and ride date/time.
  * Enabled via a per-city flag, with automatic fallback to the legacy invoice layout when receipt data isn’t available.
  * Expanded typed invoice line-item labels for richer fare breakdowns.
* **Bug Fixes**
  * Invoice template lookup now retries with English when a requested language template is missing.
  * Cancellation invoices now explicitly reference the associated ride.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FinanceInvoice.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FinanceInvoice.hs
@@ -2,11 +2,14 @@ module Domain.Action.UI.FinanceInvoice (getFinanceInvoicePdf) where
 
 import qualified API.Types.UI.FinanceInvoice as API
 import qualified Data.Time as DT
+import qualified Domain.Types.Booking as DRB
 import qualified Domain.Types.FareBreakup as DFareBreakup
 import Domain.Types.Invoice (InvoiceType (..))
 import qualified Domain.Types.Merchant as DM
 import qualified Domain.Types.OfferEntity as DOfferEntity
 import qualified Domain.Types.Person as DP
+import qualified Domain.Types.Ride as DRide
+import qualified Domain.Types.RiderConfig as DRC
 import Environment (Flow)
 import EulerHS.Prelude hiding (id)
 import Kernel.External.Types (Language (ENGLISH))
@@ -14,7 +17,7 @@ import Kernel.Prelude (head, listToMaybe, showBaseUrl)
 import Kernel.Types.Id
 import Kernel.Utils.Common
 import Lib.ConfigPilot.Interface.Types (getConfig)
-import Lib.Finance.Domain.Types.Invoice (InvoiceStatus (..))
+import Lib.Finance.Domain.Types.Invoice (Invoice, InvoiceStatus (..))
 import Lib.Finance.Invoice.PdfService
 import qualified Lib.Finance.Invoice.RenderTemplate as FRT
 import qualified Lib.Finance.Storage.Queries.IndirectTaxTransaction as QIndirectTaxExtra
@@ -25,14 +28,13 @@ import SharedLogic.Booking (getfareBreakups)
 import qualified SharedLogic.Finance.RideReceipt as RideReceipt
 import qualified SharedLogic.RenderInvoiceFromTemplate as RIFT
 import Storage.Beam.Payment ()
+import qualified Storage.CachedQueries.Merchant as CQM
 import qualified Storage.CachedQueries.Merchant.RiderConfig as CQRC
 import Storage.ConfigPilot.Config.RiderConfig (RiderConfigDimensions (..))
 import qualified Storage.Queries.Booking as QBooking
 import qualified Storage.Queries.FareBreakup as QFareBreakup
 import qualified Storage.Queries.OfferEntity as QOfferEntity
 import qualified Storage.Queries.Person as QPerson
-import qualified Storage.Queries.QueriesExtra.BookingLite as QBookingLite
-import qualified Storage.Queries.QueriesExtra.RideLite as QRideLite
 import qualified Storage.Queries.Ride as QRide
 import Tools.Error
 import "beckn-services" Tools.InvoicePdf (generateFinanceInvoicePdf)
@@ -53,14 +55,17 @@ getFinanceInvoicePdf (mbPersonId, _) mbFrom mbInvoiceId mbInvoiceType mbLimit mb
   personId <- mbPersonId & fromMaybeM (PersonNotFound "No person found")
   person <- QPerson.findById personId >>= fromMaybeM (PersonNotFound personId.getId)
   mbRiderConfig <- getConfig (RiderConfigDimensions {merchantOperatingCityId = person.merchantOperatingCityId.getId}) (Just (CQRC.findByMerchantOperatingCityId person.merchantOperatingCityId))
+  merchant <- CQM.findById person.merchantId >>= fromMaybeM (MerchantNotFound person.merchantId.getId)
 
-  let lang = fromMaybe ENGLISH person.language
+  let flagOn = merchant.useFareBreakupLineItems == Just True
+      lang = fromMaybe ENGLISH person.language
       tz = maybe DT.utc (\rc -> DT.minutesToTimeZone (fromIntegral rc.timeDiffFromUtc `div` 60)) mbRiderConfig
       tmplLogoUrl = mbRiderConfig >>= (.invoiceConfig) >>= (.logoUrl) <&> showBaseUrl
 
-      -- Itemized ride receipt (RiderInvoiceConfig.useFareBreakupLineItems): line items
+      -- Itemized ride receipt (Merchant.useFareBreakupLineItems): line items
       -- from fareBreakup display tags; the row only supplies invoiceNumber/name, so
       -- pre-finance_invoice rides render too. tz/logo from the render city's config.
+      itemizedRideReceiptPdf :: Maybe DRC.RiderConfig -> InvoiceType -> DRide.Ride -> DRB.Booking -> Flow API.FinanceInvoicePdfResp
       itemizedRideReceiptPdf mbRenderCfg invType ride booking = do
         let renderTz = maybe DT.utc (\rc -> DT.minutesToTimeZone (fromIntegral rc.timeDiffFromUtc `div` 60)) mbRenderCfg
             renderLogoUrl = mbRenderCfg >>= (.invoiceConfig) >>= (.logoUrl) <&> showBaseUrl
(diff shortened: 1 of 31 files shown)


khuzema786 (dismissed):

khuzema786 (dismissed):

khuzema786:
@ClaudeWeb Please review this PR

ClaudeWeb:
I'll fetch the PR diff, description, and comments to review this thoroughly.Let me read the full diff and explore the relevant source files.Now let me check a few specific details in the workspace files to verify my findings.Now I have enough context to write a thorough review. Let me compile my findings.I have everything I need. Here is the full review:

---

## PR Review: `backend/fix: non-lynx invoice logic` (#15183)

**Verdict: Request Changes.** The approach is solid and the architecture is well-thought-out, but there is one confirmed bug in the line-item classifier and one consistency defect in address formatting that will affect real receipts.

---

### 🔴 Critical — Bugs

#### 1. `CGST` tag silently title-cased as "Cgst" on every receipt with platform fee or slab GST

**File:** `Backend/lib/finance-kernel/src/Lib/Finance/Invoice/LineItems.hs` (classifyTag, line ~1127)

`mkFareParamsBreakups` on the BPP side emits **both** `CGST` and `SGST` tag names (`show Enums.CGST`/`show Enums.SGST`). The classifier maps `"SGST"` → `"Taxes (GST)"` but has **no entry for `"CGST"`**, so `CGST` falls through to the unknown-tag path and renders as "Cgst" — a separate extra line item that inflates the visible total.

The same applies to the BAP side: `fare_breakup` rows include `"CGST"` (confirmed in `Beckn/ACL/Common.hs:149` and `Invoice.hs:48`).

```haskell
-- Confirmed from FareCalculator.hs:313
cgstCaption = show Enums.CGST   -- produces "CGST"
mbCgstItem = mkBreakupItem cgstCaption . mkPrice <$> det.cgst
```

**Fix:** Either merge CGST into the same `TaxesGst` bucket, or add a dedicated `CGST`/`IGST` tag and constructor:

```haskell
-- In LineItems.hs classifyTag:
"SGST"  -> charge "Taxes (GST)" TaxesGst 18
"CGST"  -> charge "Taxes (GST)" TaxesGst 18   -- ← missing; dedupe collapses SGST+CGST pair
"IGST"  -> charge "Taxes (GST)" TaxesGst 18   -- ← also missing for inter-state rides
```

The `dedupe` logic (which already collapses same-description + same-amount pairs) would handle SGST+CGST correctly once both map to the same description. For rides where SGST ≠ CGST in value (partial slabs), you may want two rows — in that case add distinct `LineItemDescription` constructors `Sgst`/`Cgst`.

---

#### 2. `fullAddress` fallback differs between BAP and BPP twins

**Files:**
- `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Finance/RideReceipt.hs:908`
- `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Finance/RideReceipt.hs:287`

```haskell
-- BAP (falls back to area — a neighbourhood string):
if null parts then fromMaybe "" a.area else T.intercalate ", " parts

-- BPP (falls back to fullAddress — the complete stored address):
if null parts then fromMaybe "" a.fullAddress else T.intercalate ", " parts
```

The BPP version is more correct. When all structured parts are absent, `a.area` gives only a neighbourhood string while `a.fullAddress` gives the complete stored address. The BAP version will silently show a less complete address for locations without structured parts (common for older location records).

**Fix:** Align the BAP fallback to match BPP:

```haskell
-- BAP SharedLogic/Finance/RideReceipt.hs:908
if null parts then fromMaybe "" a.fullAddress else T.intercalate ", " parts
```

---

### 🟡 Warning — Potential Issues

#### 3. `synthFinanceInvoice` sets `issuedByType = "BUYER"` for a customer receipt

**File:** `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Finance/RideReceipt.hs:832`

```haskell
issuedByType = "BUYER",
```

In the finance domain, the *issuer* of a ride receipt is the merchant/platform (the "seller"), not the "BUYER". Real invoices from the same flow use something like `"MERCHANT"` or `"PROVIDER"` here. If any JsonLogic invoice template switches on `issuedByType`, this synthetic invoice will render incorrectly.

Worth verifying against what real `Invoice` records carry in production for this field before the feature is enabled.

#### 4. Duplicate `invoiceOrder` values cause ambiguous sort for discount/cancellation items

**File:** `Backend/lib/finance-kernel/src/Lib/Finance/Invoice/LineItems.hs` (classifyTag, lines ~1153–1158)

```haskell
"CANCELLATION_CHARGES"       -> charge "Cancellation Charges"       CancellationCharges     24
"RIDE_CANCELLATION_CHARGES"  -> charge "Ride Cancellation Charges"  RideCancellationCharges 24  -- same as above

"RIDE_CANCELLATION_TAX"       -> charge ...   25
"BUSINESS_DISCOUNT"           -> discount ...  25  -- same
"BUSINESS_DISCOUNT_PERCENTAGE"-> discount ...  25  -- same
"PERSONAL_DISCOUNT"           -> discount ...  25  -- same
"PERSONAL_DISCOUNT_PERCENTAGE"-> discount ...  25  -- same
```

Haskell's `sortOn` is stable so the relative order of same-rank items is preserved from the input list (tag order from `mkFareParamsBreakups`). This is not a crash, but it makes the displayed order implementation-dependent. Assign distinct order values to avoid surprises when the breakup source changes.

---

### ✅ What's Good

- **Architecture**: Clean separation of BAP/BPP receipt assembly with a shared `mkLineItemsFromTaggedAmounts` in `finance-kernel`. The `isSpecialZone` flag correctly relabels waiting→pickup charges on both sides.
- **Cancellation invoice bug fix** (`referenceId = Just ride.id.getId` in `CancelRide/Internal.hs`): Correct fix — without this the BPP itemized path can never resolve the ride for a `RideCancellation` invoice.
- **Language fallback** in `InvoiceTemplate.hs`: The ENGLISH retry is a safe improvement. The guard `| lang /= ENGLISH -> lookupForLang ENGLISH` correctly avoids an infinite loop.
- **Security**: Auth checks are present on both sides (`booking.riderId == personId` on BAP; `chosenInv.issuedToId` on BPP). No hardcoded secrets.
- **Backward compat**: `useFareBreakupLineItems :: Maybe Bool` defaults to `Nothing` (feature off), so existing merchants are unaffected.
- **Dedup of ONDC v2.1.0 alias tags** (`PARKING_CHARGES`, `NIGHT_CHARGES`): handled correctly via `dedupe`.
- **`QVehicle.findById ride.driverId`** on BPP: correct — NammaYatri's vehicle table is driver-keyed.

TejaswiniT167:
Re: the review above (@ClaudeWeb) — verified each point against the codebase and the consumer app. Context that changes the verdicts: this PDF's contract is byte-parity with the in-app receipt (`getInvoiceFare` in `fareEntityHelper.ts`), so the app's behavior is the source of truth for labels/ordering.

1. **CGST → "Cgst"**: the app's `getFareType` has no `CGST` case either — it falls to the default → `getAnyFare` title-cases it, so the in-app receipt shows "Cgst" today as well (the `CGST` entries at fareEntityHelper.ts:1202/1251 are yatriSathi-only filter lists, not label mappings). Mapping `CGST` → "Taxes (GST)" here would make the PDF diverge from the app receipt. If the "Cgst" cosmetics should improve, it needs a coordinated app + backend follow-up, not a PDF-only change. (It also doesn't inflate the total — the amount is a real charge component the app includes too.)

2. **`fullAddress` fallback**: rider-app's `Domain.Types.LocationAddress` has no `fullAddress` field — the suggested fix doesn't compile on BAP. That stored-address column only exists on the BPP schema, which is why the twins differ; each side uses the best data its schema has.

3. **`issuedByType = "BUYER"`**: `Lib.Finance.FinanceM.invoiceInner` (FinanceM.hs:728) hardcodes `"BUYER"` for every real finance_invoice row, so the synthetic invoice matches production data exactly — any JL switching on this field sees identical values for real and synthesized invoices. Changing it only in the synth would introduce the inconsistency, not remove it.

4. **Duplicate `invoiceOrder` values**: faithful port — the app assigns the same duplicates (`CANCELLATION_CHARGES: 24`, all discounts `25`, fareEntityHelper.ts:154-160), and both JS `Array.sort` and Haskell `sortOn` are stable, so tie-breaking (input order) matches the app exactly. Assigning distinct values here could diverge the printed order from the in-app receipt.

No code changes made for these; happy to take any of them as coordinated follow-ups if we want to change the app behavior too.

ClaudeWeb:
Thank you for the thorough follow-up — all four findings are invalidated by the production context you've provided. Updating the verdict:

---

**Revised Verdict: Approve.**

All four reported issues were false positives once the full context is applied:

| Finding | Disposition |
|---|---|
| CGST → "Cgst" | **App parity by design.** `getFareType` title-cases CGST identically; a backend-only fix would diverge the PDF from the in-app receipt. |
| `fullAddress` fallback | **Schema-constrained by necessity.** `LocationAddress` on the BAP has no `fullAddress` column; each twin uses the best field its schema offers. The suggested fix doesn't compile on BAP. |
| `issuedByType = "BUYER"` | **Matches production.** `FinanceM.invoiceInner` hardcodes `"BUYER"` for every real invoice row; the synthetic invoice correctly mirrors that. |
| Duplicate `invoiceOrder` values | **Faithful port.** `fareEntityHelper.ts` assigns the same ties; stable sort on both sides means tie-breaking is identical to the app. |

The architecture is clean, the flag-gated rollout is safe for existing merchants, the cancellation `referenceId` fix is correct, and the ENGLISH template fallback is a sound improvement. No blocking issues remain.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/FinanceInvoice.hs
```diff
@@ -47,108 +54,193 @@ getFinanceInvoicePdf (mbPersonId, _) mbFrom mbInvoiceId mbInvoiceType mbLimit mb
...
+      | invType `elem` [Ride, RideCancellation] -> do
+        rideLite <- QRideLite.findByIdLite (Id rideId) >>= fromMaybeM (RideNotFound rideId)
+        bookingLite <- QBookingLite.findByIdLite rideLite.bookingId >>= fromMaybeM (BookingDoesNotExist rideLite.bookingId.getId)
+        unless (bookingLite.riderId == personId) $
+          throwError $ InvalidRequest "Ride does not belong to this rider"
+        mbRenderCfg <-
+          if bookingLite.merchantOperatingCityId == person.merchantOperatingCityId
+            then pure mbRiderConfig
+            else getConfig (RiderConfigDimensions {merchantOperatingCityId = bookingLite.merchantOperatingCityId.getId}) (Just (CQRC.findByMerchantOperatingCityId bookingLite.merchantOperatingCityId))
+        let flagOn = (mbRenderCfg >>= (.invoiceConfig) >>= (.useFareBreakupLineItems)) == Just True
```
khuzema786: Merchant table

PR #15328, opened 2026-06-23
backend/feat/customer-rewards-and-coupons

yashikakaushik:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Release Notes

* **New Features**
  * Added Rewards management to the dashboard (campaign create/edit, cohort create, CSV code upload, status updates, campaign details & stats).
  * Added Rewards APIs for riders (list available unlocks, claim coupons, and mark rewards as redeemed), backed by trigger-based evaluation.
* **Configuration**
  * Introduced `enableRewardsManagement` to enable/disable Rewards management.
* **Database Changes**
  * Added reward campaign/cohort/unlock tracking tables, plus a partial unique constraint to prevent duplicate active unlocks.
* **Testing**
  * Added Rewards end-to-end integration flows and unit tests.


sidsethu:
Flagging two **critical** issues from review that should block merge — both small, surgical fixes.

---

### 1. Non-atomic read-modify-write on rider stats counters (regression in cancellation-rate path)

[`SharedLogic/BehaviourManagement/CustomerCancellationRate.hs`](Backend/app/rider-platform/rider-app/Main/src/SharedLogic/BehaviourManagement/CustomerCancellationRate.hs) — `incrementRiderStatToday`:

```haskell
current <- readHashCount key field
void $ Redis.hSetExp key field (T.pack (show (current + count))) ttl
```

This is **read-then-set, not atomic**. Concurrent writers can read the same `current` and overwrite each other's increments, losing counts. The previous implementation used `SWC.incrementWindowCount` (atomic Redis `INCR` underneath).

**Why this is blocking:** `incrementRiderStatToday` is wired into `incrementCancelledCount` and `incrementAssignedCount`, which feed the customer-blocking cancellation-rate logic in `nudgeOrBlockCustomer`. Under-counting `assigned` biases the ratio upward → false-positive blocks during burst traffic. This is a hidden behavior change in an existing fraud-prevention path, not a rewards-only concern.

**Fix:** use atomic `Redis.hIncrBy key field count` (HINCRBY) and set TTL separately via `EXPIRE` on first write.

**QA must cover:** parallel ride completions / cancellations for the same rider — verify hash value equals event count; verify blocking is not triggered earlier than before.

---

### 2. Leaked Pool coupon code when `createIfNoActive` returns False

[`Domain/Action/Rewards/Consumer.hs`](Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Rewards/Consumer.hs) — `unlockOne`:

```haskell
mCode <- Coupon.claimCoupon campaign cohort rid       -- LPOP pool → ZADD inflight
case mCode of
  Just code -> do
    ...
    inserted <- QRUE.createIfNoActive unlock
    Coupon.finalizeClaim campaign cohort code         -- ZREM inflight (unconditional!)
    if inserted && sentSoFar < maxPushesPerMessage
```

If `createIfNoActive` returns `False` (duplicate detected by `findOneWithKV` OR partial-unique-index race), we still call `finalizeClaim` and remove the code from inflight without writing any `RewardUnlock`. **The code is gone from both pool and inflight — leaked forever.** Reconciliation can't recover it because no inflight entry remains.

The upfront `existingNonReclaimed` filter usually catches duplicates, but it's a best-effort check; the race window is the issue.

**Fix:** call `finalizeClaim` only when `inserted == True`. When `inserted == False` and source is Pool: `pushBackToPool` first, then `removeFromInflight`.

**QA must cover:** Pool campaign with concurrent eval events for the same rider/cohort — verify pool size decreases by exactly 1, not more.

---

Both fixes are isolated changes. The rest of the review (audit trail in `createdBy`, CSV size cap, GET writes in `/rewards`, Postgres index for `findAllByCampaign`, etc.) can land in follow-ups.

sidsethu:
Adding to the critical list — the CodeRabbit comments below are also real and need fixing. CI passing here is misleading; the workflow runs HLint + CodeRabbit only, no `cabal build`, so `-Werror` violations and missing-symbol errors slipped through. Verified each against the current branch:

### Build-blockers (won't compile under `-Werror`)

1. **Function name mismatch** — [`src/Domain/Action/Dashboard/Rewards.hs`](Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Dashboard/Rewards.hs) exports `postCampaign` / `putCampaign` / `postCampaignCohort` / `postCampaignCohortCodes` / `postCampaignStatus` / `getCampaign` / `getCampaigns` / `getCampaignStats`, but the generated handler [`src-read-only/API/Action/Dashboard/Management/Rewards.hs`](Backend/app/rider-platform/rider-app/Main/src-read-only/API/Action/Dashboard/Management/Rewards.hs) calls `postRewardsCampaign` / `putRewardsCampaign` / … (with the `Rewards` prefix). Rename the 8 domain functions to the prefixed form to match the generator's convention.

2. **Unused import in [`Domain/Action/Rewards/Coupon.hs:25`](Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Rewards/Coupon.hs#L25)** — `import Control.Exception (SomeException, try)`; neither symbol is used anywhere in the file. Delete the line.

3. **Non-existent constructors in [`Domain/Action/UI/Rewards.hs`](Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Rewards.hs)** — `toApiRedemptionTargetType` constructs `API.Internal` / `API.ExternalUrl` / `API.ExternalManual`, but `API.Types.UI.Rewards` does not define `RedemptionTargetType` — both response types embed `Domain.Types.RewardCampaign.RedemptionTargetType` directly. Delete the function and use the domain type as-is (constructors are `InApp` / `ExternalUrl` / `ExternalManual`).

4. **`Maybe`-wrapping mismatch in [`Domain/Action/UI/Rewards.hs`](Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Rewards.hs)** — domain `getRewards` / `postRewardsClaim` / `postRewardsRedeemed` take `(Id Person, Id Merchant)`, but the generated handler wraps with `Control.Lens.over _1 Just`, passing `(Maybe (Id Person), Id Merchant)`. Either change the three domain signatures to accept `Maybe (Id Person)` (standard TokenAuth pattern) or adjust the spec.

5. **Type error in [`Tools/Rewards/RedisPool.hs:86`](Backend/app/rider-platform/rider-app/Main/src/Tools/Rewards/RedisPool.hs#L86)** — `let cutoff = floor (utcTimeToPOSIXSeconds now) - fromIntegral maxAgeSeconds :: Double`. `floor` returns `Integral`; `Double` isn't `Integral`. Replace with `realToFrac (utcTimeToPOSIXSeconds now) - fromIntegral maxAgeSeconds :: Double` (matches the pattern already used in `claimFromPool`).

### Security — same severity as the two I posted earlier

6. **Missing merchant/city ownership check across all campaign-id operations** — `putCampaign`, `postCampaignCohort`, `postCampaignCohortCodes`, `postCampaignStatus`, `getCampaign`, `getCampaignStats`, `postRewardsTriggerEval` underscore-ignore `merchantShortId` / `opCity` and look the campaign up by ID with no ownership validation. Any dashboard operator who knows or guesses a valid `campaignId` can mutate campaigns belonging to another merchant/city. Add a single `loadAuthorizedCampaign` helper that verifies `campaign.merchantId == merchant.id && campaign.merchantOperatingCityId == merchantOpCity.id` and route all call sites through it.

---

CodeRabbit's two SQL findings (CR7 `ON CONFLICT` target and CR8 nullable `enable_rewards_management` in the read-only migrations file) can be ignored — the actual deployment migration `1545-rider-config-enable-rewards-management.sql` already has the correct `NOT NULL DEFAULT FALSE`, and the `ON CONFLICT DO NOTHING` behavior in the seed is intentional given the existing `UNIQUE(role_id, api_entity)` constraint.

Suggest running `cabal build all` locally before the next push — none of #1–#5 would have survived a real build. Might also be worth adding a `cabal build` job to the workflow.

sidsethu (approved):

PR #15284, opened 2026-06-18
feat(fare-product): setEnabled endpoint for airport-ops enable/disable

swami-patil-2025:
Enable/disable a vehicle for a special location across all gates/areas, trip categories, search sources and all peaks by flipping FareProduct.enabled in place (FarePolicy values preserved; duplicate rows collapsed to one). Builds clean: dynamic-offer-driver-app + provider-dashboard.

- Storage/FareProduct.yaml: findAllByMerchantOpCityIdAreaVehicle + updateFareProductEnabled
- Management API: POST /config/fareProduct/setEnabled (SetFareProductEnabledReq)
- Driver-app handler postMerchantConfigFareProductSetEnabled + dashboard forwarder
- Generated servant API / client / types / access SQL
- Fix provider-dashboard-common.dhall: add CONFIG_PILOT/_configPilot fields now required by latest namma-dsl (un-migrated dashboard config blocked codegen)

## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

- [ ] This PR modifies the database schema (database migration added)
- [ ] This PR modifies dhall configs/environment variables
- [ ] This PR contains API breaking changes

## Checklist

- [ ] I formatted the code and addressed linter errors `./dev/format-all-files.sh`
- [ ] I reviewed submitted code
- [ ] I added unit tests for my changes where possible
- [ ] I added integration tests for my changes where possible
- [ ] I tested the changes using integration tests
- [ ] I added a [CHANGELOG](/CHANGELOG.md) entry if applicable
- [ ] No leak detected in leakcanary

## Summary by CodeRabbit

* **New Features**
  * Added a merchant configuration endpoint to enable or disable fare products: `POST /config/fareProduct/setEnabled`.
  * Request supports required `areas` and `vehicleServiceTiers`, plus optional `tripCategories` and `searchSources` for targeted updates.
  * The update validates non-empty inputs, resolves duplicate fare-product records deterministically, removes orphaned fare-policy links, and refreshes cached lists immediately.
* **Permissions**
  * Added the required permission for `PROVIDER_MANAGEMENT/MERCHANT/POST_MERCHANT_CONFIG_FARE_PRODUCT_SET_ENABLED`.

PR #15388, opened 2026-06-25
feat(dashboard): fareProduct/setEnabled on prodHotPush-BPP (cherry-pick #15284)

swami-patil-2025:
## What
Cherry-picks **#15284** (squash `0467990ebd`) onto `prodHotPush-BPP` — airport-ops enable/disable of fare products via `POST /merchant/config/fareProduct/setEnabled`. The handler groups products by `(tripCategory, searchSource, timeBounds)`, keeps exactly one row per key (preferring the live one so its `FarePolicy`/fares survive a re-enable), flips `enabled` in place, collapses duplicates, and deletes only truly-orphaned fare policies.

## Why a fresh PR (not a clean cherry-pick)
`prodHotPush-BPP` diverged from `main`, so the cherry-pick conflicted and was hand-resolved:

- **spec `FareProduct.yaml`** — kept the two new queries (`findAllByMerchantOpCityIdAreaVehicle`, `updateFareProductEnabled`); dropped main-only `findAllFareProductByMerchantOpCityIdAndArea` (absent on this branch, unused by the handler).
- **generated read-only files** (`Endpoints/Merchant.hs`, the two servant `Merchant.hs`) — the setEnabled wiring was **spliced by hand** in this branch's existing format, **not** re-run through codegen.

## ⚠️ Action needed before merge
These read-only files are normally produced by `, run-generator`. PR #15284 was generated with the pinned namma-dsl which **re-parenthesizes** generated code, so canonical codegen output will differ in formatting from this hand-splice. **Please re-run `, run-generator` on this branch** (it also needs the `configPilot` dhall field, included here in `provider-dashboard-common.dhall`) to normalize, and let CI do the full build. The handler logic, spec, request type, SQL migration, and forwarders are complete and were verified to reference only functions that exist on this branch.

Draft until CI build passes / codegen is re-run.


swami-patil-2025:
### Review: `setEnabled` deletes fare products + fare policies as a side effect of a toggle

Flagging the one item I'd block merge on, independent of CI.

In `postMerchantConfigFareProductSetEnabled` ([`Domain/Action/Dashboard/Management/Merchant.hs`](https://github.com/nammayatri/nammayatri/blob/cherry/setEnabled-prodHotPush/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs)), the handler doesn't just flip `enabled` — for every group it keeps one row and **permanently deletes the rest** (`CQFProduct.delete`), then garbage-collects their `FarePolicy` if no longer referenced (`CQFP.delete`).

The grouping key is `(tripCategory, searchSource, timeBounds)` and **does not include `farePolicyId`**:

```haskell
groupKey p = (show p.tripCategory, show p.searchSource, show p.timeBounds)
```

**Concern:** if two fare products can share that key but carry *different* fare policies (different fares), then disabling — or even enabling — irreversibly drops all but one, and orphan-deletes the dropped policies. The in-code comment claims keeping one row guarantees a later re-enable can restore the config, but that only holds if the collapsed duplicates were truly identical. A `setEnabled` toggle silently destroying fare configuration would be surprising, and this is targeting a hotpush branch.

**Question for maintainers:** is there a uniqueness guarantee that `(merchantOpCity, area, vehicle, tripCategory, searchSource, timeBounds)` maps to at most one `FareProduct` (so collapse is pure dedup), or can legitimate distinct-policy rows share that key? If the latter, the delete + orphan-cleanup should be reconsidered (e.g. scope the destructive collapse out of the enable/disable path).

For context: the redundant-bracket alerts have been fixed, and a full `run-generator` on this branch's pinned namma-dsl reproduces the read-only files with no diff — so the hand-spliced wiring matches canonical codegen output. This data-loss question is the remaining substantive item.

H-C-21 (approved):

PR #15406, opened 2026-06-28
Backend/admin onboarding docs changes

FCJ6202:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a vehicle document status lookup to onboarding (by RC number or RC id).
  * Enabled optional per-document metadata in document status responses when configured (covers common driver docs, plus PUC/fitness/insurance/permit and UDYAM/TAN/LDC).
  * Added configuration-driven document upload permissions and “mark valid on validation skip”.
* **Bug Fixes**
  * Improved document approval/rejection handling, including inspection-hub rejection, TAN/UDYAM updates, and richer local residence proof support.
  * Added permit expiry to RC/vehicle registration details.


Backend/app/dashboard/CommonAPIs/spec/ProviderPlatform/Management/API/DriverRegistration.yaml
```diff
@@ -620,8 +627,12 @@ types:
   ServiceType:
     enum: "HyperVerge,Idfy"
     derive: "HttpInstance"
+  LocalResidenceProofApproveDetails:
+    - documentImageId: Id Image
+    - state: Maybe IndianState
+    - proofDocumentType: Maybe AddressDocumentType
```
0utkarsh, line 644: address?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1351,6 +1351,7 @@ TransporterConfig:
     isGstPanLinkCheckRequired: Maybe Bool
     invoiceConfig: Maybe InvoiceConfig
     enableManualDocumentStatusCheck: Maybe Bool
+    enableDocumentMetadata: Maybe Bool # When True, document metadata (DL number, Aadhaar number, PAN, address proof details) is computed and returned in status responses.
```
0utkarsh, line 1354: Why here? It should have been in query param. 
Status handler is called from multiple places, it may need these details only at some places. We should not make it as city level config

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/Image.hs
```diff
@@ -232,22 +232,32 @@ validateImageHandler isDashboard mbUploaderRole mbDocConfigs (personId, _, merch
...
-        then do
-          validationOutput <-
-            Verification.validateImage merchantId merchantOpCityId $
-              Verification.ValidateImageReq {image, imageType = castImageType imageType, driverId = person.id.getId}
-          when validationOutput.validationAvailable $ do
-            checkErrors imageEntity.id imageType validationOutput.detectedImage
-          Query.updateVerificationStatusOnlyById Documents.VALID imageEntity.id
-        else when (isNothing validationStatus) $ Query.updateVerificationStatusOnlyById Documents.MANUAL_VERIFICATION_REQUIRED imageEntity.id
+      (isImageValidationRequired, markValidOnSkip) <- case person.role of
+        Person.FLEET_OWNER -> do
```
0utkarsh: Why `FLEET_BUSINESS` removed? Can we keep this same as earlier and just return `markValidOnSkip`?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/Image.hs
```diff
@@ -232,22 +232,32 @@ validateImageHandler isDashboard mbUploaderRole mbDocConfigs (personId, _, merch
...
+              fleetDocConfigs >>= (.markImageValidOnValidationSkip)
+            )
+        _ -> do
+          let mbDocCfg = find (\c -> c.vehicleCategory == fromMaybe CAR vehicleCategory) docConfigs
+          return
+            ( maybe True (.isImageValidationRequired) mbDocCfg,
+              mbDocCfg >>= (.markImageValidOnValidationSkip)
+            )
+      case markValidOnSkip of
+        Just True -> Query.updateVerificationStatusOnlyById Documents.VALID imageEntity.id
```
0utkarsh: We should do this when idfy is skipped, right? On line 260

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverUdyamExtra.hs
```diff
@@ -12,3 +15,51 @@ import Storage.Queries.DriverUdyam ()
 findAllByEncryptedUdyamNumber :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => DbHash -> m [DriverUdyam]
 findAllByEncryptedUdyamNumber udyamNumberHash = do
   findAllWithKV [Se.Is Beam.udyamNumberHash $ Se.Eq udyamNumberHash]
+
+findValidUdyamByHashExcludingDriver :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => DbHash -> Kernel.Types.Id.Id Domain.Types.Person.Person -> m (Maybe DriverUdyam)
```
0utkarsh: Why not in DSL with status filter?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverUdyamExtra.hs
```diff
@@ -12,3 +15,51 @@ import Storage.Queries.DriverUdyam ()
...
+findValidUdyamByHashExcludingDriver udyamNumberHash driverId =
+  findOneWithKV
+    [ Se.And
+        [ Se.Is Beam.udyamNumberHash $ Se.Eq udyamNumberHash,
+          Se.Is Beam.driverId $ Se.Not $ Se.Eq (Kernel.Types.Id.getId driverId),
+          Se.Is Beam.verificationStatus $ Se.Eq Documents.VALID
+        ]
+    ]
+
+findValidUdyamByHash :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => DbHash -> m (Maybe DriverUdyam)
```
0utkarsh: Same and same for others in this file

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1383,9 +1390,137 @@ approveAndUpdateFitnessCertificate req@Common.FitnessApproveDetails {..} mId mOp
...
+      DVC.GSTCertificate -> QGstin.deleteByDriverIdAndStatus reqDriverId Documents.INVALID
+      DVC.BusinessLicense -> QBLExtra.deleteByDriverIdAndStatus reqDriverId Documents.INVALID
+      DVC.UDYAMCertificate -> QUdyamExtra.deleteByDriverIdAndStatus reqDriverId Documents.INVALID
+      _ -> pure ()
+    findDocumentByNumber docNum = case documentType of
+      DVC.DriverLicense -> fmap DLApproveData <$> QDL.findByDLNumber docNum
+      DVC.PanCard -> do
+        panHash <- getDbHash docNum
+        panList <- QPan.findAllByEncryptedPanNumber panHash
+        pure $ PanApproveData <$> (find (\pan -> pan.driverId == reqDriverId) panList <|> listToMaybe panList)
```
0utkarsh, line 1469: find doesn't look correct

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1383,9 +1390,137 @@ approveAndUpdateFitnessCertificate req@Common.FitnessApproveDetails {..} mId mOp
...
+      DVC.AadhaarCard -> do
+        aadhaarHash <- getDbHash docNum
+        aadhaarList <- QAadhaarCard.findAllByEncryptedAadhaarNumber (Just aadhaarHash)
+        pure $ AadhaarApproveData <$> (find (\aadhaar -> aadhaar.driverId == reqDriverId) aadhaarList <|> listToMaybe aadhaarList)
```
0utkarsh, line 1473: find - same for other places

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -2029,15 +2287,13 @@ handleRejectRequest rejectReq merchantId merchantOperatingCityId = do
...
-      driver <- QDriver.findById driverUdyam.driverId >>= fromMaybeM (PersonNotFound driverUdyam.driverId.getId)
-      let notifyOpCityId = fromMaybe merchantOperatingCityId driverUdyam.merchantOperatingCityId
-      void $
-        withTryCatch "CommonDocumentReject:sendRejectionNotification" $
-          sendDocumentRejectionNotification notifyOpCityId (show DVC.UDYAMCertificate) udyamRejectReq.reason driver
-      rejectAndUpdateUdyamDocument udyamRejectReq merchantOperatingCityId
+    Common.InspectionHubReject inspectionHubRejectReq -> do
+      let requestId = Id inspectionHubRejectReq.requestId :: Id DOHR.OperationHubRequests
+      request <- QOHR.findByPrimaryKey requestId >>= fromMaybeM (InternalError "Inspection hub request not found")
+      unless (request.requestStatus == DOHR.APPROVED) $
```
0utkarsh, line 2287: Why can't it reject while its Pending? And what if it rejects post Bot Approval?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverOnboarding/VehicleDocs.hs
```diff
@@ -74,6 +77,113 @@ data VehicleDocumentItem = VehicleDocumentItem
   }
   deriving (Show, Eq, Generic, ToJSON, FromJSON, ToSchema)
 
+data DLDocumentMetadata = DLDocumentMetadata
```
0utkarsh, line 81: Why these definitions are in `VehicleDocs.hs`? 
And why not through YAML?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1383,9 +1390,137 @@ approveAndUpdateFitnessCertificate req@Common.FitnessApproveDetails {..} mId mOp
...
+validateDocumentApprovalChecks :: DVC.DocumentType -> Maybe Text -> Id DP.Person -> Maybe ApproveDocumentData -> Flow ()
+validateDocumentApprovalChecks documentType mbReqDocNum reqDriverId mbExistDocData = do
+  mbDocData <- case mbExistDocData of
+    Just docData -> pure (Just docData)
+    Nothing -> maybe (pure Nothing) findDocumentByNumber mbReqDocNum
+  whenJust mbDocData $ \docData -> do
+    mbStoredNumber <- approveDocNumber docData
+    whenJust mbReqDocNum $ \reqDocNum ->
+      whenJust mbStoredNumber $ \storedNumber ->
+        when (reqDocNum /= storedNumber) $
```
0utkarsh: Previously, it was allowed to update license number

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1671,10 +1816,13 @@ approveAndUpdateAadhaar req mId mOpCityId = do
                 DAadhaar.nameOnCard = req.nameOnCard <|> aadhaar.nameOnCard,
                 DAadhaar.dateOfBirth = req.dateOfBirth <|> aadhaar.dateOfBirth,
                 DAadhaar.address = req.address <|> aadhaar.address,
+                DAadhaar.aadhaarBackImageId = mbImageId2 <|> aadhaar.aadhaarBackImageId,
```
0utkarsh, line 1814: Where is front image's id?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1383,9 +1390,137 @@ approveAndUpdateFitnessCertificate req@Common.FitnessApproveDetails {..} mId mOp
...
+  mbDriverDoc <- findValidDocumentByDriverId
+  whenJust mbDriverDoc $ \driverDoc ->
+    case mbDocData of
+      Just docData ->
+        when (approveDocDriverId driverDoc /= approveDocDriverId docData) $
+          throwError DriverAlreadyLinked
+      Nothing -> throwError DriverAlreadyLinked
+  -- Clean up the requesting driver's stale INVALID document so the approve flow can
+  -- upsert the approved row cleanly (callers must upsert, not update, after this).
+  deleteInvalidDocumentOfDriver
```
0utkarsh: Promote it to top-level handler. There can be cases where we delete invalid doc but approval also fails

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -1901,40 +2107,81 @@ updateFleetOwnerInfoOnDocApproval :: DP.Person -> (Id DP.Person -> Flow ()) -> F
...
+
+-- | On any exception from the approval action, delete the uploaded image and re-throw
+-- so the caller receives an error response. withTryCatch ensures the error is logged.
+doApproveWithRevert :: Maybe (Id Common.Image) -> Flow () -> Flow ()
+doApproveWithRevert mbImageId action = do
+  result <- withTryCatch "handleApproveRequest" action
+  case result of
+    Right () -> pure ()
+    Left err -> do
+      whenJust mbImageId $ \imgId -> QImage.deleteById (Id imgId.getId)
```
0utkarsh, line 2178: Why are we deleting image unconditionally? What if doc is linked with some other driver? Approval will throw error and image will get deleted for a valid doc

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverOnboarding/Status.hs
```diff
@@ -1307,105 +1316,167 @@ getProcessedDriverDocuments role driverId entityImagesInfo docType useHVSdkForDL
...
     DVC.LocalResidenceProof -> do
       let (status, reason, url) = checkImageValidity entityImagesInfo DVC.LocalResidenceProof
       mbIdentityInfo <- QDII.findByDriverId driverId
       let hasAddressDetails =
             maybe False (\info -> isJust info.address && isJust info.addressDocumentType && isJust info.addressState) mbIdentityInfo
-      return (if hasAddressDetails then status else Just NO_DOC_AVAILABLE, reason, url, Nothing, mbS3Path, mbImageId, Nothing)
+      mbLocalMetadata <-
+        if enableMetadata
+          then do
+            mbIdentityInfo <- QDII.findByDriverId driverId
```
0utkarsh: Already queried in line 1386

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/UdyamVerification.hs
```diff
@@ -115,14 +116,15 @@ onVerifyUdyam verificationReq output serviceName = do
           let updated =
                 driverUdyam
                   { DUdyam.verificationStatus = Documents.VALID,
+                    DUdyam.documentImageId = Just verificationReq.documentImageId1 <|> driverUdyam.documentImageId,
```
0utkarsh: Left side is already Just

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
FCJ6202: At that time, we discussed doing this config way. I'm not sure why I added this to the transporter config. But yeah, we should pass this value from query params only, else we will compute this data for every request. I will move this config

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/Image.hs
FCJ6202: Didn't get this. Because let's say in our case, imageValidation is required, then it always do verification which we don't want

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverUdyamExtra.hs
FCJ6202: Yeah, we can do that.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverUdyamExtra.hs
FCJ6202: I will delete this. Not using it anywhere right now. I think previously I used it, but in the refactor, I am not using it

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: Why?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: > Why can't it reject while its Pending

Sure, I can add these checks also, or we can throw an error when requestStatus!= rejected. This will handler all other cases.

> what if it rejects post Bot Approval?

Then, we also reject, right?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverOnboarding/VehicleDocs.hs
FCJ6202: > Why these definitions are in VehicleDocs.hs?

DocumentStatusItem is already defined in this file. That's why I added this file. 

> And why not through YAML?

On the dashboard side, we generated it through YAML. But in driver side, we need to define manually

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: We did this because while uploading the image, we make an entry as valid in the image table, and if approve api gets failed, then from register/status we still show valid, which doesn't make sense. We will discuss which condition we should add here.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/UdyamVerification.hs
FCJ6202: Need to discuss

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: ok. Means any entity uploads a document with some number L1, and while the admin tries to approve that doc and put a different number, can we update the admin passed doc number?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: We can discuss

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/Image.hs
FCJ6202: For our case, imageValidation required is false and in the else part we should put the markValidOnSkip check

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: expected

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: We shouldn't throw error. Need to change

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: We should query findByDriverId and PanHash

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: We should query findByDriverId and PanHash

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverOnboarding/VehicleDocs.hs
FCJ6202: Will refactor this in next pr

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: deleteInvalidDocumentOfDriver, We should call this before creation after all validation done

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/UdyamVerification.hs
FCJ6202: We can make image column mandatory in refactor pr

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
FCJ6202: Will refactor this properly in next pr

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
FCJ6202: Will do it in next pr

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Onboarding.hs
```diff
@@ -230,6 +317,35 @@ castDocsVerificationStatus = \case
...
+  merchant <- findMerchantByShortId merchantShortId
+  merchantOpCity <- CQMOC.findByMerchantIdAndCity merchant.id opCity >>= fromMaybeM (MerchantOperatingCityNotFound $ "merchantShortId: " <> merchantShortId.getShortId <> " ,city: " <> show opCity)
+  transporterConfig <- getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = merchantOpCity.id.getId}) (Just (SCTC.findByMerchantOpCityId merchantOpCity.id Nothing)) >>= fromMaybeM (TransporterConfigNotFound merchantOpCity.id.getId)
+  rc <- case mbRcId of
+    Just rcId -> RCQuery.findById (Id rcId) >>= fromMaybeM (InvalidRequest $ "RC not found by id: " <> rcId)
+    Nothing -> case mbRcNo of
+      Just rcNo -> do
+        rcNoEnc <- encrypt rcNo
+        RCQuery.findByCertificateNumberHash (rcNoEnc & hash) >>= fromMaybeM (InvalidRequest $ "RC not found for number: " <> rcNo)
+      Nothing -> throwError (InvalidRequest "Either rcNo or rcId must be provided")
```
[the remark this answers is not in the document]
0utkarsh: Please do this also in next PR

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/DriverRegistration.hs
```diff
@@ -2220,8 +2462,13 @@ postDriverRegistrationDocumentsUpdate _merchantShortId _opCity _req = do
       mbRcId <- getApproveTargetRcId approveReq
       processPostUpdate mbPersonIdRaw mbRcId
     Common.Reject rejectReq -> do
-      handleRejectRequest rejectReq merchant.id merchantOpCityId
       mbPersonIdRaw <- getRejectTargetPersonId rejectReq
+      whenJust mbPersonIdRaw $ \personId -> do
+        mbDriverInfo <- QDriverInfo.findById personId
+        whenJust mbDriverInfo $ \driverInfo ->
+          when driverInfo.onRide $
+            throwError (InvalidRequest "Cannot reject document for a driver who is currently on a ride")
```
0utkarsh, line 2476: We may need some checks here, should only be needed for mandatory docs

RitikaHanish:
@ClaudeWeb Can you review this PR by making sure efficient system and good design principles, nothing should break, review thoroughly ?

ClaudeWeb:
I'll fetch the PR diff, description, and comments in parallel to start the review.Let me read the full diff and explore the workspace.Let me read the rest of the diff and also explore the workspace for context.Let me read the final portions of the diff.Now I have a complete picture of the PR. Let me compile the review.Now I have everything needed for a thorough review. Here is the full analysis:

---

## PR Review: Backend/Admin Onboarding Docs Changes (#15406)

**Verdict: Request Changes** — this is a substantial, well-structured PR with solid logic, but there are three hard blocking issues: a data-destructive bug on approval failures, full PII exposure in metadata, and multiple breaking API changes without migration support.

---

### 🔴 Critical

#### 1. `doApproveWithRevert` deletes driver images on *validation* failures (Confidence: 92)

**File:** `Domain/Action/Dashboard/Management/DriverRegistration.hs`

```haskell
doApproveWithRevert :: Maybe (Id Common.Image) -> Flow () -> Flow ()
doApproveWithRevert mbImageId action = do
  result <- withTryCatch "handleApproveRequest" action
  case result of
    Right () -> pure ()
    Left err -> do
      whenJust mbImageId $ \imgId -> QImage.deleteById (Id imgId.getId)  -- ← deletes on ANY error
      throwM err
```

`withTryCatch` catches ALL exceptions — including business-logic errors like `DLAlreadyLinked`, `PanAlreadyLinked`, `DriverAlreadyLinked`, etc. thrown by `validateDocumentApprovalChecks`. If an admin enters a wrong document number or the number is already registered, the driver's correctly-uploaded image is **permanently deleted**, forcing the driver to re-upload with no explanation.

The intent (revert infrastructure failures) is good, but the scope is too broad.

**Fix:** Only delete the image for non-domain errors, or categorize errors before reverting:

```haskell
doApproveWithRevert :: Maybe (Id Common.Image) -> Flow () -> Flow ()
doApproveWithRevert mbImageId action = do
  result <- withTryCatch "handleApproveRequest" action
  case result of
    Right () -> pure ()
    Left err -> do
      -- Only revert image on infrastructure/unexpected errors, not on domain validation failures
      whenJust mbImageId $ \imgId ->
        unless (isDomainError err) $ QImage.deleteById (Id imgId.getId)
      throwM err
```

Or, more conservatively: remove automatic image deletion entirely and handle it explicitly only where it makes semantic sense (e.g., the upload-and-approve atomic flow).

---

#### 2. Full Aadhaar number exposed in API response (Confidence: 90)

**File:** `SharedLogic/DriverOnboarding/Status.hs`

```haskell
DVC.AadhaarCard -> do
  ...
  mbAadhaarMetadata <-
    if enableMetadata
      then forM mbAadhaarCard $ \aadhaar -> do
        aadhaarNumberDec <- mapM decrypt aadhaar.aadhaarNumber   -- ← full decrypted number
        pure $ AadhaarMetadata AadhaarDocumentMetadata
          { aadhaarNumber = aadhaarNumberDec, ...}  -- ← placed in API response
```

The full 12-digit Aadhaar number is decrypted and returned in the status API response. The Aadhaar Act and UIDAI guidelines prohibit transmitting or storing Aadhaar numbers in plain text. There's already masking logic in `approveAndUpdateAadhaar` (`T.replicate (T.length req.aadhaarNumber - 4) "X" <> T.takeEnd 4 req.aadhaarNumber`) — the same pattern must be applied here.

**Fix:**
```haskell
aadhaarNumberDec <- mapM decrypt aadhaar.aadhaarNumber
let maskedAadhaar = fmap (\n -> T.replicate (T.length n - 4) "X" <> T.takeEnd 4 n) aadhaarNumberDec
pure $ AadhaarMetadata AadhaarDocumentMetadata { aadhaarNumber = maskedAadhaar, ... }
```

Similarly, PAN and DL numbers in metadata should be reviewed for whether partial masking is appropriate per your data-handling policy.

---

#### 3. Three breaking API changes in `RejectDetails` / `ApproveDetails` with no migration path (Confidence: 87)

**Files:** `CommonAPIs/spec/ProviderPlatform/Management/API/DriverRegistration.yaml`, generated `.hs`

**a) UDYAM reject type renamed + field changed:**
```diff
- UDYAMReject UDYAMRejectDetails   (field: udyamId: Id DriverUdyam)
+ InspectionHubReject InspectionHubRejectDetails  (field: requestId: Text)
```

**b) UDYAM approve field changed:**
```diff
- UDYAMApprove UDYAMApproveDetails  { udyamId :: Id DriverUdyam, ... }
+ UDYAMApprove UDYAMApproveDetails  { documentImageId :: Id Image, ... }
```

**c) LocalResidenceProof approve constructor changed:**
```diff
- LocalResidenceProofImg (Id Image)
+ LocalResidenceProofApprove LocalResidenceProofApproveDetails
```

All three are polymorphic JSON sum types. Any existing admin dashboard client sending `{"UDYAMReject": {"udyamId": "..."}}` or `{"LocalResidenceProofImg": "..."}` will get a silent deserialization failure or `400` with no useful message. These need either:
- A deprecation alias keeping the old constructor alive during a transition window, **or**
- Explicit documentation that this is a breaking version bump requiring coordinated client updates before deployment.

---

### ⚠️ Warning

#### 4. Missing SQL index for `driver_udyam.document_image_id` (Confidence: 85)

**File:** `dev/migrations-read-only/dynamic-offer-driver-app/driver_udyam.sql`

```sql
ALTER TABLE atlas_driver_offer_bpp.driver_udyam ADD COLUMN document_image_id character varying(36);
-- ← no CREATE INDEX
```

The storage spec marks `documentImageId` as `SecondaryKey`, which generates an in-memory KV index, but the SQL migration has no corresponding `CREATE INDEX`. All SQL replica queries via `findByImageId` will do a full-table scan.

**Fix:**
```sql
CREATE INDEX IF NOT EXISTS idx_driver_udyam_document_image_id
  ON atlas_driver_offer_bpp.driver_udyam (document_image_id);
```

---

#### 5. `findByImageId` on `DriverUdyam` accepts `Maybe (Id Image)` — footgun (Confidence: 75)

**File:** `Storage/Queries/DriverUdyam.hs` (line ~929)

```haskell
findByImageId documentImageId = do
  findOneWithKV [Se.Is Beam.documentImageId $ Se.Eq (Kernel.Types.Id.getId <$> documentImageId)]
```

Calling with `Nothing` evaluates to `Se.Eq Nothing` which in the KV/SQL layer means `WHERE document_image_id IS NULL`. Before this migration runs, **every existing UDYAM row has `NULL` for this column**, so `findByImageId Nothing` would return an arbitrary existing UDYAM record — potentially the wrong driver's document.

The current call site `QUdyam.findByImageId (Just imageId)` is correct, but the function signature invites misuse.

**Fix:** Change the signature to accept `Id Image` (non-optional):
```haskell
findByImageId :: (EsqDBFlow m r, MonadFlow m, CacheFlow m r)
  => Id Domain.Types.Image.Image -> m (Maybe Domain.Types.DriverUdyam.DriverUdyam)
findByImageId imageId =
  findOneWithKV [Se.Is Beam.documentImageId $ Se.Eq (Just $ Kernel.Types.Id.getId imageId)]
```

---

#### 6. `castPersonRoleToDashboardAccessType` collapses `FLEET_BUSINESS` into `FLEET_OWNER` (Confidence: 70)

**File:** `SharedLogic/DriverOnboarding.hs`

```haskell
castPersonRoleToDashboardAccessType FLEET_OWNER   = Just FLEET_OWNER
castPersonRoleToDashboardAccessType FLEET_BUSINESS = Just FLEET_OWNER  -- ← FLEET_BUSINESS → FLEET_OWNER
```

A config stored as `rolesAllowedToUploadDocument = [FLEET_BUSINESS]` would be returned to the dashboard as `[FLEET_OWNER]`. Any client enforcing role-based document upload permissions would incorrectly allow `FLEET_OWNER` users. If `FLEET_BUSINESS` has no exact equivalent in `DashboardAccessType`, this should be documented explicitly, or a new `FLEET_BUSINESS` variant should be added to that enum rather than silently merging.

---

#### 7. N+1 `FleetOwnerInformation` queries for TAN/LDC metadata (Confidence: 65)

**File:** `SharedLogic/DriverOnboarding/Status.hs`

```haskell
DVC.TANCertificate -> do
  ...
  mbFoi <- QFOI.findByPrimaryKey driverId    -- DB call 1
  ...
DVC.LDCCertificate -> do
  ...
  mbFoi <- QFOI.findByPrimaryKey driverId    -- DB call 2 — same row, same driver
```

When both TAN and LDC documents appear in the same status response with `enableDocumentMetadata = true`, the same `FleetOwnerInformation` row is fetched twice. The `FleetOwnerInformation` lookup should be hoisted to the `fetchDriverDocuments` call site and passed down, or memoized with `StateT`.

---

### 💡 Suggestion

#### 8. MSIL migration updates all document types indiscriminately

**File:** `dev/migrations-after-release/dynamic-offer-driver-app/0007-mark-image-valid-on-validation-skip-msil.sql`

```sql
UPDATE atlas_driver_offer_bpp.document_verification_config
SET mark_image_valid_on_validation_skip = true
WHERE merchant_operating_city_id = 'db1c62db-01ba-4cbf-9cce-ede8d2e1361e';
```

This enables skip-and-mark-valid for **every** document type in that city, including documents that might have active validation (e.g., selfie, DL). If only certain document types need this for MSIL, the migration should filter by `document_type` to avoid unintended side effects.

---

### ✅ What's Done Well

- The unified `validateDocumentApprovalChecks` abstraction is a significant improvement over per-document duplicate validation logic.
- The `markImageValidOnValidationSkip` feature flag is well-gated and the Beam type/from-to-type mapping is clean.
- The `findApproveImage` extraction reduces repeated error-message boilerplate.
- The `enableDocumentMetadata` feature flag correctly guards expensive decrypt operations.
- Migration files are present and complete for all new columns.
- The new `GET /vehicle/documents` endpoint correctly validates that at least one of `rcNo`/`rcId` is provided.
- The `UDYAMApproveDetails` refactor to use `documentImageId` instead of `udyamId` is architecturally consistent with all other document approve flows.

0utkarsh (approved):

PR #15416, opened 2026-06-29
backend/feat/backend-apis-for-aggregate-and-transactional-SAP-view

RahulGupta0706:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Finance Management APIs to list SAP journal entries and retrieve SAP journal transactions, with pagination, date windows, and filters (transaction type, status, batch/BELNR, subscription context).
  * Introduced new response schemas including journal headers and detailed debit/credit transaction rows.
* **Bug Fixes**
  * Improved settlement report ingestion and transaction handling to propagate subscription purchase identifiers consistently.
* **Chores**
  * Updated access permissions and expanded backend query/storage support (including new journal entry transaction persistence and supporting migrations).


RahulGupta0706:

ClaudeWeb:
I'll fetch the PR diff, description, and existing comments in parallel.Let me read the full diff and explore the workspace for context.Let me read the rest of the diff and examine the key query files.Now I have the full picture. Let me write the review.

---

## PR Review: `backend/feat/backend-apis-for-aggregate-and-transactional-SAP-view`

**Request changes.** The overall structure is sound — new SAP view APIs are properly wired through the dashboard layer with auth, the new `JournalEntryTransaction` table is correctly scaffolded, and the Beam/KV setup looks consistent with the codebase patterns. However, there are several correctness bugs that will cause silent data corruption or runtime errors in production.

---

### 🔴 Critical — Bugs

#### 1. `saveSapJournalEntries`: throws `InternalError` on empty request headers (regression)

**File:** `SAPReportDispatch.hs`

```haskell
results <- forM reqHeaders $ \reqHeader -> do
  ...
  pure (sapEntryId, reqHeader.batchId)
listToMaybe results & fromMaybeM (InternalError "SAP journal entry creation returned no entry ID")
```

If `req.headers` is empty, `results = []`, `listToMaybe []` → `Nothing`, and `fromMaybeM` throws `InternalError`. The original code used `forM_` and silently returned `()`. This is a regression — a valid SAP response with empty headers now crashes the dispatch job instead of completing successfully.

**Fix:** Guard before the call, or propagate as `Maybe`:
```haskell
case listToMaybe results of
  Nothing -> throwError $ InternalError "SAP journal entry creation returned no entry ID"
  Just (entryId, bId) -> pure (entryId, bId)
```
…or better, rethink whether the caller should handle empty-headers gracefully before reaching this function.

---

#### 2. Multiple SAP headers: only first journal entry ID is used for all transaction rows

**File:** `SAPReportDispatch.hs`

`saveSapJournalEntries` creates one `SapJournalEntry` per request header (the inner `forM reqHeaders`), but returns only `listToMaybe results` — i.e., the *first* entry's ID. Then:

```haskell
(sapEntryId, sapBatchId) <- saveSapJournalEntries ...
forM_ subRows $ \rows -> saveSubscriptionTransactions mId mocid sapEntryId sapBatchId rows
forM_ pgRows $ \rows -> savePGSettlementTransactions mId mocid sapEntryId sapBatchId rows
```

All per-transaction records are linked to only the first SAP journal entry, regardless of which header they belong to. If a multi-header SAP request is dispatched, the `sap_journal_entry_id` / `sap_batch_id` on those transaction rows will be incorrect.

**Fix:** Either (a) assert that `reqHeaders` always contains exactly one header (and document this invariant), or (b) return all `(entryId, batchId, rows)` tuples and zip them correctly when saving transactions.

---

### 🟡 Warning — Incorrect Behaviour

#### 3. `Summary.totalCount` and `SapJournalTransactionsRes.total` reflect page size, not true total

**File:** `FinanceManagement.hs` (provider-platform)

```haskell
-- SAP Journals
let totalItems = length entries  -- entries is already paginated (max 100)
    summary = Dashboard.Common.Summary {totalCount = totalItems, count = totalItems}

-- SAP Transactions
let items = map toTransactionItem entries
pure API.SapJournalTransactionsRes {total = length items, transactions = items}
```

`entries` is fetched with `limit` applied. So `totalCount` / `total` is at most 100, and is always equal to the page size — not the true number of matching records. Clients cannot paginate correctly because they cannot determine how many total records exist.

**Fix:** Issue a separate `countAllWithOptionsKV` (or equivalent) call with the same filters but no limit/offset, and use that as `totalCount`.

---

#### 4. `savePGSettlementTransactions`: both debit and credit set to same amount

**File:** `SAPReportDispatch.hs`

```haskell
JET.JournalEntryTransaction
  { debitAmount = row.amount,
    creditAmount = row.amount,   -- ← same value
    ...
  }
```

Compare with `saveSubscriptionTransactions` which correctly uses distinct `row.debitAmount` / `row.creditAmount`. Setting both sides to the same value in double-entry accounting is incorrect — one side should be zero (or they represent different amounts). This looks like a copy-paste omission.

---

#### 5. `resolveTimeRange` doesn't validate `from <= to`

**File:** `FinanceManagement.hs` (provider-platform)

```haskell
when (from < pStart || to > pEnd) $
  throwError $ InvalidRequest "Cannot filter in this time range"
pure (from, to)
```

A caller can supply `fromTime` after `toTime` (e.g., `fromTime=2025-12-31, toTime=2025-01-01`). No error is thrown; the query proceeds and silently returns zero results.

**Fix:** Add `when (from > to) $ throwError $ InvalidRequest "fromTime must be before toTime"`.

---

#### 6. `filterQuery` change breaks existing `findAllByMerchantOpCityIdWithFilters` callers

**File:** `PgPaymentSettlementReportExtra.hs`

```haskell
-  <> [Se.Is Beam.referenceId $ Se.Eq (Just subscriptionPurchaseId) | Just subscriptionPurchaseId <- [mbSubscriptionPurchaseId]]
+  <> [Se.Is Beam.subscriptionPurchaseId $ Se.Eq (Just subscriptionPurchaseId) | Just subscriptionPurchaseId <- [mbSubscriptionPurchaseId]]
```

`subscriptionPurchaseId` is a brand-new nullable column; all existing rows have `NULL` for it. Any existing caller that passes `mbSubscriptionPurchaseId = Just someId` will now get zero results (the new column is NULL for all old rows), whereas it used to match on `referenceId`. This is a silent breaking change for any live query relying on that filter.

---

#### 7. `pgTxnTypeToSJE`: silent fallback to `SJE.Order` for unknown types

**File:** `SAPReportDispatch.hs`

```haskell
pgTxnTypeToSJE t = case T.toLower t of
  "order" -> SJE.Order
  "refund" -> SJE.Refund
  "chargeback" -> SJE.Chargeback
  _ -> SJE.Order   -- ← any unexpected value recorded as Order
```

If `show tt` (where `tt :: PgDom.TxnType`) produces an unexpected string (e.g., a new enum variant added later), this silently records the transaction as `Order`. A loud failure (or `Maybe` return type) would be safer.

---

#### 8. Invalid `status` filter silently dropped

**File:** `FinanceManagement.hs` (provider-platform)

```haskell
mbSt = mbStatus >>= readMaybe . toString
```

If the caller passes `status=success` (wrong casing or invalid string), `readMaybe` returns `Nothing`, the filter is silently ignored, and the API returns results for all statuses. No error is returned to the caller. Consider returning `400 InvalidRequest` for unrecognised status values.

---

### 🔵 Suggestion — Performance

#### 9. Missing indexes on `journal_entry_transaction`

**File:** `journal_entry_transaction.sql`

The migration adds only a primary key on `id`. The main query in `JournalEntryTransactionExtra.hs` filters by `merchant_id`, `merchant_operating_city_id`, `transaction_type`, and `created_at`. Without composite indexes, this will be a full table scan as the table grows.

**Suggested index:**
```sql
CREATE INDEX idx_jet_merchant_type_created
  ON atlas_driver_offer_bpp.journal_entry_transaction
  (merchant_id, merchant_operating_city_id, transaction_type, created_at DESC);
```

---

#### 10. N+1 writes in `saveSubscriptionTransactions` / `savePGSettlementTransactions`

**File:** `SAPReportDispatch.hs`

Both functions call `QJETExtra.create` individually inside `forM_`. For a large batch of subscription transactions, this generates one `INSERT` per row. The already-generated `createMany` in `JournalEntryTransaction.hs` supports bulk inserts — use it instead.

```haskell
-- Instead of forM_:
let records = map (buildRecord now txnId) rows
QJETExtra.createMany records
```

---

**Summary:** The API structure and auth wiring are correct. The critical issues (#1 and #2) can cause job crashes or silent data corruption in the SAP dispatch path; #3 makes the APIs' pagination unusable; #6 is a silent breaking change. These should be fixed before merge.

0utkarsh (approved):

PR #15472, opened 2026-06-30
 refac : docker.nix to use the inc cabal build

AVBharath10:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

### DIFF
--- a/Backend/nix/docker.nix
+++ b/Backend/nix/docker.nix
@@ -7,60 +7,84 @@ let
   commitHash = builtins.substring 0 6 (self.rev or "dev");
   tagPrefix = builtins.getEnv "DOCKER_TAG_PREFIX";
   imageTag = if tagPrefix != "" then "${tagPrefix}${commitHash}" else commitHash;
+  localBinariesPath = builtins.getEnv "LOCAL_BINARIES_PATH";
+  localBuild = localBinariesPath != "";
 in
 {
   config = {
-    perSystem = { self', pkgs, lib, ... }: {
-      packages = lib.optionalAttrs pkgs.stdenv.isLinux {
-        dockerImage = (pkgs.dockerTools.buildImage {
-          name = imageName;
-          created = "now";
-          tag = imageTag;
-          copyToRoot = pkgs.buildEnv {
-            paths = with pkgs; [
-              cacert
-              awscli
-              coreutils
-              bash
-              self'.packages.nammayatri
-              gdal
-              postgis
-              curl
-              htop
-              wget
-              zbar
-              wkhtmltopdf-bin
-              openssh # For SFTP CLI support
-              sshpass # For password-based SFTP authentication
-            ];
-            name = "beckn-root";
-            pathsToLink = [
-              "/bin"
-              "/opt"
-            ];
-          };
-          config = {
-            Env = [
-              "SSL_CERT_FILE=${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt"
-              # Ref: https://hackage.haskell.org/package/x509-system-1.6.7/docs/src/System.X509.Unix.html#getSystemCertificateStore
-              "SYSTEM_CERTIFICATE_PATH=${pkgs.cacert}/etc/ssl/certs/ca-bundle.crt"
-            ];
-            Cmd = [ "${self'.packages.nammayatri}/bin/rider-app-exe" ];
-          };
+    perSystem = { self', pkgs, lib, config, ... }:
+      let
+        nammayatriBinaries =
+          if localBuild
+          then
+            pkgs.runCommand "nammayatri-local"
+              {
+                nativeBuildInputs = [ pkgs.autoPatchelfHook pkgs.makeWrapper ];
+                buildInputs =
+                  config.haskellProjects.default.outputs.devShell.buildInputs
+                  ++ [ config.haskellProjects.default.outputs.finalPackages.cac_client pkgs.libsodium pkgs.rdkafka ];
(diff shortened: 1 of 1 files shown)


khuzema786 (approved):

AVBharath10:
@ClaudeWeb can u check this docker.nix  file change and compare it to main

ClaudeWeb:
I'll fetch the PR diff, description, and compare with the main branch file simultaneously.Good, I have the diff. Let me now read the main branch file and surrounding Nix context.The repo isn't in /workspace. Let me fetch the main branch file via GitHub API.Now let me fetch the main branch file and the flake/haskell nix files for context.Now I have the full picture. The main branch SHA (`d0b80e1d09e`) matches the PR's new version — this PR has already been merged. I'll still review the changes thoroughly.I have everything I need. Here's the full review:

---

## PR #15472 Review — `docker.nix`: local incremental cabal build support

> **Note:** The file blob SHA on `main` (`d0b80e1d09e`) matches the PR's new version exactly — this PR appears to already be merged into main. Review is provided for record.

**Verdict: Changes work for the happy path, but there are 3 issues worth addressing as a follow-up.**

---

### What Changed

The PR adds a `localBuild` code path controlled by `LOCAL_BINARIES_PATH` env var. When set, it creates a `pkgs.runCommand` derivation that:
1. Copies pre-compiled binaries from the local path into the Nix store
2. Runs `autoPatchelfHook` to fix ELF interpreter paths
3. Wraps each binary with `makeWrapper` to inject the correct `LD_LIBRARY_PATH`

This lets CI/developers use an **incremental `cabal build`** output directly in the Docker image, bypassing the full Nix cabal rebuild.

---

### Warning — `builtins.getEnv` silently returns `""` in pure flake evaluation (confidence: 92)

**File:** `Backend/nix/docker.nix`, line 8–9

Nix flakes evaluate in **pure mode by default**. `builtins.getEnv` returns `""` in pure mode, so `localBuild` is always `false` unless the user explicitly passes `--impure`. There's no comment, trace, or error to signal this. The feature silently does nothing without that flag.

```nix
# Current — no indication --impure is required
localBinariesPath = builtins.getEnv "LOCAL_BINARIES_PATH";
localBuild = localBinariesPath != "";
```

**Fix:**
```nix
# LOCAL_BINARIES_PATH requires: nix build --impure (getEnv returns "" in pure mode)
localBinariesPath = builtins.getEnv "LOCAL_BINARIES_PATH";
localBuild = localBinariesPath != "";
```
Or emit a visible trace so the user knows the env var was picked up:
```nix
localBuild =
  if localBinariesPath != ""
  then builtins.trace "docker.nix: using local binaries from ${localBinariesPath}" true
  else false;
```

---

### Warning — `extraCommands` sanity check will fail for local builds (confidence: 82)

**File:** `Backend/nix/docker.nix`, lines 67–71 (post-merge numbering)

```nix
extraCommands = ''
  ls opt/app/rider-app-exe   # only valid for self'.packages.nammayatri layout
  ls opt/app/swagger          # swagger dir won't exist in a cabal build output
'';
```

The `self'.packages.nammayatri` Nix package places binaries under `/opt/app/`. A local cabal build output has them under `/bin/`. The `swagger` directory in particular is generated by the Nix build and will not exist in a plain cabal output, so `nix build --impure` with `LOCAL_BINARIES_PATH` set will **always fail** at the `extraCommands` step.

**Fix:**
```nix
extraCommands = lib.optionalString (!localBuild) ''
  # Executables are under /opt/app
  ls opt/app/rider-app-exe
  # Swagger configs are copied over
  ls opt/app/swagger
'';
```

---

### Warning — Bash glob without `nullglob` passes literal `*` to `wrapProgram` (confidence: 80)

**File:** `Backend/nix/docker.nix`, inside the `pkgs.runCommand` script

```bash
for exe in $out/bin/*; do
  wrapProgram "$exe" ...
done
```

In bash (the default Nix builder shell), if `$out/bin/` is empty or doesn't exist, the glob doesn't expand — bash passes the literal string `$out/bin/*` to `wrapProgram`, which will fail with a confusing "no such file" error rather than a clean no-op.

**Fix:**
```bash
chmod -R +w $out/bin
shopt -s nullglob
for exe in $out/bin/*; do
  wrapProgram "$exe" \
    --prefix LD_LIBRARY_PATH : "..."
done
```

---

### Suggestion — Document what `LOCAL_BINARIES_PATH` should point to (confidence: 75)

`src = /. + localBinariesPath;` coerces the env var to a Nix path and **copies the entire directory into the Nix store at evaluation time**. If someone accidentally points this at `dist-newstyle/` (the full cabal build tree), it'll copy gigabytes into `/nix/store`. A comment clarifying the expected structure would prevent this.

```nix
# LOCAL_BINARIES_PATH must be an absolute path to a directory structured as:
#   /bin/<exe-name>   (cabal install --installdir=... output)
# Do NOT point at dist-newstyle/ — the whole tree gets copied into the Nix store.
src = /. + localBinariesPath;
```

---

### What's Good

- The two-path design (`localBuild` vs `self'.packages.nammayatri`) is clean — no impact on the standard Nix build path.
- `autoPatchelfHook` + `makeWrapper` is the correct way to handle locally-compiled ELF binaries in a Nix sandbox.
- `--prefix LD_LIBRARY_PATH` (not `--set`) is correct — it doesn't clobber any existing path.
- `config` was properly added to the `perSystem` destructuring to access `haskellProjects`.

PR #15520, opened 2026-07-02
[refactor]: Driver Fee to be in KV

Yashwanth1906:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Ride start/end and downstream processing now fan out by driver-fee shard using an active-drivers Redis set/queue, with shardNum propagated through job payloads (overlays, notifications, payouts, mandates, and driver-fee execution).
  * Added active-driver shard-count configuration and shard-aware driver batching for mandate execution.

* **Bug Fixes**
  * Improved driver-fee bill-number locking to be date/time-zone scoped and safe for concurrent shard runs.
  * Missing/unsupported plan outcomes now mark fees for manual review; cancellation penalties and dispute cleanup are tracked per driver to ensure correct state transitions.


dushyant-2002:
@ClaudeWeb review this PR .
what are the changes in this PR

ClaudeWeb:
Let me fetch the PR details, diff, and existing comments in parallel.Let me read the full diff and also get the PR files list for a thorough review.Now let me check the `Redis.shardHashTag` implementation and related code for the core logic correctness.Let me check the Allocator dhall config and the cancellation penalty change more carefully.Now I have a complete picture of the PR. Let me write the review.

---

## PR #15520 — [refactor]: Driver Fee to be in KV

**Verdict: Request Changes.** The PR introduces a well-architected shard-based fan-out for driver fee processing using Redis KV, but has a critical bug in shard-completion signalling that can produce duplicate downstream jobs, and a TTL edge-case that can cause the active-drivers set to expire before the billing job reads it.

---

### Critical

#### 1. `closeOutWindow` deduplication key is shard-specific — duplicate `scheduleJobs` calls possible

**File:** `SharedLogic/Allocator/Jobs/DriverFeeUpdates/DriverFee.hs` · lines 113–245

`hashedJobData` is computed from `show jobData`, and `jobData` now includes `shardNum :: Maybe Int`. Each shard job therefore produces a **unique hash**. The `closeOutWindow` SETNX guard is:

```haskell
Hedis.setNxExpire (jobDuplicationPreventionKey hashedJobData "DriverFeeCalc") ...
```

When `completedShards >= totalShards` is satisfied (a non-atomic read-check-act), **multiple shards can each call `closeOutWindow` with a different SETNX key, all succeeding**, and all invoking `scheduleJobs`. Inside `scheduleJobs`, the notification/overlay/manual-payment-link jobs are created **before** the `setDriverFeeCalcJobCache` guard, so they get duplicated N times (once per shard that fires `closeOutWindow`).

**Fix:** Base the deduplication key on the billing-window identity (merchantOpCityId + serviceName + startTime + endTime), not on the per-shard job payload:

```haskell
let windowKey = "DriverFeeCalcCloseOut:" <> merchantOpCityId.getId
                  <> ":" <> show serviceName
                  <> ":" <> show startTime
                  <> ":" <> show endTime
let closeOutWindow =
      Hedis.setNxExpire windowKey (3600 * 12) True
        >>= \duplicationKey ->
          when duplicationKey $
            scheduleJobs transporterConfig startTime endTime merchantId merchantOpCityId serviceName subscriptionConfigs jobData
```

---

### Warnings

#### 2. Active-drivers set TTL expires exactly when the billing job fires

**File:** `SharedLogic/ActiveDriversList.hs` · lines 51–59

`getTTLForActiveDriversList` sets the TTL to expire at **exactly** midnight local time (the same moment the billing window closes and the `CalculateDriverFees` job is scheduled). If the scheduler adds even a few seconds of latency — or `dfCalculationJobTs` is a positive delay — the Redis set is already gone when `getNextDriverBatch` first tries to populate the per-shard queue from it, yielding an empty queue and silently skipping all drivers.

```haskell
-- current: expires AT midnight local — too tight
pure $ ceiling (realToFrac (diffUTCTime tomorrowUTC now) :: Double)

-- fix: keep the set alive for at least the expected max job delay + buffer
pure $ ceiling (realToFrac (diffUTCTime tomorrowUTC now) :: Double) + (2 * 24 * 3600)
```

The queue itself already has a 48-hour TTL, so keeping the source set alive for 48 hours is cheap and safe.

---

#### 3. `billNumberGenerationLockKey` scope change is a silent semantic shift

**File:** `Domain/Types/Extra/DriverFee.hs` · line 263

```haskell
-- before: one lock per city (coarse)
billNumberGenerationLockKey merchantOpCityId = "DriverFee:BillNumber:Processing:MerchantOpCityId" <> merchantOpCityId

-- after: one lock per billNumberKey = one lock per (city × service × date)
billNumberGenerationLockKey billNumberKey = "DriverFee:BillNumber:Processing:" <> billNumberKey
```

The narrower lock is strictly correct for concurrent shard runs, but the parameter rename from `merchantOpCityId` to `billNumberKey` silently changes what the lock key contains — callers passing an arbitrary `Text` (the full Redis key) instead of just an ID. If any future caller passes only a city ID instead of the full `billNumberKey`, the lock degrades to the old (too-coarse) granularity without a compile error. Consider making the type more explicit:

```haskell
newtype BillNumberKey = BillNumberKey Text
billNumberGenerationLockKey :: BillNumberKey -> Text
```

---

#### 4. Ride-notification retry broken under sharding

**File:** `SharedLogic/Allocator/Jobs/Mandate/Notification.hs` · lines 81–113

`findDriverFees` with a `Just shardNum` drains from a per-shard Redis queue (populated once via `getNextDriverBatch`). Once the queue is exhausted on the first pass, every subsequent call — including **retry passes** — returns `[]`. Drivers whose payment-gateway notification actually failed but whose DB status remains `PAYMENT_PENDING` / `NOTIFICATION_SCHEDULED` are invisible to the retry scan (they were drained from the queue). In the non-shard path, retries re-query the DB by status and find them correctly.

Consider re-queuing failed drivers back into the shard queue, or falling back to a DB query on retry passes (`retryCount > 0`):

```haskell
let findDriverFees = case (jobData.shardNum, retryCount > 0) of
      (Just shardNum, False) -> ... -- queue path
      _                      -> QDF.findDriverFeeInRangeWithNotifcationNotSentServiceNameAndStatus ...
```

---

### Suggestions

#### 5. Misleading fork label in `StartRide.hs`

**File:** `Domain/Action/UI/Ride/StartRide.hs` · line 249

```haskell
fork "startRide - Complete pickup zone request and add driver to active list" $ do
  ADL.addDriverToActiveList driverId ...
-- ↑ this fork does NOT complete the pickup zone request; the next fork does
```

Rename to `"startRide - Add driver to active drivers list"` to avoid confusing log traces.

---

#### 6. PR checklist is entirely unchecked

The PR description has no boxes checked (refactoring type, testing, formatting). For a change that touches billing job scheduling and Redis key lifecycle, noting how it was tested (manual trigger? staging run?) would help reviewers.

---

### What's Good

- `getNextDriverBatch` correctly uses a **blocking** Redis lock (`withWaitAndLockRedis`) rather than a try-lock to prevent a populate-then-drain race — this is the right call and is well-documented.
- The `mkActiveDriversSetKey` (write) and `mkActiveDriversSetKeyForShard` (read) key formats are compatible — both emit `{shard-N}` hash tags via `Redis.shardHashTag`.
- Backward compatibility is preserved: all `shardNum :: Maybe Int` fields default to `Nothing` when deserializing old job JSON, so in-flight jobs at deploy time are handled correctly.
- `billNumberGenerationLockKey` narrowing to per-(city × service × date) is the right fix for concurrent shard safety.
- Short-circuiting all `*ByDriverIds` queries when `driverIds` is empty (`| null driverIds = pure []`) avoids accidental full-table scans.

PR #15561, opened 2026-07-06
Backend:enhc: decoupled enabledOtpLessRide from safetySettings

dhruv-1010:
## Summary by CodeRabbit

* **New Features**
  * Added support for OTP-less rides, including a new rider profile setting and city-level availability control.
  * Booking and ride flows now carry this setting through confirmation, status, and ride start/end experiences.
* **Bug Fixes**
  * Ride setup now uses the booking’s OTP-less ride setting consistently, instead of relying on unrelated safety settings.
  * Profile updates now block OTP-less rides when the feature is not enabled for the city.
* **Tests**
  * Added end-to-end coverage for an OTP-less ride booking journey.


dhruv-1010:
@ClaudeWeb review this PR

ClaudeWeb:
I'll fetch the PR details, diff, and comments in parallel.Let me read the full diff and explore the related files.Now let me look at the key source files for deeper context.Let me check whether SocialLogin calls through the gated `buildPerson` path.Now I have the full picture. Let me compose the review.I now have full context. Here is the review.

---

## PR Review: `#15561` — Decouple `enableOtpLessRide` from `safetySettings`

**Overall verdict: Request Changes.** The architectural direction is correct — snapshotting `enableOtpLessRide` onto the Booking at confirm time (instead of a live SafetySettings lookup per ride) is a clear improvement. However there are two real regressions caused by a missing data-migration, plus a dual-state consistency risk.

---

### Warning 1 — Silent preference loss for existing users (Confidence: 90)

**File:** `Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Confirm.hs` (line ~322) and `dev/migrations-read-only/rider-app/booking.sql`

**Problem:** The PR moves the source of truth from `SafetySettings.enableOtpLessRide` (read live at ride-creation time) to `Person.enableOtpLessRide` (snapshotted onto the Booking at confirm time). However, no backfill migration syncs `Person.enableOtpLessRide` from `SafetySettings.enableOtpLessRide`.

Users who previously enabled OTP-less rides via the PUT `/emergency-settings` endpoint had the value stored **only in SafetySettings**, not in `Person`. After this deploy, when they book a new ride:

```haskell
-- Confirm.hs (new code)
let bookingEnableOtpLessRide = if riderConfig.enableOtpLessRide
                                 then person.enableOtpLessRide  -- NULL for these users!
                                 else Nothing
```

Their `person.enableOtpLessRide` is `NULL`, so `bookingEnableOtpLessRide = Nothing`, so OTP-less is silently disabled. The user's preference is lost without any notification.

**Fix:** Add a migration to backfill `Person.enableOtpLessRide` from SafetySettings before deploying, e.g.:

```sql
UPDATE atlas_app.person p
SET enable_otp_less_ride = true
FROM atlas_safety.safety_settings ss
WHERE p.id = ss.person_id
  AND ss.enable_otp_less_ride = true
  AND (p.enable_otp_less_ride IS NULL OR p.enable_otp_less_ride = false);
```

---

### Warning 2 — In-flight bookings break on deploy (Confidence: 85)

**Files:** `Domain/Action/Beckn/OnInit.hs`, `Domain/Action/Beckn/Common.hs`, `Domain/Action/Beckn/OnStatus.hs`

**Problem:** The booking column migration adds `enable_otp_less_ride boolean` with **no default**:

```sql
-- booking.sql
ALTER TABLE atlas_app.booking ADD COLUMN enable_otp_less_ride boolean ;
```

Any booking that is `CONFIRMED` or `TRIP_ASSIGNED` at deploy time will have `NULL` in this column. All three ride-creation paths now read:

```haskell
enableOtpLessRide = Just $ isMeterRide || fromMaybe False booking.enableOtpLessRide
-- fromMaybe False Nothing = False → OTP is now required
```

A rider who had `safetySettings.enableOtpLessRide = true` and a pending booking will find their driver suddenly requires an OTP for a ride that was accepted as OTP-less. This is a UX regression and could cause ride failures.

**Fix — Option A (preferred):** Add `DEFAULT false` to the booking column migration so existing rows get `false` instead of `NULL`. For users who had OTP-less enabled, this is still a regression, but it at least has a predictable default. Combine with the backfill in Warning 1.

**Fix — Option B:** Add a short-lived fallback in `OnInit`/`Common`/`OnStatus`: if `booking.enableOtpLessRide` is `Nothing`, fall back to a SafetySettings lookup (keeping the old behavior) until a backfill migration runs.

---

### Suggestion — Dual state with no atomicity guarantee (Confidence: 75)

**File:** `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Profile.hs` (line ~733)

`updateEmergencySettings` now writes to **two separate tables** sequentially:

```haskell
-- Write 1: SafetySettings (separate service)
void $ Lib.upsert (cast personId) emergencyInfo ...
-- Write 2: Person table
whenJust enableOtpLessRide $ \v -> QPerson.updateEnableOtpLessRide personId (Just v)
```

If Write 1 succeeds but Write 2 fails (transient DB error), GET `/emergency-settings` (which reads from SafetySettings line 793) will return `enableOtpLessRide = true`, but new bookings (reading `person.enableOtpLessRide`) will see `Nothing`. The user sees "enabled" in the UI but their rides will require OTP.

Consider making `Person` the sole source of truth and having `getEmergencySettings` read from Person, rather than maintaining parallel state across two tables.

---

### Minor — Migration files missing trailing newline

**Files:** `dev/migrations-read-only/rider-app/booking.sql`, `dev/migrations-read-only/rider-app/rider_config.sql`

Both end with `\ No newline at end of file`. This is a minor inconsistency with other migration files but worth fixing for clean git diffs.

---

### What's good

- Removing the per-ride `SafetySettings` DB lookup (`QSafetySettings.findByPersonId`) in `Common.hs` and `OnStatus.hs` is a solid performance improvement — one fewer DB round-trip per ride assignment.
- The city-config gate (`RiderConfig.enableOtpLessRide`) is correctly applied in all write paths: `buildPerson` (Registration), `updatePerson` (Profile), and `updateEmergencySettings`.
- The integration test collection (`15-OtpLessRideFlow.json`) is thorough and covers the full ride lifecycle.
- The feature migration SQL for Helsinki is well-guarded (handles missing merchant gracefully with a RAISE NOTICE).

PR #15583, opened 2026-07-07
backend/enhancement - added proper code formatting and check

RitikaHanish:
backend/enhancement - added proper code formatting and check

## Summary by CodeRabbit

* **Bug Fixes**
  * Vehicle onboarding status now resolves more consistently across driver and dashboard flows.
  * Active vehicle-association lookups now return the most recent matching record first, reducing unexpected reassignment results.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Onboarding.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Fleet/Onboarding.hs
@@ -174,7 +174,7 @@ getOnboardingRegisterVehicleStatus ::
 getOnboardingRegisterVehicleStatus merchantShortId opCity mbRegistrationNo mbRcId = do
   merchant <- findMerchantByShortId merchantShortId
   merchantOpCity <- CQMOC.findByMerchantIdAndCity merchant.id opCity >>= fromMaybeM (MerchantOperatingCityNotFound $ "merchantShortId: " <> merchantShortId.getShortId <> " ,city: " <> show opCity)
-  (registrationNo, verified, approved, documents) <- DOnboarding.rcVerifyStatus Nothing True (Just merchantOpCity.id) mbRegistrationNo mbRcId
+  (registrationNo, verified, approved, documents) <- DOnboarding.rcVerifyStatus (DOnboarding.DashboardCaller merchantOpCity.id) mbRegistrationNo mbRcId
   pure $
     CommonOnboarding.RcVerifyStatusResp
       { registrationNo,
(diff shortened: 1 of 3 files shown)


RitikaHanish:
@ClaudeWeb Please review this PR ?

ClaudeWeb:
I'll fetch the PR details, diff, and comments in parallel before diving into the review.Good, I have the diff. Let me now read the surrounding source files for more context.The workspace doesn't have the repo. Let me use GitHub MCP tools to gather context from the relevant files.Now let me fetch the full source files to understand the surrounding context better.Let me read the large DriverOnboardingV2.hs file around the changed area.Let me check the module export list and the existing docstring in the dashboard file.I have all the context I need. Let me now write the full review.---

## PR Review: `backend/enhancement - added proper code formatting and check`

**Verdict: Approve with minor fix** — The core refactor is well-executed and idiomatic Haskell. One stale comment should be fixed before merge.

---

### What this PR does

Three focused improvements:
1. **`DriverOnboardingV2.hs`** — Replaces the old `(Maybe callerId, Bool isDashboard, Maybe cityId)` triple with a `RcVerifyCaller` sum type (`DriverCaller | DashboardCaller`), making illegal states unrepresentable and closing an IDOR hole where the driver path could silently skip RC-ownership auth if `mbCallerId` was accidentally `Nothing`.
2. **`Onboarding.hs`** — Updates the dashboard call-site to use the new `DashboardCaller merchantOpCity.id` constructor.
3. **`FleetRCAssociationExtra.hs`** — Adds `ORDER BY associatedOn DESC` to `findAllActiveAssociationByRCId`, so callers using `listToMaybe` deterministically get the most-recent active association during RC reassignment windows.

All three changes are correct and the logic is sound.

---

### Warning

**`Onboarding.hs` — Stale docstring on `getOnboardingRegisterVehicleStatus` (~line 163)**

The haddock comment still describes the old API:

```haskell
-- | Dashboard vehicle-only RC verify-status. Delegates to the shared @rcVerifyStatus@ with @isDashboard =
--   True@ (and no caller), which skips the per-person RC-ownership gate ...
```

The `isDashboard = True` flag and `(and no caller)` no longer exist. Should be updated to match the new `DashboardCaller` terminology:

```haskell
-- | Dashboard vehicle-only RC verify-status. Delegates to the shared @rcVerifyStatus@ as a
--   @DashboardCaller@, which skips the per-person RC-ownership gate — dashboard callers are already
--   merchant/city-scoped by ApiAuthV2. The RC's resolved city must match this operating city.
```

---

### Suggestion

**`DriverOnboardingV2.hs` — `RcVerifyCaller` defined after its first use**

The data type is declared after `getDriverRegisterVehicleStatus` which uses `DriverCaller`. Haskell compiles this fine (no forward-declaration requirement), but readers see the constructor before its definition. Moving the type block above `getDriverRegisterVehicleStatus` would follow the module's existing convention and aid readability.

---

### Positives worth calling out

- **IDOR fix is correct**: The `DriverCaller` path now always runs `DAQuery.findLinkedByRCIdAndDriverId callerId rc.id now >>= fromMaybeM RCNotLinked`, wrapped in `Just <$>`. The old `forM mbCallerId` could silently return `Nothing` if `mbCallerId` was `Nothing`, bypassing the ownership check entirely. The sum type eliminates this class of bug at the type level.
- **`DashboardCaller` city pinning**: `mbPassedCityId` is cleanly derived from the constructor payload, and the `whenJust` city-equality check downstream is unchanged — the refactor didn't accidentally weaken it.
- **`findAllWithOptionsKV` ordering**: The `Se.Desc Beam.associatedOn` sort is consistent with `findLatestByRCIdAndFleetOwnerId`'s existing `Se.Desc Beam.associatedTill` pattern — no new query-layer conventions introduced.

PR #15631, opened 2026-07-08
backend/fix: Adding rcStatus call for vehicle in bot flow

Brajmohank1:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vehicle review submission by validating the most up-to-date Registration Certificate (RC) verification status before accepting a review request.
  * If the RC is not verified, the submission is now blocked with a clear validation error.
  * Helps ensure unverified RC documents are not treated as approved during the review flow.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/Driver.hs
@@ -36,6 +36,7 @@ import qualified Domain.Action.Internal.DriverMode as DDriverMode
 import qualified Domain.Action.UI.DriverOnboarding.CourtRecordCheck as CourtRecordCheck
 import qualified Domain.Action.UI.DriverOnboarding.Referral as DOR
 import qualified Domain.Action.UI.DriverOnboarding.VehicleRegistrationCertificate as DomainRC
+import qualified Domain.Action.UI.DriverOnboardingV2 as DOnboarding
 import qualified Domain.Action.UI.OperationHub as Domain
 import qualified Domain.Action.UI.Registration as DReg
 import qualified Domain.Types.AadhaarCard as DAadhaarCard
@@ -1023,6 +1024,8 @@ postDriverSubmitReviewRequest merchantShortId opCity requestorId req = do
         when (rcInfo.verified /= Just True) $ throwError (InvalidRequest "RC is not verified")
         if isDocReqEmpty
           then do
+            (_, verified, _, _) <- DOnboarding.rcVerifyStatus (DOnboarding.DashboardCaller merchantOpCity.id) Nothing (Just reqId)
+            unless verified $ throwError (InvalidRequest "RC is not verified")
             unless (rcInfo.approved == Just True) $ QVRC.updateByPrimaryKey rcInfo {DVRC.approved = Just True}
             pure (DRR.COMPLETED, rcInfo.unencryptedCertificateNumber, Nothing)
           else do


Brajmohank1:
@ClaudeWeb Please review this PR by making sure everything is correct.

ClaudeWeb:
I'll review PR #15631 now. Let me fetch the diff, description, and existing comments in parallel.Good, the diff is small. Let me read the surrounding code for full context.The workspace doesn't have the full codebase. Let me use GitHub MCP tools to fetch the relevant files.Now let me fetch the key source files in parallel — the modified file on the PR branch, the `rcVerifyStatus` implementation, and the surrounding context.Now let me read the `rcVerifyStatus` function definition and the modified file's context around the changed lines.Let me extract the relevant sections from both large files.Let me get more context: the rest of `rcVerifyStatus` and the full `postDriverSubmitReviewRequest` function.I now have all the context I need. Let me analyze the complete picture.

**Summary:**

The diff is 2 lines. The `rcVerifyStatus` signature is `RcVerifyCaller -> Maybe Text -> Maybe Text -> Flow (Text, Bool, Maybe Bool, [DocumentStatusItem])`, and the call in the PR correctly constructs a `DashboardCaller`. The existing function already uses this same pattern in `getOnboardingRegisterVehicleStatus`. I have read all relevant sections of both files.

---

## PR Review — `backend/fix: Adding rcStatus call for vehicle in bot flow` (#15631)

**Verdict: Approve with minor warnings.** The logic is correct and the pattern is consistent with existing dashboard usage. Two issues worth noting.

---

### What the change does

In `postDriverSubmitReviewRequest` (VEHICLE branch, `isDocReqEmpty = True` path), before marking the RC `approved = True`, the PR now calls `rcVerifyStatus` to re-compute the live document validity rather than relying solely on the stored `rcInfo.verified` field. This prevents approving a vehicle whose docs have since expired or been invalidated between the initial verification and the review action.

---

### Warning — Duplicate RC database fetch

**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Operator/Driver.hs` ~line 1020 and inside `rcVerifyStatus`

The RC row is fetched twice in sequence:
1. `rcInfo <- QVRC.findByPrimaryKey rcId` (start of the VEHICLE branch)
2. `RCQuery.findById (Id rcId)` — inside `rcVerifyStatus` when given `(Nothing, Just rcId)`

Both resolve the same row. This is an unnecessary extra DB round-trip in a hot code path (operator review submit).

**Suggested fix:** Not strictly necessary right now, but `rcVerifyStatus` could accept a pre-fetched `rc` record as an overload, or the call could pass the `registrationNo` from `rcInfo.unencryptedCertificateNumber` instead of the ID, skipping the internal fetch:

```haskell
-- Pass registrationNo instead of rcId to avoid second fetch
(_, verified, _, _) <- DOnboarding.rcVerifyStatus
  (DOnboarding.DashboardCaller merchantOpCity.id)
  rcInfo.unencryptedCertificateNumber  -- Maybe Text
  Nothing
```

However, `unencryptedCertificateNumber` is `Maybe Text`, so this only works if it's guaranteed to be set for an RC that passed the `verified == Just True` gate. Worth confirming.

---

### Warning — `rcVerifyStatus` has a write side-effect, not just a read

**File:** `Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboardingV2.hs`

When `transporterConfig.enableBotFlow == Just True`, `rcVerifyStatus` writes `verified = allValid` back to the DB via `RCQuery.updateVerifiedByCertificateNumberHash`. This means calling it in the submit-review path is not a pure check — it **mutates** `rc.verified` as a side-effect.

```haskell
-- Inside rcVerifyStatus:
when (transporterConfig.enableBotFlow == Just True) $ do
  rcHash <- getDbHash registrationNo
  RCQuery.updateVerifiedByCertificateNumberHash (Just allValid) rcHash
```

Consequence: if `allValid = False` (docs have become invalid), the function sets `verified = False` in the DB and *then* the outer `unless verified $ throwError` fires. The DB mutation happens before the throw. This is semantically correct (keeping state fresh) but is non-obvious and differs from a "read-only check." If `enableBotFlow = False`, no mutation occurs but the check still gates approval, which is also correct.

This is intentional and consistent with existing usage in `getOnboardingRegisterVehicleStatus`, but worth a comment at the call site.

---

### Positive side-effect (implicit improvement)

Using `DashboardCaller merchantOpCity.id` introduces a city-match guard inside `rcVerifyStatus`:

```haskell
whenJust mbPassedCityId $ \passedCityId ->
  when (passedCityId /= merchantOpCityId) $
    throwError (InvalidRequest $ "RC belongs to city: " <> show merchantOperatingCity.city)
```

Before this PR, an operator in city A could approve a vehicle whose RC resolves to city B. Now that would be blocked. This is a security improvement.

---

### Suggestion — PR description is empty

The description body, motivation, testing, and all checklist items are blank. Even a one-liner on what the bug was and how it was manually verified would help reviewers and future bisectors.

---

### No issues found in

- Type alignment: `DashboardCaller merchantOpCity.id`, `Nothing :: Maybe Text`, `Just reqId :: Maybe Text` — all match `rcVerifyStatus`'s signature.
- Pattern match `(_, verified, _, _)` correctly extracts `allValid :: Bool` (index 1).
- The new check is correctly placed **before** `QVRC.updateByPrimaryKey rcInfo {DVRC.approved = Just True}`, so approval only proceeds after live validation.
- `reqId` (the RC UUID string) is correctly passed as `mbRcId` — consistent with how `rcId = Kernel.Types.Id.Id reqId` is used just above.

PR #15803, opened 2026-07-19
backend/fix/payment not getting created during onInit

Hari-Krishna-Dupaguntla:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added conditional persistence of booking payment references and an initialization completion flag.
  * Added self-healing payment retry during booking status checks for eligible, initialized journeys.
* **Bug Fixes**
  * Payment creation failures are now caught and logged without failing the booking request.
  * Payment creation proceeds only when all related journey legs are marked initialized.
  * Improved device metadata handling by returning `UNKNOWN` when a device is present but lacks a device type.

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSStatus.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSStatus.hs
@@ -6,6 +6,7 @@ import qualified BecknV2.FRFS.Enums as Spec
 import BecknV2.FRFS.Utils
 import Control.Monad.Extra hiding (fromMaybeM)
 import qualified Data.Text as Text
+import qualified Domain.Action.Beckn.FRFS.OnInit as DOnInit
 import qualified Domain.Types.FRFSQuote as DFRFSQuote
 import qualified Domain.Types.FRFSQuoteCategory as FRFSQuoteCategory
 import Domain.Types.FRFSQuoteCategoryType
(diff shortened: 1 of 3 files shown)


Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSStatus.hs
```diff
@@ -102,7 +103,16 @@ frfsBookingStatus (personId, merchantId_) isMultiModalBooking withPaymentStatusR
   let commonPersonId = Kernel.Types.Id.cast @DP.Person @DPayment.Person person.id
   logInfo $ "Booking status: " <> show booking.status
   case booking.status of
-    DFRFSTicketBooking.NEW -> buildFRFSTicketBookingStatusAPIRes booking quoteCategories Nothing
+    DFRFSTicketBooking.NEW -> do
+      when (booking.journeyOnInitDone == Just True && booking.validTill > now) $ do
+        mbPaymentBooking <- B.runInReplica $ QFRFSTicketBookingPayment.findTicketBookingPayment booking
+        when (isNothing mbPaymentBooking) $ do
```
hkmangla: Why will this case happen?

hkmangla (changes requested):

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/FRFSStatus.hs
Hari-Krishna-Dupaguntla: Even after onInit completes, the payment is still not being created sometimes. In those cases, the execution never reaches the createPayment() function.
This only happens when booking.journeyOnInitDone != true. Ideally, it should happen in the first try only. if it didn't happened in the first try then we are again trying in the frfsStatus.

Hari-Krishna-Dupaguntla: I checked the cases for Yatri Sathi and Chennai One. In both cases, booking.journeyOnInitDone was true, but the payment was still not created.
If the execution had reached the createPayment() function, we would at least have seen the error logs for the failed cases. However, there are no such logs, which indicates that the execution never reached the createPayment() function.

hkmangla: Then it is not because of journeyOnInitDone, right?

Hari-Krishna-Dupaguntla: but that is the only condition is there

Hari-Krishna-Dupaguntla: We update journeyOnInitDone and immediately fetch the same booking back through KV to evaluate the condition. The write is sync to Redis but propagates async to PG/replica. the KV reads in Redis first with a replica fallback. when the redis lookup is missing the fallback returns the pre-update row so the gate saw the flag as not-done milliseconds after we set it, and silently skipped payment creation.

hkmangla: But if you do like this, we might end up creating double payment entries, right?

PR #15876, opened 2026-07-22
backend/feat: rider consent for sharing mobile number with driver

witcher-shailesh:
A rider now controls whether the driver sees their real mobile number. Default is off: drivers keep getting the cloud-telephony (exophone) masked number. The real number is shared only when **both** halves of an AND-gate agree: merchant config allows direct calling **and** the rider consented.

## Changes

| Layer | Change |
|---|---|
| **BECKN spec** (`beckn-spec`) | New `SHARE_MOBILE_NUMBER_WITH_DRIVER` tag in the `CUSTOMER_INFO` group |
| **Rider app — storage** | `SafetySettings.shareMobileNumberWithDriver :: Maybe Bool` (tri-state) + `...UpdatedAt` audit timestamp; nullable migrations, no backfill |
| **Rider app — API** | Field exposed on the existing emergency-settings endpoints (see below) |
| **Rider app — BECKN** | Consent resolved (`fromMaybe False`) at `on_init` and emitted as a person tag at `confirm` — value-add NPs only |
| **Driver app — storage** | `RiderDetails.shareMobileNumberWithDriver :: Bool` over a nullable column (existing rows read as `false`) + generated update query |
| **Driver app — BECKN** | Tag parsed fail-closed (absent/malformed/non-value-add-NP → `false`), persisted at confirm; repeat riders re-snapshotted every confirm — so consent changes apply from the next ride |
| **Driver app — the gate** | `getRiderMobileNumber` returns the real number only if `shouldShareRiderMobileNumber option consent` — i.e. `DirectCall/DualCall` **and** consent; plus a short-circuit that skips the DB read entirely for `AnonymousCall` (the config default, hot ride-polling path) |
| **Security hardening** | BPP refuses the consent tag from non-value-add BAPs — a third party can't assert consent for a rider we hold no record for |
| **Tests** | Driver app's first unit suite (12 examples: parser fail-closed + full 3×2 gate matrix), 5 beckn-spec tag tests, and the 45-step `PhoneShareConsentFlow` E2E suite (`./run-tests.sh phone-consent`) — all green, E2E verified on a live local stack |

**Untouched:** the exophone path. When the gate says no, the driver client falls back to `exoPhone` exactly as before — zero behaviour change for non-consenting riders.

## Consumer-side API changes (rider app)

Both are existing emergency-settings routes; `token` header (rider auth token).

**Read — render the toggle / decide whether to prompt:**
```
GET /profile/getEmergencySettings
```
```json
{ ..., "shareMobileNumberWithDriver": null }
```
The field is deliberately tri-state:
- `null` — rider has **never been asked** → show a first-time prompt
- `true` — consented → toggle on
- `false` — **explicitly declined** → toggle off, don't re-prompt

**Update — when the rider flips the toggle:**
```
PUT /profile/updateEmergencySettings
Content-Type: application/json

{ "shareMobileNumberWithDriver": true }
```
Partial body is fine (all fields optional) — sending just this field changes nothing else. Returns `{"result": "Success"}`. The backend stamps the consent audit timestamp automatically.

**Effect applies from the next booking's `confirm`** — mid-ride toggles don't affect the active ride. Driver side: `riderMobileNumber` in ride details is populated when the gate passes, `null` otherwise; the driver client should dial `riderMobileNumber` if present, else `exoPhone` (always present).

https://claude.ai/code/session_011HxFQapEewsHkJrzvhb2NR

## Summary by CodeRabbit

* **New Features**
  * Added rider consent control for sharing mobile number with drivers.
  * Consent is captured at confirm time, persisted with rider details, and drives whether the driver app can view the rider’s mobile number based on merchant calling option plus rider consent.
  * Updated rider-side profile/emergency settings to read and write this consent.
* **Bug Fixes**
  * Ensures consent is correctly stored and applied on subsequent (existing) riders and future rides.
* **Tests**
  * Added an end-to-end “Phone Share Consent” integration suite with local environments, seeding, and multi-ride scenarios covering enabled/disabled consent.


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Confirm.hs
```diff
@@ -118,8 +119,9 @@ handler merchant req validatedQuote = do
   unless (booking.status == DRB.NEW) $ throwError (BookingInvalidStatus $ show booking.status)
   let mbMerchantOperatingCityId = Just booking.merchantOperatingCityId
 
-  (riderDetails, isNewRider) <- SRD.getRiderDetails booking.currency merchant.id mbMerchantOperatingCityId req.customerMobileCountryCode req.customerPhoneNumber booking.bapId req.nightSafetyCheck
+  (riderDetails, isNewRider) <- SRD.getRiderDetails booking.currency merchant.id mbMerchantOperatingCityId req.customerMobileCountryCode req.customerPhoneNumber booking.bapId req.nightSafetyCheck req.shareMobileNumberWithDriver
   unless isNewRider $ QRD.updateNightSafetyChecks req.nightSafetyCheck riderDetails.id
```
hkmangla: Better have a single query for both updates.

hkmangla (changes requested):

witcher-shailesh:
On the two `safety_settings` columns: `SafetySettings` is a single spec in `lib/shared-services`, and both apps depend on it and maintain their own `safety_settings` table (the driver app stores the *driver's* own safety prefs there). So any field added to the shared spec generates a column in both schemas — same as `nightSafetyChecks` and every other field on that table; the two schemas always carry identical column sets.

The driver-app copy is unused (the driver app reads consent from `rider_details`, never from `safety_settings`), and moving the field out of `SafetySettings` to avoid the dead column would mean re-plumbing the rider emergency-settings API and abandoning the canonical home for privacy toggles — a worse trade than keeping it.

Good catch that the old name (`shareMobileNumberWithDriver`) read badly on a *driver's* settings, though. Renamed the field to `consentToShareMobileNumber` (tag `CONSENT_TO_SHARE_MOBILE_NUMBER`) — a generic consent name that reads correctly on either a rider's or a driver's `safety_settings`, applied consistently across the whole surface (spec, both DB columns, BECKN tag, rider API, driver persistence, gate). Pushed.

witcher-shailesh:
Done — collapsed the two updates into a single query. `updateNightSafetyChecks` and `updateConsentToShareMobileNumber` were only used here, together, so I replaced both with one `updateNightSafetyChecksAndConsent` (params `[nightSafetyChecks, consentToShareMobileNumber]`). The confirm path now does one `updateOneWithKV` — a single write to the row instead of two. Pushed.

hkmangla (approved):

PR #15920, opened 2026-07-23
backend/feat: sync driver operating city with ride end location

harshit-singla23:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

When a ride ends, compare the ride's end-location city against the driver's stored
`merchantOperatingCityId` (on `Person`) and update it on mismatch — so a driver's
operating city stays accurate as they relocate, without requiring a manual dashboard
update.

Implementation (`Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs`,
inside `endRideTransaction`):
- Reuses `Storage.Queries.Geometry.findGeometriesContainingGps` to resolve the ride's
  end lat/long into a city — the same geometry lookup `Domain.Action.UI.Driver.getCity`
  already uses for this exact purpose.
- Reuses `CQMOC.findByMerchantIdAndCity` to resolve that city into a `MerchantOperatingCity`.
- Reuses `QPerson.updateMerchantOperatingCityId` — the same call the dashboard's manual
  "change operating city" flow (`postDriverChangeOperatingCity`) already uses.
- No existing function signatures changed; purely additive.
- Skips `InterCity`/`CrossCity` rides, since those are *expected* to end in a different
  city than they started — applying the sync there would misclassify a normal cross-city
  trip as the driver having relocated.
- Runs in a fork (`syncDriverOperatingCityWithRideEndLocation`), so any failure (e.g. no
  geofence match for the end location) is best-effort and can never affect ride
  completion or the synchronous `end` API response.

### Additional Changes

## Motivation and Context

Drivers' `merchantOperatingCityId` currently only changes via manual dashboard action.
If a driver relocates and starts taking rides in a new city, their stored operating
city silently goes stale, which can affect operating-city-scoped config (driver pool,
fare policy, notifications, etc.). Syncing it automatically off the ride they just
completed keeps it accurate without operator intervention.

## How did you test it?

Tested manually against the local dev stack (no automated test added yet — see
checklist). `cabal build dynamic-offer-driver-app` passes clean with `-Werror`
(no new warnings).

End-to-end manual flow: created a driver via direct DB seeding (bypassing the local
OCR-mock onboarding pipeline, which isn't a fair proxy for prod OCR), ran full
search → select → accept → start → end cycles via curl against rider-app (8013) /
driver-app (8016) / location-tracking-service (8081):

1. **Cross-city ride**: OneWay ride, pickup in Bangalore, drop in Mysore. Driver's
   `merchant_operating_city_id` correctly updated from Bangalore's op-city id to
   Mysore's after `end`.
2. **Stale stored city, single-city ride**: reset driver's stored city to Bangalore,
   then ran a OneWay ride with pickup *and* drop both in Mysore (no city-crossing
   mid-ride). Still correctly synced to Mysore — confirms the fix triggers off the
   driver's actual location mismatch, not just a pickup/drop city difference within
   one ride.
3. In both cases, `ride`/`booking` still transitioned to `COMPLETED` and the `end`
   API still returned `200 Success` — confirms the new fork doesn't disturb the core
   end-ride path.
4. Verified via driver-app logs (`Fork syncDriverOperatingCityWithRideEndLocation`)
   and direct `person.merchant_operating_city_id` checks in Postgres.

Not yet run: negative test confirming an `InterCity`/`CrossCity` ride leaves the
operating city unchanged (guard is implemented and verified by code inspection, not
yet exercised end-to-end).

## Screenshot of test results

_(no screenshots — verification was via curl responses, driver-app log greps, and
direct psql queries; can attach terminal output if needed)_

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Driver operating-city changes are now processed automatically after eligible rides end.
  * Driver access, city settings, and document status are synchronized with the new operating city.
  * Drivers receive relevant notifications, billing tags, and missing-document reminders when applicable.

* **Bug Fixes**
  * Prevented duplicate migration processing within a 24-hour period.
  * Missing-document reminders are deduplicated and expire after 14 days.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
@@ -251,6 +251,10 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
 
   fork "processEndRideFinance" $ processEndRideFinance merchant ride booking newFareParams driverId driverInfo thresholdConfig
 
+  -- Driver operating-city migration on relocation is handled by kafka-consumers'
+  -- RIDE_EVENTS_CONSUMER (Processor.RideEvents.Handlers.handleDriverCityMigration), off the
+  -- RideEndedEvent published below -- not here. See that module for the full design.
+
   let validRide = isValidRide ride
   -- Publish RideEndedEvent to Redis Stream "ride.events.shard<N>". kafka-consumers
   -- RIDE_EVENTS_CONSUMER fans this out to: analytics Kafka events, ride-interpolation,
--- a/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
+++ b/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
@@ -15,19 +15,27 @@ module Processor.RideEvents.Handlers
     handleRideEndNotifications,
     handleLeaderboard,
     handleReferral,
+    handleDriverCityMigration,
   )
 where
 
 import qualified Data.Aeson as A
+import qualified "dynamic-offer-driver-app" Domain.Action.UI.Registration as DReg
 import "dynamic-offer-driver-app" Domain.Action.UI.Ride.EndRide (RideInterpolationData (..))
 import qualified "dynamic-offer-driver-app" Domain.Types.Booking as SRB
+import qualified "dynamic-offer-driver-app" Domain.Types.DocumentVerificationConfig as DTO
 import "dynamic-offer-driver-app" Domain.Types.Event.RideEndedEvent (RideEndedEvent (..))
+import qualified "dynamic-offer-driver-app" Domain.Types.Merchant as DM
+import qualified "dynamic-offer-driver-app" Domain.Types.MerchantOperatingCity as DMOC
+import qualified "dynamic-offer-driver-app" Domain.Types.Person as DP
 import qualified "dynamic-offer-driver-app" Domain.Types.Ride as Ride
 import qualified "dynamic-offer-driver-app" Domain.Types.RideRelatedNotificationConfig as DRN
 import "dynamic-offer-driver-app" Domain.Types.TransporterConfig (TransporterConfig)
+import qualified "beckn-spec" Domain.Types.Trip as DTrip
 import Kernel.Beam.Lib.Utils (pushToKafka)
 import Kernel.External.Encryption (EncFlow)
 import qualified Kernel.External.Encryption as EncFlow
+import qualified Kernel.External.Notification.FCM.Types as FCM
 import Kernel.External.Types (SchedulerFlow)
 import Kernel.Prelude
 import qualified Kernel.Storage.Clickhouse.Config as CHConfig
@@ -38,10 +46,12 @@ import Kernel.Streaming.Kafka.Producer.Types (HasKafkaProducer, KafkaProducerToo
 import qualified Kernel.Tools.Metrics.CoreMetrics as CoreMetrics
 import Kernel.Types.Common (MonadFlow)
 import Kernel.Types.Confidence (Confidence (..))
+import qualified Kernel.Types.Documents as Documents
 import Kernel.Types.Id
 import Kernel.Utils.Common
   ( CacheFlow,
     HasShortDurationRetryCfg,
+    Hours,
     fromMaybeM,
     getCurrentTime,
     getLocalCurrentTime,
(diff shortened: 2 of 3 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
```diff
@@ -251,6 +254,30 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
...
+        DTrip.InterCity _ _ -> True
+        DTrip.CrossCity _ _ -> True
+        _ -> False
+  unless isInterOrCrossCityRide $
+    fork "syncDriverOperatingCityWithRideEndLocation" $
+      whenJust ride.tripEndPos $ \tripEndPos -> do
+        geoms <- QGeometry.findGeometriesContainingGps tripEndPos
+        let mbEndCity = case filter (\geom -> geom.city /= Context.City "AnyCity") geoms of
+              (g : _) -> Just g.city
+              [] -> (.city) <$> find (\geom -> geom.city == Context.City "AnyCity") geoms
```
hkmangla: In this case we can consider it Nothing. No need to look for AnyCity geom.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
```diff
@@ -251,6 +254,30 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
...
+        whenJust mbEndCity $ \endCity -> do
+          mbEndLocationOpCity <- CQMOC.findByMerchantIdAndCity merchant.id endCity
+          whenJust mbEndLocationOpCity $ \endLocationOpCity -> do
+            driverPerson <- QPerson.findById (cast driverId) >>= fromMaybeM (PersonNotFound driverId.getId)
+            when (driverPerson.merchantOperatingCityId /= endLocationOpCity.id) $
+              QPerson.updateMerchantOperatingCityId (cast driverId) endLocationOpCity.id
```
hkmangla: We might need to update their plans etc etc?
@dushyant-2002 If we just change the city of the driver like that will it work? 

@harshit-singla23 We might need to check if this driver will qualify for all the document required for that city. Because different city can have different requireemnts.

hkmangla (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
dushyant-2002: This should work because we already have a fallback to the planId that was associated with the driver at the time of endRide.

One thing to keep in mind: if a driver moves to another city and selects a plan for that city, we update both driverPlan and driverFee.cityId. 
@piyushKumar-1  i think we have to add that driver to activeDriverSet of that city also then?

dushyant-2002: we should test it thoroughly in master

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
```diff
@@ -251,6 +256,69 @@ endRideTransaction driverId booking ride mbFareParams mbRiderDetailsId newFarePa
 
   fork "processEndRideFinance" $ processEndRideFinance merchant ride booking newFareParams driverId driverInfo thresholdConfig
 
+  -- Best-effort: keep the driver's operating city in sync with where they actually ended
+  -- the ride. Reuses the same geometry-lookup query that resolves lat/long -> city
+  -- elsewhere (e.g. Domain.Action.UI.Driver.getCity). Skipped for InterCity/CrossCity
+  -- rides, which are *expected* to end outside the driver's operating city and don't
+  -- imply the driver has relocated there. Forked so any failure here (e.g. no geofence
+  -- match for the end location) never affects ride completion.
+  let isInterOrCrossCityRide = case ride.tripCategory of
```
hkmangla: Instead of figuring out the city from the trip end lat longs and checking for intercity, lets just check consider the ride.merchantOperatingCityId as the city id, and check it with driver city id, if matched then do nothing else if merchantId is matched then update the the city the way it is being done now, but if merchantid is also not matched, then we should throw error

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs
hkmangla: @harshit-singla23 Get it checked with @piyushKumar-1 once.

hkmangla (changes requested):

hkmangla (dismissed):

hkmangla (approved):

PR #15928, opened 2026-07-24
endotp requirement table

sathikabegum-nammayatri:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * End-ride OTP requirements can now be configured by operating city, trip category, and trip mode.
  * Ride start and end flows use these settings, with existing defaults preserved when no override is configured.
  * Added database storage, caching, and seed data for end-OTP configuration.
* **Bug Fixes**
  * Improved handling of trip category and mode combinations when determining OTP requirements.


kavya-shree-s (commented):
## Review: endotp requirement table

Reviewed the full diff and verified claims against the repo. The design is sound and the seed data is carefully behavior-preserving, but two things need fixing before merge.

---

### CRITICAL

**[C-1] The seed migration will never run — the feature is a silent no-op**

`dev/seed-migrations/<svc>` is only executed if that path is listed in the service's `migrationPath`. `Backend/dhall-configs/dev/dynamic-offer-driver-app.dhall:490-495` lists only three paths:

```dhall
migrationPath =
  [ "dev/migrations-read-only/dynamic-offer-driver-app"
  , env:DYNAMIC_OFFER_DRIVER_APP_MIGRATION_PATH as Text ? "dev/ddl-migrations/dynamic-offer-driver-app"
  , "dev/migrations-after-release/dynamic-offer-driver-app"
  ]
```

Compare `provider-dashboard.dhall:134-139`, which *does* include `"dev/seed-migrations/provider-dashboard"`. `dev/seed-migrations/` also currently contains only `provider-dashboard`, `rider-dashboard`, `safety-dashboard` — this is the first app-level seed dir.

Consequence: `end_otp_config` stays empty everywhere → every lookup returns `Nothing` → `checkEndOtpRequired` always falls through to `DTC.isEndOtpRequired`. The PR does nothing at runtime, and the silent fallback means nobody notices.

**Fix:** add `"dev/seed-migrations/dynamic-offer-driver-app"` to `migrationPath`, after the ddl-migrations entry (matching the dashboard convention).

---

**[C-2] End-ride re-resolves the OTP policy, so a config edit mid-ride can wedge a ride**

`src/Domain/Action/UI/Ride/StartRide.hs:236` and `src/Domain/Action/UI/Ride/EndRide.hs:342-344`

Before this PR `DTC.isEndOtpRequired` was pure, so start and end always agreed. Now the answer can change between the two — a row edit, or cache TTL expiry on a row inserted after the ride started.

Failure path: policy flips `false → true` mid-ride. `rideOld.endOtp` is `Nothing`, but end-ride now takes the `endOtpRequired` branch. If the driver submits any OTP, `Just endRideOtp == rideOld.endOtp` is `Just x == Nothing` = False → `throwError IncorrectOTP` on every retry. The driver cannot end the ride. The only thing preventing this from being guaranteed is the `Nothing -> pure ()` escape at `EndRide.hs:343`.

**Fix:** snapshot the decision at ride start — gate end-ride validation on `rideOld.endOtp` being present rather than re-reading config.

---

### IMPORTANT

**[I-1] No negative caching → two extra Postgres round-trips per ride**
`src/Storage/CachedQueries/EndOtpConfig.hs:31-40` — `getEndOtpConfigFromDB` caches only the `Just` case. `end_otp_config` is not in `enableKVForWriteAlso`/`enableKVForRead` in `kv_configs`, so `findOneWithKV` goes to Postgres. Combined with C-1 (empty table), *every* startRide and endRide pays an uncached query returning nothing. Cache the miss with a sentinel, or load all rows for a city once.

**[I-2] No `clearCache`, no invalidation path**
Every other cached-queries module in this service exposes one (`RideRelatedNotificationConfig.hs:59`, `CoinsConfig`, `Exophone`, `FareProduct`, …). Since there's no API to write this table, the only way to change a value is direct SQL — after which pods serve stale values until `configsExpTime`. Also worth adding an explicit export list; the module is a bare `where`, unlike its siblings.

**[I-3] `CachedQueries.EndOtpConfig.create` is dead code**
Nothing calls it, and there's no dashboard endpoint or API spec for writing this table. Either add the write path (if operators are meant to edit it) or drop `create`.

**[I-4] Unrelated migration rename bundled in**
`dev/feature-migrations/0041-cancellation-commission-helsinki.sql` → `0042-...`. This does fix a pre-existing duplicate number (`0041-loyalty-program-map-conversion-rate.sql` also exists on main), but it's unrelated to end-OTP config, and renaming a migration file that environments may already have applied risks it re-running. Worth splitting out, or confirming the runner tracks by content rather than filename.

**[I-5] No build verification for this service**
Checks that ran: `hlint` ×3, `check-conditions`, `CodeRabbit`, and "Incremental Repl Check — **Dashboard**". There's no `dynamic-offer-driver-app` compile check. Given `-Werror`, a new SharedLogic module, and a `beckn-spec` change, please run `cabal build dynamic-offer-driver-app` locally and confirm.

---

### Good to have

**[G-1] Three comments in `Trip.hs` are now false**
- `Trip.hs:47` — `EasyBooking EasyBookingMode -- ... end-OTP requirement is hardcoded below, not configurable`
- `Trip.hs:380` — `-- Move it to configs later if required`
- `Trip.hs:384` — `-- hardcoded (not config-driven); flip this one word + redeploy if you want it True instead`

All three describe exactly the mechanism this PR builds. Worth updating them to point at `SharedLogic.EndOtpConfig.checkEndOtpRequired`, since calling `DTC.isEndOtpRequired` directly is now the wrong entry point.

**[G-2] `excludedFields` lists `tripCategory` and `tripMode`**
`spec/Storage/EndOtpConfig.yaml:31` — `excludedFields` suppresses the generator's *auto-added* default fields (`merchantId`, `merchantOperatingCityId`, `createdAt`, `updatedAt`; cf. `DriverStats.yaml:235`). `tripCategory`/`tripMode` are explicitly declared, so listing them is a no-op that will mislead the next reader.

**[G-3] `trip_category` / `trip_mode` are free-text with no validation**
A typo'd row (`'Intercity'`, `'RideOTP'`) is indistinguishable from "not configured" and silently falls back. Consider a CHECK constraint, or at minimum a `logWarning` on the `Nothing` branch so misconfiguration is observable.

**[G-4] Dead commented code adjacent to the change**
`EndRide.hs:345`: `-- throwError $ EndRideOtpRequired (show booking.tripCategory)`. Pre-existing, but it sits directly under the modified block — and it's the enforcement C-2 is really about.

**Note (not a defect in this PR):** the end-OTP check is bypassable today — a client that simply omits `endRideOtp` skips validation entirely (`Nothing -> pure ()`). Building a config table around an optional check gives an illusion of enforcement; worth confirming with product.

---

### Verified as correct

- **Seed data is exactly behavior-preserving.** All 26 rows reproduce `isEndOtpRequired` for every `TripCategory` × mode combination (`OneWayMode` has 4 constructors incl. `MeterRide`, `TripMode` has 2): Rental 2 true, InterCity 4 true, Delivery 4 true, EasyBooking 2 false, OneWay 4 false, RideShare 2 false, CrossCity 4 false, Ambulance 4 false. The `MeterRide` rows called out in the comment are genuinely needed. `Show` on both mode types is derived, so the seed text matches `tripCategoryAndModeText`'s `show mode` output.
- **`tripCategoryAndModeText` is total** — all 8 `TripCategory` constructors covered. Dropping InterCity/CrossCity's city `Text` is correct since the table is already keyed by `merchantOperatingCityId`.
- **Index:** the query filters on exactly the composite PK `(merchant_operating_city_id, trip_category, trip_mode)` — fully covered, no extra index needed.
- **Migration numbering:** `0846` is the next free number (`0845` is highest). No collision.
- **DB back-compat:** new table only; no `DROP`/`RENAME`/`ALTER TYPE`. `ADD COLUMN ... NOT NULL` without `DEFAULT` is safe here because `CREATE TABLE` precedes it in the same file, matching the generated house style.
- No Redis pattern-matching, no PII fields, no API spec changes (so no UI compatibility risk).
- `merchantOperatingCityId` is a primary key here, not a secondary key — the secondary-key deny list doesn't apply.

---

### Impact

| Aspect | Status |
|--------|--------|
| Touches ride booking flow | **YES** — StartRide + EndRide |
| Has DB migrations | YES (new table + seed; back-compat OK) |
| API changes | NO |
| Commits | 1 (no squash needed) |
| Risk level | **MEDIUM-HIGH** |

BECKN actions potentially affected: `on_update` (RIDE_STARTED / RIDE_ENDED). No search/select/init/confirm impact.

**QA should test:**
1. Every trip category (OneWay, Rental, RideShare, InterCity, CrossCity, Ambulance, Delivery, EasyBooking) start→end with the table **empty** — behavior must be identical to today.
2. Same, with the seed applied — must still be identical.
3. Flip one row (e.g. `EasyBooking`/`RideOtp` → true) and verify a *newly started* ride demands an end OTP, and that the change takes effect only after cache TTL (see I-2).
4. **Start a ride, flip the config, then end it** — this is the C-2 path.
5. End-ride via dashboard and cron paths (they skip the OTP block entirely) — confirm unaffected.

Issue #10, opened 2023-02-27
Drop stack files


srid:
Once we are confident that all developers can use Nix (#9), remove the `stack.*` files from the repo.

Also,

- [ ] Remove the `FromDhall Word16` instance in mobility-core (https://github.com/nammayatri/shared-kernel/pull/12/files#diff-8e1df6643baeaf8e2830052f24f9dbc719727c85be219deb7c6022e5c2bdaf1dR59)

PR #15942, opened 2026-07-24
Backend/fix/reduce 5xx errors batch 2

Praveenkumar02023:
## Type of Change
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description

Two fixes to prevent false E500 errors from background fork threads:

**1. Guard `checkBookingsForStatus` against bookings without `bppBookingId`** (`Booking.hs`)

When a rider opens their booking list, a background thread (`checkBookingsForStatus`) iterates over all bookings to sync stale ones with the BPP via `callOnStatus`. Bookings in `NEW` status don't have a `bppBookingId` yet (the BPP hasn't confirmed them), so `callOnStatus` → `buildStatusReqV2` crashes with `E500 BOOKING_FIELD_NOT_PRESENT` when it tries to extract the null field.

Fix adds an early guard using `case currBooking.bppBookingId of Nothing -> skip`. Bookings without a `bppBookingId` can't be synced with a BPP that doesn't know about them — skipping is the correct behavior.

**2. Catch `VehicleNotFound` in `syncServiceTiersOnRatingChange`** (`Rating.hs`)

Riders can rate a ride days or weeks after completion. The rating fork saves the rating and recalculates the driver's average successfully, but as a final step calls `syncServiceTiersOnRatingChange` → `fetchVehicleTierForDriverWithUsageRestriction`, which strictly expects the driver's vehicle to exist. If the driver has unlinked their vehicle since the ride, `VehicleNotFound` is thrown unhandled, killing the fork and logging E500.

Fix wraps the `fetchVehicleTierForDriverWithUsageRestriction` call in `try` and pattern-matches `VehicleNotFound` — if no vehicle exists, we log a warning and skip the tier sync (which isn't needed without a vehicle). Any other exception is rethrown.

### Additional Changes

## Motivation and Context

Both errors are false E500s inflating the Grafana 5xx dashboard. Neither represents an actual server failure — they are expected conditions (booking not yet confirmed by BPP, driver vehicle unlinked after ride completion) that the code should handle gracefully instead of crashing the fork thread.

## How did you test it?

- Built both `rider-app-exe` and `dynamic-offer-driver-app-exe` locally with `cabal build` — all modules compiled and linked successfully.

## Screenshot of test results

N/A — error handling changes only, no new logic.

## Checklist
- [x] I formatted the code and addressed linter errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rating updates when vehicle information is unavailable by safely skipping service-tier synchronization and recording a warning.
  * Prevented status synchronization attempts for bookings without a booking provider reference.
  * Preserved status updates for eligible bookings based on configured timing thresholds.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Rating.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Rating.hs
@@ -171,9 +171,14 @@ syncServiceTiersOnRatingChange ::
   m ()
 syncServiceTiersOnRatingChange driverStats newRating personId merchantOpCityId = do
   let updatedDriverStats = driverStats {DDriverStats.rating = newRating}
-  tierResults <- fetchVehicleTierForDriverWithUsageRestriction SelectedServiceTiers Nothing Nothing (Just updatedDriverStats) Nothing personId merchantOpCityId
-  let newTiers = (.serviceTierType) . fst <$> filter (not . snd) tierResults
-  QVehicle.updateSelectedServiceTiers newTiers personId
+  try (fetchVehicleTierForDriverWithUsageRestriction SelectedServiceTiers Nothing Nothing (Just updatedDriverStats) Nothing personId merchantOpCityId) >>= \case
+    Left (VehicleNotFound _) -> do
+      logWarning $ "Vehicle not found for driver " <> personId.getId <> ". Skipping service tier sync on rating change."
+      pure ()
+    Left err -> throwError err
+    Right tierResults -> do
+      let newTiers = (.serviceTierType) . fst <$> filter (not . snd) tierResults
+      QVehicle.updateSelectedServiceTiers newTiers personId
 
 calculateAverageRating ::
   (CacheFlow m r, EsqDBFlow m r, EncFlow m r, Redis.HedisFlow m r, HasField "serviceClickhouseCfg" r CH.ClickhouseCfg, HasField "serviceClickhouseEnv" r CH.ClickhouseEnv) =>
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Booking.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Booking.hs
@@ -168,20 +168,23 @@ callOnStatus currBooking = do
 
 checkBookingsForStatus :: [SRB.Booking] -> Flow ()
 checkBookingsForStatus (currBooking : bookings) = do
-  riderConfig <- getConfig (RiderConfigDimensions {merchantOperatingCityId = currBooking.merchantOperatingCityId.getId}) (Just (CQRC.findByMerchantOperatingCityId currBooking.merchantOperatingCityId)) >>= fromMaybeM (RiderConfigDoesNotExist currBooking.merchantOperatingCityId.getId)
-  case (riderConfig.bookingSyncStatusCallSecondsDiffThreshold, currBooking.estimatedDuration) of
-    (Just timeDiffThreshold, Just estimatedEndDuration) -> do
-      now <- getCurrentTime
-      let estimatedEndTime = DT.addUTCTime (fromIntegral estimatedEndDuration.getSeconds) currBooking.startTime
-      let diff = DT.diffUTCTime now estimatedEndTime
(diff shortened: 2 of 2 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Rating.hs
```diff
@@ -171,9 +171,14 @@ syncServiceTiersOnRatingChange ::
   m ()
 syncServiceTiersOnRatingChange driverStats newRating personId merchantOpCityId = do
   let updatedDriverStats = driverStats {DDriverStats.rating = newRating}
-  tierResults <- fetchVehicleTierForDriverWithUsageRestriction SelectedServiceTiers Nothing Nothing (Just updatedDriverStats) Nothing personId merchantOpCityId
-  let newTiers = (.serviceTierType) . fst <$> filter (not . snd) tierResults
-  QVehicle.updateSelectedServiceTiers newTiers personId
+  try (fetchVehicleTierForDriverWithUsageRestriction SelectedServiceTiers Nothing Nothing (Just updatedDriverStats) Nothing personId merchantOpCityId) >>= \case
```
sidsethu, line 174: Correct and type-safe — `Left (VehicleNotFound _)` pins `try` to `VehicleError` (an `HTTPException` child with an `Exception` instance), so this catches exactly `VehicleNotFound`, rethrows other `VehicleError`s, and lets everything else propagate. 👍

Non-blocking nit: consider the explicit `try @_ @VehicleError (...)` idiom (as in `IdfyWebhook.hs:281`). Same behavior, but it makes the caught type obvious and prevents a future refactor from silently widening what gets swallowed.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Booking.hs
```diff
@@ -168,20 +168,23 @@ callOnStatus currBooking = do
...
+          let estimatedEndTime = DT.addUTCTime (fromIntegral estimatedEndDuration.getSeconds) currBooking.startTime
+          let diff = DT.diffUTCTime now estimatedEndTime
+          let callStatusConditionNew = (currBooking.status == SRB.NEW && diff > fromIntegral timeDiffThreshold) || (currBooking.status == SRB.CONFIRMED && diff > fromIntegral timeDiffThreshold)
+              callStatusConditionTripAssigned = currBooking.status == SRB.TRIP_ASSIGNED && diff > fromIntegral timeDiffThreshold
+          when callStatusConditionNew $ do
+            callOnStatus currBooking
+          when callStatusConditionTripAssigned $ do
+            callOnStatus currBooking
+          checkBookingsForStatus bookings
+        (_, _) -> logError "Nothing values for time diff threshold or booking end duration"
```
sidsethu, line 187: Pre-existing, not introduced by this PR — flagging for a follow-up: this `(_, _) -> logError …` arm doesn't recurse into `checkBookingsForStatus bookings`, so a booking that *has* a `bppBookingId` but whose config threshold or `estimatedDuration` is `Nothing` still terminates the whole list walk early (the same class of "one booking halts the rest" bug your `Nothing` guard just fixed for the null case). Worth making every terminal arm recurse in a later cleanup.

sidsethu (approved):
Approve — both fixes are correct, and #6 is actually more thorough than the minimum we scoped.

**#10 (Rating.hs)** — wraps the tier fetch in `try … >>= \case`, swallows only `VehicleNotFound` (warn + skip), rethrows other `VehicleError`s, and lets everything else propagate. Verified the bare `try` type-infers to `VehicleError` correctly (it's an `HTTPException` child with an `Exception` instance), so it compiles and actually catches. One non-blocking idiom suggestion inline.

**#6 (Booking.hs)** — the top-level `bppBookingId` guard sits above *all* `callOnStatus` calls, so it protects the `buildStatusReqV2` crash site for NEW/CONFIRMED/TRIP_ASSIGNED, not just NEW. And because the `Nothing` branch recurses, it also fixes the latent "one unconfirmed booking crashes the fork and kills status-sync for every later booking" issue. Nice — this is better than the minimal condition-guard.

One pre-existing follow-up flagged inline (a sibling `logError` arm that doesn't recurse) — not introduced here, not blocking.

Also good to see the branch/PR naming now following the `backend/fix/…` convention. Note visible CI is HLint + Dashboard REPL only, so the rider-app/driver-app compile rests on your local `cabal build` — fine for changes this size.

PR #16003, opened 2026-07-30
backend/feat: resolve driver calling number server-side

witcher-shailesh:
Driver ride response now returns **one field saying which number to dial**, so the app never picks:

```json
"callingNumber": { "number": "9876543210", "countryCode": "+91", "numberType": "DIRECT" }
```

Fixes the call button silently doing nothing when a rider declined consent (app dialled `undefined`).

### Flow

```mermaid
flowchart TD
  S([Driver fetches ride]) --> O{"driverCallingOption<br/>= DirectCall / DualCall?"}
  O -->|no| X["callingNumber = exophone · ANONYMOUS<br/>riderMobileNumber = null"]
  O -->|yes| A{"ride still active?"}
  A -->|no| X
  A -->|yes| B["riderMobileNumber = real number<br/>(never gated on consent)"]
  B --> T{"CONSENT_TO_SHARE_MOBILE_NUMBER<br/>tag from BAP"}
  T -->|absent = flow off| D["callingNumber = real number · DIRECT"]
  T -->|True| D
  T -->|False| E["callingNumber = exophone · ANONYMOUS"]
  T -->|malformed| E
```

### Dry run

| # | Merchant option | Consent flow | Rider chose | `riderMobileNumber` | `callingNumber` |
|---|---|---|---|---|---|
| 1 | DirectCall | **off** (default) | — | real number | real number · `DIRECT` |
| 2 | DirectCall | on | share | real number | real number · `DIRECT` |
| 3 | DirectCall | on | decline | real number | exophone · `ANONYMOUS` |
| 4 | AnonymousCall | any | any | `null` | exophone · `ANONYMOUS` |

Row 1 = today's behaviour, so **nothing changes until a merchant opts in.**

### Notes

- **One switch:** `RiderConfig.enableShareNumberWithDriver` (default `false`) gates the whole feature — it drives the consumer-app consent UI *and* whether the BECKN tag is emitted. The BPP holds no config; it infers from the tag. Enable per city with one `rider_config` update.
- **`riderMobileNumber` is deliberately consent-free** (as before #15876). Every released driver build reads it, so re-gating it would have silently changed their behaviour. Consent gates only `callingNumber`.
- `number` is bare like `riderMobileNumber`/`exoPhone`; `countryCode` is separate (`null` when `ANONYMOUS`). No migration for `RiderDetails` — column was already nullable.

### Testing

`cabal build all` green. `PhoneShareConsentFlow` E2E **run locally and passing** across all three consent states.

Client: ny-react-native#7267.


witcher-shailesh:
### Summary of changes

**1. `callingNumber` on the driver ride response** — one always-populated object saying which number to dial, so the app never has to choose:

```json
"callingNumber": { "number": "9876543210", "countryCode": "+91", "numberType": "DIRECT" }
```

`DIRECT` only when the merchant allows direct calling **and** the rider consented **and** the ride is active; otherwise `ANONYMOUS` with the exophone and `countryCode: null`. `number` is bare, same format as `riderMobileNumber`/`exoPhone`.

**2. `riderMobileNumber` is consent-free again** (as before #15876) — released driver builds read it, so re-gating it would have silently changed their behaviour. Consent gates only `callingNumber`.

**3. One switch for the whole feature** — `RiderConfig.enableShareNumberWithDriver`, default `false`:
- exposed on the emergency-settings response, so the consumer app shows the consent row only when the backend enables it (replaces the app-side config key)
- when off: `on_init` → `Nothing`, confirm omits the `CONSENT_TO_SHARE_MOBILE_NUMBER` tag, BPP shares the number as before consent existed
- BPP infers from the tag instead of its own config — absent → flow off, `True`/`False` → rider's choice, malformed → mask
- `RiderDetails.consentToShareMobileNumber` → `Maybe Bool`; column already nullable, **no migration**

**Rollout:** consent ships disabled. Enable per city with one UPDATE on `atlas_app.rider_config` (ConfigPilot-cached — needs a Redis flush **and** a rider-app restart to take effect).

**Testing:** `cabal build all` green. `PhoneShareConsentFlow` covers all three consent states and seeds both flags. Not covered, documented in `Rules.md`: the flow-off case and the merchant kill switch — both need config flips mid-collection that the in-process cache makes impossible.

**Follow-up:** `ny-react-native` — dial `callingNumber.number`, drop `enableDirectCalling` and the app-side `flowConfig.enableShareNumberWithDriver`.

PR #16060, opened 2026-08-03
backend/feat: pass eligibility on AppDynamicLogic + dashboard pass ca…

Cyanid0:
…talog CRUD

Moves pass purchase eligibility off per-pass JSON logic onto the PASS_PURCHASE_ELIGIBILITY dynamic-logic domain, registers that domain with the dashboard, and adds CRUD so the pass catalog can be managed without raw SQL.

Eligibility
  Replaces the inline A.object projection in buildPassAPIEntity with a typed
  SharedLogic.Pass.Eligibility.PassEligibilityData, matching how
  CANCELLATION_REASONS / PICKUP_ETA_CALCULATION / OFFERS_FRAUD_CHECKS build
  their runLogics input. Field names and JSON serialisation are preserved so
  rules backfilled from pass.purchase_eligibility_json_logic evaluate
  identically. Open-on-empty, closed-on-error. The rules fetch is hoisted out of
  the per-pass loop. A migration unnests each pass's rules into
  app_dynamic_logic_element, wrapping each in a passEnum/passCode discriminator
  so one per-MOC ruleset serves every pass.

Dashboard registration
  The domain was reachable from code but not from the NammaTag dashboard: both
  per-domain dispatch sites fell through to their catch-alls, so a ruleset could
  only be created via SQL. Adds verifyAndUpdateDynamicLogic and DomainSchemaResp
  cases, plus the ToSchema/FromJSON/Default instances they need.

Pass catalog CRUD
  GET    /pass/catalog/list?passTypeId=&enable=
  POST   /pass/catalog/create              -> {passId}
  PUT    /pass/catalog/{passId}/update     partial patch, all fields optional
  DELETE /pass/catalog/{passId}/delete

  Delete is soft (enable = false), matching deleteEDCMachine; a hard delete
  would orphan purchased_pass rows referencing pass_id. Create/update omit the
  legacy eligibility JSON columns, since purchase rules now live in the
  dynamic-logic domain and the redeem column has no reader. Every mutation busts
  CQPass's per-(passTypeId, enable) cache -- update clears old and new keys,
  delete clears both halves. All four endpoints enforce city scoping, and
  duplicate pass codes are rejected per city because the ruleset falls back to
  passCode as its discriminator when a pass type has no passEnum.

Known limit: PassUpdateReq uses <|> semantics, so a patch can set a nullable field but cannot clear one back to null.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
[...]

PR #16081, opened 2026-08-04
backend/fix: use riderConfig.timeDiffFromUtc for pass eligibility rollout selection

Cyanid0:
Stacked on #16060 — the line this touches is introduced by that PR, so this targets its branch rather than `main`. Rebase onto `main` once #16060 lands, or fold it in there if you'd rather keep it to one PR.

## Problem

`getMultimodalPassAvailablePasses` hardcodes the IST offset when deriving the local time it hands to `getAppDynamicLogic`:

```haskell
localTime <- getLocalCurrentTime (19800 :: Seconds)
```

That `localTime` is what `selectAppDynamicLogicVersion` matches against a rollout's `timeBounds` to pick a version. So for any operating city that isn't IST, the version is selected against the wrong wall clock — and once `PASS_PURCHASE_ELIGIBILITY` has more than one time-bounded rollout, that means resolving the wrong ruleset. Today's rollouts are `Unbounded`, so nothing misbehaves yet; this closes the gap before one is added.

## Fix

Use the idiom this same file already applies in four other places (lines 219, 764, 969, 1035) — read `riderConfig` for the person's operating city, fall back to 19800 only when absent:

```haskell
mbRiderConfig <- getConfig (RiderConfigDimensions {merchantOperatingCityId = person.merchantOperatingCityId.getId}) (Just (CQRC.findByMerchantOperatingCityId person.merchantOperatingCityId))
let timeDiffFromUtc = maybe (Seconds 19800) (.timeDiffFromUtc) mbRiderConfig
localTime <- getLocalCurrentTime timeDiffFromUtc
```

Every identifier used is already imported; no new dependencies. It was the only `getLocalCurrentTime` call site in rider-app passing a literal.

## Type of Change

- [x] Bugfix

## Additional Changes

## How did you test it?

Not runtime-tested — behaviour is unchanged for IST cities, which is every city with passes today. The fetch mirrors line 225 exactly, and CI covers compilation.

## Checklist

- [x] I formatted the code and addressed linter errors

### DIFF
--- a/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Pass.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Pass.hs
@@ -133,7 +133,9 @@ getMultimodalPassAvailablePasses (mbPersonId, _merchantId) mbLanguage = do
   when (null passCategories) $
     logError $ "getMultimodalPassAvailablePasses: no pass categories for mocId " <> person.merchantOperatingCityId.getId
 
-  localTime <- getLocalCurrentTime (19800 :: Seconds)
+  mbRiderConfig <- getConfig (RiderConfigDimensions {merchantOperatingCityId = person.merchantOperatingCityId.getId}) (Just (CQRC.findByMerchantOperatingCityId person.merchantOperatingCityId))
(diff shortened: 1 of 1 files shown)


Cyanid0:
Folding this into #16060 directly instead — the change is three lines on a file that PR already touches, so a stacked PR was more ceremony than it was worth.

PR #16093, opened 2026-08-04
backend/fix/ride-tags-write-in-consumer

H-C-21:
Moved Ride Tags Calculation Back to End Ride (Synchronous)

## Summary by CodeRabbit

* **Bug Fixes**
  * Ride tags now use the ride’s actual trip end time to determine the local end day and duration.
  * Successfully computed tags are now saved and merged with existing ride tags.
  * Ride tags are now computed and applied during ride completion, improving their availability immediately after the ride ends.

### DIFF
--- a/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
+++ b/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
@@ -195,8 +195,8 @@ handleNammaTags ev = withRideAndBooking ev $ \ride booking -> do
         isJust mbDriverMobileHash
           && isJust mbRiderMobileHash
           && mbDriverMobileHash == mbRiderMobileHash
-  now <- getCurrentTime
-  let merchantLocalDay = utctDay $ addUTCTime (secondsToNominalDiffTime thresholdConfig.timeDiffFromUtc) now
+  rideEndTime <- maybe getCurrentTime pure ride.tripEndTime
+  let merchantLocalDay = utctDay $ addUTCTime (secondsToNominalDiffTime thresholdConfig.timeDiffFromUtc) rideEndTime
   priorRidesSameCustomer <-
     QRide.countPriorCompletedRidesWithSameCustomer
       (cast ride.driverId)
@@ -208,8 +208,8 @@ handleNammaTags ev = withRideAndBooking ev $ \ride booking -> do
         riderBlockedForCoins
           || priorRidesSameCustomer > thresholdConfig.sameRiderDriverRideCountThreshold
       rideDurationSeconds =
-        maybe 0 (\tStart -> max 0 $ roundToIntegral (diffUTCTime now tStart)) ride.tripStartTime
-  void $
+        maybe 0 (\tStart -> max 0 $ roundToIntegral (diffUTCTime rideEndTime tStart)) ride.tripStartTime
+  nammaTags <-
     withTryCatch "ride-events:computeNammaTags" $
       LYDL.computeNammaTagsWithDebugLog
         LYDL.Driver
@@ -217,6 +217,7 @@ handleNammaTags ev = withRideAndBooking ev $ \ride booking -> do
         Yudhishthira.RideEnd
         (Just booking.transactionId)
         (Y.EndRideTagData ride booking isDriverSameAsCustomer shouldBlockCoinsForSameRiderFlow rideDurationSeconds)
+  QRide.updateRideTags (ride.rideTags <> eitherToMaybe nammaTags) ride.id
 
 ------------------------------------------------------------
 -- P1b-4 : Fleet + Operator analytics
--- a/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
+++ b/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
@@ -9,7 +9,6 @@
 module Processor.RideEvents.Handlers
   ( handleAnalyticsKafka,
     handleRideInterpolation,
-    handleNammaTags,
     handleFleetOperatorStats,
     handleGpsTollBehavior,
     handleRCStatsReminders,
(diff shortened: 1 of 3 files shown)


Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
```diff
@@ -208,15 +208,16 @@ handleNammaTags ev = withRideAndBooking ev $ \ride booking -> do
...
+        maybe 0 (\tStart -> max 0 $ roundToIntegral (diffUTCTime rideEndTime tStart)) ride.tripStartTime
+  nammaTags <-
     withTryCatch "ride-events:computeNammaTags" $
       LYDL.computeNammaTagsWithDebugLog
         LYDL.Driver
         (cast booking.merchantOperatingCityId)
         Yudhishthira.RideEnd
         (Just booking.transactionId)
         (Y.EndRideTagData ride booking isDriverSameAsCustomer shouldBlockCoinsForSameRiderFlow rideDurationSeconds)
+  QRide.updateRideTags (ride.rideTags <> eitherToMaybe nammaTags) ride.id
```
sidsethu: **Blocker: `eitherToMaybe` is not in scope — this won't compile.**

Checked at `40e5d96`:

- `Kernel.Prelude` doesn't re-export it (it pulls in `Data.Maybe`, `Safe`, and only `Universum.Debug` / `Universum.Print` / `Universum.String.Conversion` — no `Data.Either.Extra`)
- `Kernel.Utils.Common` is imported here with an explicit list (lines 45–54) that doesn't include it
- No other import in this module provides it

The equivalent driver-app code needs the explicit import — `Domain/Action/UI/Ride/CancelRide/Internal.hs:31` has `import Data.Either.Extra (eitherToMaybe)` for exactly this `ride.rideTags <> eitherToMaybe nammaTags` expression.

Note that adding that import isn't sufficient on its own: `extra` is not in `Backend/app/kafka-consumers/package.yaml` `dependencies:`, so it would need to be added there plus a `, hpack`. Two alternatives that avoid the new dependency:

- `either (const Nothing) Just nammaTags`
- `hush nammaTags` — `errors` is already a dependency, and provides it via `Control.Error.Util`

Worth flagging that CI won't catch this: the four checks on this PR are HLint and check-conditions only, and HLint parses without scope-checking. Please run `cabal build kafka-consumers` locally before merge.

Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
```diff
@@ -208,15 +208,16 @@ handleNammaTags ev = withRideAndBooking ev $ \ride booking -> do
         riderBlockedForCoins
           || priorRidesSameCustomer > thresholdConfig.sameRiderDriverRideCountThreshold
       rideDurationSeconds =
-        maybe 0 (\tStart -> max 0 $ roundToIntegral (diffUTCTime now tStart)) ride.tripStartTime
-  void $
+        maybe 0 (\tStart -> max 0 $ roundToIntegral (diffUTCTime rideEndTime tStart)) ride.tripStartTime
+  nammaTags <-
```
sidsethu: **This populates the column but doesn't change any ride-validity behaviour — worth stating that in the description.**

Nothing that acts on ride validity reads `ride_tags`. All three consumers call `Tools.Utils.isValidRide`, which is a pure function over an in-memory `Ride`, and they all receive the same `updRide` built in `driverEndRide` — which `EndRide.hs:612` still defines as `let updRide = updRide'`, with no tags merged in:

- `EndRide/Internal.hs:255` → `publishRideEnded ... validRide` → `ev.isValidRide` → referral payout
- `Lib/DriverCoins/Coins.hs:368,374,518` → coin events
- `Beckn/ACL/Common/Order.hs:318` → the `IS_VALID_RIDE` tag sent to the BAP

All three run in the synchronous end-ride path, before this handler ever sees the event, so an async write can't reach them. `handleReferral` in this same consumer also takes `ev.isValidRide` off the event payload rather than re-reading the ride, so even the sequential in-consumer ordering doesn't help it.

Since `isValidRide` fail-opens (`maybe True (elem validRideTag) ride.rideTags`), today every ride is treated as valid. After this PR the column will fill in correctly while that stays true — which makes the problem *harder* to spot, since the data will look healthy.

Still worth landing for dashboards / analytics / backfills. Two suggestions:

1. Retitle to something like "persist ride tags for analytics", so it isn't read as a fix to referral payouts, coins, or the rider-side signal.
2. Track the actual fix separately — restoring the tag merge into `updRide` in `EndRide.hs`. It was dropped in f4a68456e3, which replaced `let updRide = updRide' {DRide.rideTags = ride.rideTags <> eitherToMaybe newRideTags}` with `let updRide = updRide'`.

If that sync-path merge is restored, this handler should be removed rather than kept — otherwise the same rules get evaluated twice per ride and two writers race on the same column.

sidsethu (changes requested):
Two things before this can go in — details inline.

1. **Won't compile** — `eitherToMaybe` isn't in scope in this module, and `extra` isn't a dependency of `kafka-consumers`. CI here is HLint-only, so nothing type-checks this PR.

2. **Scope** — writing `ride_tags` from the consumer fixes the column, but no ride-validity consumer reads that column; they all read the in-memory `updRide`, which still has no tags. Suggest reframing the PR as an analytics/persistence fix and tracking the behavioural fix (restoring the tag merge in `EndRide.hs`, dropped in f4a68456e3) separately.

The `rideDurationSeconds` change from `now` to `rideEndTime` is a genuine correctness fix — consumer lag would otherwise inflate the duration fed to the rule engine. Worth keeping regardless of the above.

One smaller note not worth blocking on: the `merchantLocalDay` half of that same change is inert. `countPriorCompletedRidesWithSameCustomer` discards both the `Id Ride` and the `Day` arguments (`RideExtra.hs:462-464` → `getDriverRiderRideCountForDays driverId riderDetailsId _ daysToFetch`), delegating to a Redis sliding-window read. So switching `now` to `rideEndTime` there has no runtime effect, and the PR description's "determine the local end day" claim doesn't hold.

sidsethu (approved):

PR #16109, opened 2026-08-05
feat: enhance scheduling logic in search and fulfillment handling

FCJ6202:
## Description
Introduces improvements to the handling of scheduled rides in the search and fulfillment processes. Key changes include:
- Updated logic to determine if a ride is scheduled based on explicit category codes.
- Enhanced the  data structure to include a definitive  field, ensuring consistent scheduling status across estimates and quotes.
- Adjusted transformer functions to utilize the new scheduling logic, ensuring that the correct category codes are assigned based on the scheduling status.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for negotiated fare adjustments.
  - Added configuration options for scheduled ride availability and negotiation fare tolerance limits.
  - These updates enable more flexible fare negotiation and scheduled ride handling across supported services.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
@@ -89,6 +89,7 @@ buildSearchReqRaw messageId bapSubscriberId bapSubscriberUrl req context actualB
       userBackendAppVersion = Beckn.OnDemand.Utils.Search.buildUserBackendAppVersion req
       emailDomain_ = Beckn.OnDemand.Utils.Search.buildEmailDomain req
       businessEmailDomain_ = Beckn.OnDemand.Utils.Search.buildBusinessEmailDomain req
+      isExplicitlyScheduled_ = Beckn.OnDemand.Utils.Search.isScheduledCategoryCode (Beckn.OnDemand.Utils.Search.getCategoryCode req)
   userBundleVersion <- mapM Kernel.Utils.Version.readVersion (Beckn.OnDemand.Utils.Search.buildUserBundleVersion req)
   userSdkVersion <- mapM Kernel.Utils.Version.readVersion (Beckn.OnDemand.Utils.Search.buildUserSdkVersion req)
   riderPreferredOption <- case Beckn.OnDemand.Utils.Search.buildRiderPreferredOption req of
@@ -133,6 +134,7 @@ buildSearchReqRaw messageId bapSubscriberId bapSubscriberUrl req context actualB
         dropLocation = dropLocation_,
         isReallocationEnabled = isReallocationEnabled_,
         fareParametersInRateCard = fareParametersInRateCard_,
+        isExplicitlyScheduled = isExplicitlyScheduled_,
         messageId = messageId_,
         pickupAddress = pickupAddress_,
         pickupLocation = pickupLocation_,
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Common.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Common.hs
@@ -111,7 +111,12 @@ data Pricing = Pricing
     personalDiscount :: Maybe HighPrecMoney,
     qar :: Maybe Double,
     area :: Maybe Text,
-    navigationInstruction :: Maybe Text
+    navigationInstruction :: Maybe Text,
+    -- | The already-decided scheduling status for this specific estimate/quote/booking
+    -- (set once, at Domain.Action.Beckn.Search.getPossibleTripOption time) -- the single
+    -- source of truth for "is this scheduled", not to be independently re-derived from
+    -- timestamps downstream when building the on_search/on_confirm response.
+    isScheduled :: Bool
   }
 
 data RateCardBreakupItem = RateCardBreakupItem
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/OnSearch.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/OnSearch.hs
@@ -58,8 +58,10 @@ tfCatalogProviders res bppConfig isValueAddNP = do
       pricings = (map (Beckn.OnDemand.Utils.Common.convertEstimateToPricing res.specialLocationName res.specialLocationSupportNumber res.fareSettlementType) res.estimates) <> (map (Beckn.OnDemand.Utils.Common.convertQuoteToPricing res.specialLocationName res.specialLocationSupportNumber res.fareSettlementType) res.quotes)
       providerFulfillments_ = map (tfProviderFulfillments res) pricings & Just
       providerItems_ = Just $ map (tfProviderItems res isValueAddNP) pricings
-      allTripCategories = Data.List.nubBy (\a b -> BecknV2.OnDemand.Utils.Common.tripCategoryToCategoryCode a == BecknV2.OnDemand.Utils.Common.tripCategoryToCategoryCode b) $ map (.tripCategory) pricings
-      providerCategories_ = Just $ map BecknV2.OnDemand.Utils.Common.mkCategory allTripCategories
+      -- Each pricing carries its own already-decided isScheduled (from the Estimate/Quote
+      -- row itself, set once at search time) rather than re-deriving it here from res.
+      allCategoryCodes = Data.List.nub $ map (\p -> BecknV2.OnDemand.Utils.Common.scheduledTripCategoryToCategoryCode p.isScheduled p.tripCategory) pricings
+      providerCategories_ = Just $ map BecknV2.OnDemand.Utils.Common.mkCategoryFromCode allCategoryCodes
   BecknV2.OnDemand.Types.Provider {providerCategories = providerCategories_, providerDescriptor = providerDescriptor_, providerFulfillments = providerFulfillments_, providerId = providerId_, providerItems = providerItems_, providerLocations = providerLocations_, providerPayments = providerPayments_}
 
 tfItemPrice :: Beckn.OnDemand.Utils.Common.Pricing -> Maybe BecknV2.OnDemand.Types.Price
@@ -82,10 +84,10 @@ tfProviderFulfillments res pricing = do
       fulfillmentCustomer_ = Nothing
       fulfillmentId_ = Just pricing.pricingId
       fulfillmentState_ = Nothing
-      -- ONDC v2.1.0: For scheduled rides (startTime > now), include pickup time window
-      -- duration so the BAP knows this is a scheduled fulfillment with a pickup window.
-      isScheduled = res.startTime > res.now
-      mbScheduledPickupDuration = if isScheduled then Just (BUtils.formatTimeDifference res.transporterConfig.scheduleRideBufferTime) else Nothing
+      -- ONDC v2.1.0: include a pickup time window duration when this fulfillment is for
+      -- a deferred (scheduled) search, so the BAP knows to expect a pickup window rather
+      -- than an immediate assignment.
+      mbScheduledPickupDuration = if pricing.isScheduled then Just (BUtils.formatTimeDifference res.transporterConfig.scheduleRideBufferTime) else Nothing
       fulfillmentStops_ = Beckn.OnDemand.Utils.Common.mkStops res.fromLocation res.toLocation res.stops res.startTime mbScheduledPickupDuration
       fulfillmentTags_ = Beckn.OnDemand.Utils.Common.mkVehicleTags pricing.vehicleServiceTierAirConditioned pricing.isAirConditioned
       fulfillmentType_ = Just pricing.fulfillmentType
@@ -101,7 +103,7 @@ tfProviderItems res isValueAddNP pricing = do
       itemPaymentIds_ = Nothing
       itemTags_ = Beckn.OnDemand.Utils.OnSearch.mkItemTags pricing isValueAddNP res.fareParametersInRateCard
       itemPrice_ = tfItemPrice pricing
-      itemCategoryIds_ = Just [BecknV2.OnDemand.Utils.Common.tripCategoryToCategoryCode pricing.tripCategory]
+      itemCategoryIds_ = Just [BecknV2.OnDemand.Utils.Common.scheduledTripCategoryToCategoryCode pricing.isScheduled pricing.tripCategory]
   let itemAddOns_ = Beckn.OnDemand.Utils.OnSearch.mkRentalAddOns pricing
   BecknV2.OnDemand.Types.Item {itemAddOns = itemAddOns_, itemCategoryIds = itemCategoryIds_, itemCancellationTerms = Just Beckn.OnDemand.Utils.Common.mkItemCancellationTerms, itemDescriptor = itemDescriptor_, itemFulfillmentIds = itemFulfillmentIds_, itemId = itemId_, itemLocationIds = itemLocationIds_, itemPaymentIds = itemPaymentIds_, itemPrice = itemPrice_, itemTags = itemTags_}
 
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Search.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Search.hs
@@ -277,6 +277,17 @@ mapCategoryCodeToRiderPreferred = \case
   "ON_DEMAND_EASY_BOOKING" -> DRPO.EasyBooking
   _ -> DRPO.OneWay
 
+-- | True when the incoming search's own category explicitly declared itself scheduled
+-- (ONDC v2.1.0 SCHEDULED_TRIP/SCHEDULED_RENTAL), independent of whatever the pickup
+-- timestamp vs. now/buffer comparison would separately conclude. Combined with that
+-- timestamp check in getPossibleTripOption so an explicit category signal is honoured
+-- even at the edge where the timestamp alone wouldn't yet cross scheduleRideBufferTime.
+isScheduledCategoryCode :: Maybe Text -> Bool
+isScheduledCategoryCode = \case
+  Just "SCHEDULED_TRIP" -> True
+  Just "SCHEDULED_RENTAL" -> True
+  _ -> False
+
 firstStop :: [Spec.Stop] -> Maybe Spec.Stop
 firstStop = find (\stop -> Spec.stopType stop == Just (show Enums.START))
 
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Search.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Search.hs
@@ -174,6 +174,10 @@ data DSearchReq = DSearchReq
     userSdkVersion :: Maybe Version,
     userBackendAppVersion :: Maybe Text,
     riderPreferredOption :: DRPO.RiderPreferredOption,
+    -- | True when the incoming search's own category explicitly said SCHEDULED_TRIP/
+    -- SCHEDULED_RENTAL (ONDC v2.1.0), independent of the pickupTime-vs-now/buffer check.
+    -- See Beckn.OnDemand.Utils.Search.isScheduledCategoryCode.
+    isExplicitlyScheduled :: Bool,
     emailDomain :: Maybe Text,
     businessEmailDomain :: Maybe Text
   }
@@ -1013,7 +1017,8 @@ checkForIntercityOrCrossCity transporterConfig mbDropLocation mbToSpecialLocatio
 getPossibleTripOption :: UTCTime -> DTMT.TransporterConfig -> DSearchReq -> Bool -> Bool -> Maybe Text -> TripOption
 getPossibleTripOption now tConf dsReq isInterCity isCrossCity destinationTravelCityName = do
   let (schedule, isScheduled) =
-        if maybe True not dsReq.isMultimodalSearch && tConf.scheduleRideBufferTime `addUTCTime` now < dsReq.pickupTime
+        if maybe True not dsReq.isMultimodalSearch
+          && (dsReq.isExplicitlyScheduled || tConf.scheduleRideBufferTime `addUTCTime` now < dsReq.pickupTime)
           then (dsReq.pickupTime, True)
           else (now, False)
       -- Local (non-intercity) bundle, scoped to what the rider actually asked for.
--- a/Backend/app/rider-platform/rider-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
+++ b/Backend/app/rider-platform/rider-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
@@ -37,13 +37,11 @@ tfCustomer taggings riderGender = do
     then Nothing
     else Just returnData
 
-tfFulfillment :: SLS.SearchRes -> Maybe BecknV2.OnDemand.Types.Fulfillment
-tfFulfillment SLS.SearchRes {..} = do
+tfFulfillment :: Bool -> SLS.SearchRes -> Maybe BecknV2.OnDemand.Types.Fulfillment
+tfFulfillment isScheduled SLS.SearchRes {..} = do
   let fulfillmentAgent_ = Nothing
       fulfillmentId_ = Nothing
       fulfillmentState_ = Nothing
-      -- ONDC v2.1.0: For scheduled rides (startTime > now), include pickup time window duration
-      isScheduled = startTime > now
       mbScheduledPickupDuration = Beckn.OnDemand.Utils.Common.mkScheduledPickupDuration isScheduled
       fulfillmentStops_ = Beckn.OnDemand.Utils.Common.mkStops origin stops startTime mbScheduledPickupDuration
       mbDropAddress = (.address) <$> Kernel.Prelude.listToMaybe (Kernel.Prelude.reverse stops)
@@ -58,12 +56,22 @@ tfFulfillment SLS.SearchRes {..} = do
     then Nothing
     else Just returnData
 
+-- | A search is "scheduled" when its pickup is far enough in the future that driver
+-- allocation gets deferred (SharedLogic.SearchTry.initiateDriverSearchBatch) rather than
+-- run immediately. Mirrors the gate used at booking-confirm time (SharedLogic.Confirm.hs);
+-- keep both in sync rather than letting each site derive its own notion of "scheduled".
+isScheduledSearch :: SLS.SearchRes -> Bool
+isScheduledSearch SLS.SearchRes {..} =
+  maybe False not searchRequest.isMultimodalSearch
+    && merchant.scheduleRideBufferTime `Kernel.Prelude.addUTCTime` now < startTime
+
 tfIntent :: SLS.SearchRes -> BecknConfig -> Maybe BecknV2.OnDemand.Types.Intent
 tfIntent res bapConfig = do
-  let intentTags_ = Nothing
-      intentFulfillment_ = tfFulfillment res
+  let isScheduled = isScheduledSearch res
+      intentTags_ = Nothing
+      intentFulfillment_ = tfFulfillment isScheduled res
       intentPayment_ = tfPayment res bapConfig
-      intentCategory_ = riderPreferredOptionToCategory res.riderPreferredOption
+      intentCategory_ = riderPreferredOptionToCategory isScheduled res.riderPreferredOption
       returnData = BecknV2.OnDemand.Types.Intent {intentCategory = intentCategory_, intentFulfillment = intentFulfillment_, intentPayment = intentPayment_, intentTags = intentTags_}
       allNothing = BecknV2.OnDemand.Utils.Common.allNothing returnData
   if allNothing
@@ -109,11 +117,11 @@ tfPerson taggings riderGender = do
     then Nothing
     else Just returnData
 
-riderPreferredOptionToCategory :: DRPO.RiderPreferredOption -> Maybe BecknV2.OnDemand.Types.Category
-riderPreferredOptionToCategory rpo = do
+riderPreferredOptionToCategory :: Bool -> DRPO.RiderPreferredOption -> Maybe BecknV2.OnDemand.Types.Category
+riderPreferredOptionToCategory isScheduled rpo = do
   let code = case rpo of
-        DRPO.OneWay -> "ON_DEMAND_TRIP"
-        DRPO.Rental -> "ON_DEMAND_RENTAL"
+        DRPO.OneWay -> if isScheduled then "SCHEDULED_TRIP" else "ON_DEMAND_TRIP"
+        DRPO.Rental -> if isScheduled then "SCHEDULED_RENTAL" else "ON_DEMAND_RENTAL"
         DRPO.InterCity -> "INTERCITY_TRIP"
         DRPO.Ambulance -> "ON_DEMAND_TRIP"
         DRPO.Delivery -> "ON_DEMAND_TRIP"
--- a/Backend/lib/beckn-spec/src/BecknV2/OnDemand/Utils/Common.hs
+++ b/Backend/lib/beckn-spec/src/BecknV2/OnDemand/Utils/Common.hs
@@ -225,6 +225,30 @@ mkCategory tc =
       categoryId = Just (tripCategoryToCategoryCode tc)
     }
 
+-- | Same as tripCategoryToCategoryCode, but upgrades ON_DEMAND_TRIP/ON_DEMAND_RENTAL to
+-- their ONDC v2.1.0 SCHEDULED_TRIP/SCHEDULED_RENTAL counterparts when the search itself
+-- was scheduled far enough ahead that driver allocation gets deferred rather than run now.
+scheduledTripCategoryToCategoryCode :: Bool -> TripCategory -> Text
+scheduledTripCategoryToCategoryCode isScheduled tc =
+  case (isScheduled, tripCategoryToCategoryCode tc) of
+    (True, "ON_DEMAND_TRIP") -> "SCHEDULED_TRIP" -- Added for Nammayatri due to backward compatibility
+    (True, "ON_DEMAND_RENTAL") -> "SCHEDULED_RENTAL" -- Added for Nammayatri due to backward compatibility
+    (_, code) -> code
+
+mkCategoryFromCode :: Text -> Spec.Category
+mkCategoryFromCode code =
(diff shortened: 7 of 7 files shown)


FCJ6202:
@ClaudeWeb, can you review this PR and let me know if you find any backward compatibility issues?

Our goal is to support ONDC 2.1.0 changes for MSIL merchants through a separate layer. When the feature flag is ON, the new ONDC 2.1.0 flow should be used. When the flag is OFF, the system should work exactly as it did before.

Please specifically check whether my changes can affect any existing flows when the flag is OFF, and let me know if you find any regressions or backward compatibility issues.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/BecknConfig.yaml
```diff
@@ -22,6 +22,7 @@ BecknConfig:
     vehicleCategory: VehicleCategory|NoRelation
     collectedBy: PaymentCollectedBy
     staticTermsUrl: Maybe BaseUrl
+    offlineContract: Maybe Bool
```
RitikaHanish: I feel, it can be BPP and BAP combo level, but after putting in this table, currently it's only at BPP level.
@FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1441,6 +1415,9 @@ TransporterConfig:
     forceDirectCalling: Maybe Bool
     allowNonFleetDriverLiveMap: Maybe Bool
     allowInvalidRcDeletionOnReplacement: Maybe Bool # When True, allows deleting an invalid RC and its driver/fleet associations when a replacement RC is submitted
+    enableScheduledCategorySignal: Maybe Bool # City-level gate for the MSIL-style Quote-based (static/scheduled) /select capability
```
RitikaHanish: Can be named better.
@FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1441,6 +1415,9 @@ TransporterConfig:
...
+    negotiationFareMinTolerancePct: Maybe Double # Lower bound tolerance (as a fraction) around the estimated fare for accepting a negotiated fare in /select
```
RitikaHanish, line 1478: This should be Int.
@FCJ6202 
like 10 means - 10%

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1441,6 +1415,9 @@ TransporterConfig:
...
+    negotiationFareMaxTolerancePct: Maybe Double # Upper bound tolerance (as a fraction) around the estimated fare for accepting a negotiated fare in /select
```
RitikaHanish, line 1479: Same as min 
@FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
RitikaHanish, line 1478: Also we can move this to fare policy, it will be more flexible, variant level, 
peak timings and other variables rather than merchant level.
@FCJ6202

Backend/lib/beckn-spec/src/BecknV2/OnDemand/Types.hs
```diff
@@ -2064,6 +2155,48 @@ optionsStop =
...
       ]
 
+-- |
+data Support = Support
+  { -- |
+    supportDescriptor :: Maybe Descriptor,
+    -- |
+    supportRefType :: Maybe Text,
+    -- |
+    supportRefId :: Maybe Text,
```
RitikaHanish, line 2165: as I remember even supportRefIdor or supportRefType, one of it is enum only.

Backend/dev/migrations-read-only/dynamic-offer-driver-app/beckn_config.sql
```diff
@@ -61,5 +61,9 @@ ALTER TABLE atlas_driver_offer_bpp.beckn_config ADD COLUMN multimodal_on_search_
 
 ------- SQL updates -------
 
-ALTER TABLE atlas_driver_offer_bpp.beckn_config ADD COLUMN network_id text ;
-ALTER TABLE atlas_driver_offer_bpp.beckn_config ADD COLUMN beckn_protocol text  default 'Beckn_V2';
\ No newline at end of file
+ALTER TABLE atlas_driver_offer_bpp.beckn_config ADD COLUMN offline_contract boolean ;
```
RitikaHanish: Shoudln't this also be moved to bapmetadata ?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Confirm.hs
```diff
@@ -84,6 +87,16 @@ confirm transporterId (SignatureAuthResult _ subscriber) reqV2 = withFlowHandler
     Redis.whenWithLockRedis (SRide.confirmLockKey dConfirmReq.bookingId) 60 $ do
       now <- getCurrentTime
       (transporter, eitherQuote) <- DConfirm.validateRequest subscriber transporterId dConfirmReq now
+      -- Verifying: store the BAP's declared BAP_TERMS.STATIC_TERMS (if any)
```
RitikaHanish, line 93: Make these comments simpler.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Confirm.hs
```diff
@@ -125,7 +138,23 @@ confirm transporterId (SignatureAuthResult _ subscriber) reqV2 = withFlowHandler
...
-      let onConfirmMessage = ACL.buildOnConfirmMessageV2 dConfirmRes pricing becknConfig mbFarePolicy bppInvoiceInfo
+      let onConfirmMessage' = ACL.buildOnConfirmMessageV2 dConfirmRes pricing becknConfig mbFarePolicy bppInvoiceInfo
+      -- Pilot: cities with enableScheduledCategorySignal get both the NEW ->
+      -- RIDE_CONFIRMED fulfillment-state fix and the BAP_TERMS/BPP_TERMS order-tag
+      -- patch (echoing back the BAP's own declared STATIC_TERMS, if known, plus
+      -- ours from becknConfig), applied in one pass by
+      -- Beckn.OnDemand.Transformer.MSIL.OnConfirm.msilOnConfirmMessageBuild --
+      -- unlike on_search/on_select/on_init, which only get BPP_TERMS. Everyone
+      -- else's onConfirmMessage is exactly what Layer 1 built, unchanged.
+      transporterConfig <- getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = dConfirmRes.booking.merchantOperatingCityId.getId}) Nothing >>= fromMaybeM (TransporterConfigDoesNotExist dConfirmRes.booking.merchantOperatingCityId.getId)
```
RitikaHanish, line 151: already fetched above.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Confirm.hs
```diff
@@ -125,7 +138,23 @@ confirm transporterId (SignatureAuthResult _ subscriber) reqV2 = withFlowHandler
       vehicleServiceTierItem <- CQVST.findByServiceTierTypeAndCityIdInRideFlow dConfirmRes.booking.vehicleServiceTier dConfirmRes.booking.merchantOperatingCityId (dConfirmRes.booking.area >>= SL.pickupSpecialZoneIdFromArea) >>= fromMaybeM (VehicleServiceTierNotFound (show dConfirmRes.booking.vehicleServiceTier))
       let pricing = Utils.convertBookingToPricing vehicleServiceTierItem dConfirmRes.booking
       bppInvoiceInfo <- ACL.resolveBPPInvoiceInfo dConfirmRes
-      let onConfirmMessage = ACL.buildOnConfirmMessageV2 dConfirmRes pricing becknConfig mbFarePolicy bppInvoiceInfo
+      let onConfirmMessage' = ACL.buildOnConfirmMessageV2 dConfirmRes pricing becknConfig mbFarePolicy bppInvoiceInfo
```
RitikaHanish: @FCJ6202 Are you sure no parsing will fail, if we always call the previous handler first

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Init.hs
```diff
@@ -78,20 +81,41 @@ init transporterId (SignatureAuthResult _ subscriber) reqV2 = withFlowHandlerBec
...
+      MSILTerms.verifyIncomingStaticTerms (Id bapId) Domain.MOBILITY incomingOrderTags
+
+    -- Verifying: Layer 1's echo-based fulfillment-id parse (QuoteId vs
+    -- DriverQuoteId) is unreliable for MSIL, since our own fulfillment.type
+    -- override on /on_select collapses the signal it depends on -- see
+    -- Beckn.OnDemand.Utils.MSIL.FulfillmentType.correctFulfillmentId.
+    dInitReq' <-
+      if isMsilPilotMerchant
+        then do
+          correctedFulfillmentId <- MSILInit.correctFulfillmentId transactionId dInitReq.fulfillmentId
```
RitikaHanish, line 104: It shouldn't be needed, Please explain why this was there.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/BapMetadata.yaml
```diff
@@ -17,6 +17,10 @@ BapMetadata:
     name : Text
     logoUrl : Maybe BaseUrl
     domain: Maybe Text # TODO: Use Domain type here
+    staticTermsUrl : Maybe BaseUrl
```
XyneSpaces, line 20: **[blocking] Repository DB/C2** — This adds persisted `bap_metadata` columns, but I don't see a matching executable migration in the PR. The generated Beam type/query now reads these columns, so a deploy before the DB has them can break existing BAP metadata reads even when `enableScheduledCategorySignal` is off. Add a safe `ALTER TABLE ... ADD COLUMN ... NULL` migration for each new column before the generated code is deployed.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/BecknConfig.yaml
XyneSpaces: **[blocking] Repository DB/C2** — `offlineContract`/`onSupportTTLSec` are new persisted config fields, but the PR only updates the storage spec/generated code. Existing `beckn_config` lookups will select the new columns regardless of the MSIL feature flag, so flag-off flows can fail against a DB without these columns. Ship the DB migration with nullable/defaulted additions.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1415,6 +1415,9 @@ TransporterConfig:
     forceDirectCalling: Maybe Bool
     allowNonFleetDriverLiveMap: Maybe Bool
     allowInvalidRcDeletionOnReplacement: Maybe Bool # When True, allows deleting an invalid RC and its driver/fleet associations when a replacement RC is submitted
+    enableScheduledCategorySignal: Maybe Bool # City-level gate for the MSIL-style Quote-based (static/scheduled) /select capability
```
XyneSpaces: **[blocking] Repository DB/C2** — The new `TransporterConfig` columns are wired into generated Beam mappings, but there is no executable migration in this PR. Since `TransporterConfig` is read in normal request paths before/around the feature-flag branch, existing flag-off merchants can still hit missing-column failures after code deploy. Add the corresponding safe `ALTER TABLE transporter_config ADD COLUMN ...` migration with defaults/nullability.

XyneSpaces (commented):
## NammaYatri Reviewer — nammayatri/nammayatri PR #16109

**Head SHA:** `4fc1675b413d068b18de375a62d349861c7f3ba3`
**Review rules:** repository-specific rules loaded from `NammaYatri Review — Consolidated`; local `CODEOWNERS`, `AGENTS.md`, and `.github/review-rules.md` were not found.
**Reviewed:** 29 files, all available changed hunks · **Findings:** 3 blocking, 0 major, 0 minor, 0 nits

### Findings

| blocking | Repository DB/C2 | `Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/BapMetadata.yaml:20` | Adds persisted `bap_metadata` fields without an executable DB migration; generated Beam readers will expect the new columns even when the MSIL flag is off. |
| blocking | Repository DB/C2 | `Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/BecknConfig.yaml:25` | Adds persisted `beckn_config` fields without an executable DB migration; existing config lookups can fail before any feature-flag branch. |
| blocking | Repository DB/C2 | `Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml:1418` | Adds persisted `transporter_config` fields without an executable DB migration; existing transporter-config reads can fail for flag-off merchants. |

### Recommended reviewer

**@prakharritik** is already requested for review — lowest pending load among the currently requested reviewers and relevant to backend Beckn/search-flow review.

| @prakharritik | `API/Beckn/*`, `Beckn/OnDemand/Transformer/MSIL/*` | current requested reviewer | 5 open PRs involving them | lowest load among requested reviewers |
| @RitikaHanish | backend/fare/config paths | current requested reviewer | 12 open PRs involving them | fare/backend context |
| @hkmangla | backend architecture | current requested reviewer | 42 open PRs involving them | broad backend approval context |
| @RitoBose | scheduled-ride config/search domain | recent related PRs #16420, #16450, #16322 | 7 open PRs involving them | strongest scheduling-domain context, but not currently requested |

Workload scope: target-repository workload (`involves:<user> repo:nammayatri/nammayatri type:pr state:open`).

I did not request a new reviewer because @prakharritik is already requested on GitHub.

### 🛢️ Migration Safety — unsafe operations found

cc @qa-nammayatri

| Migration | Tier | Statement | What it breaks | Fix |
|-----------|------|-----------|----------------|-----|
| missing migration for `bap_metadata`, `beckn_config`, `transporter_config` | 🔴 DANGEROUS | storage specs/generated Beam mappings add new columns, but PR contains no executable migration file | rolling deploy / flag-off traffic can hit generated selects/updates that reference columns absent in DB | add ordered `ALTER TABLE ... ADD COLUMN ... NULL/DEFAULT` migrations for all new persisted fields before deploying generated code |

QA deploy stance: verify the repository’s affected critical flows, backward compatibility, and old-version behavior after deployment.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Init.hs
RitikaHanish, line 104: FulfillmentID as we discussed if it's needed here.
@FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Select.hs
```diff
@@ -50,19 +56,62 @@ select ::
...
+        -- Non-pilot merchants never run this parser and never pay for the Quote
+        -- lookup below -- both only happen inside this branch.
+        let dSelectReq = MSILSelect.msilParser reqV2.selectReqMessage dSelectReq'
+            -- The wire item.id may resolve to a Quote instead of an Estimate for
+            -- the new static/scheduled /select capability (doc 25 s3). Pilot
+            -- merchants aren't restricted to Quote-based select only: an
+            -- item.id that resolves to an Estimate still falls through to the
+            -- same estimate flow non-pilot merchants use.
+            itemIdText = case dSelectReq.estimateIds of
+              (eid : _) -> eid.getId
```
0utkarsh: Why not throw error? It will fail anyway

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs
```diff
@@ -178,3 +201,61 @@ addNammaTags tagData sReq = do
...
+-- Validate the negotiated fare and update fare policy and quotes according to that
+applyNegotiatedFare :: Id DMOC.MerchantOperatingCity -> Id DQuote.Quote -> DQuote.Quote -> HighPrecMoney -> Flow DQuote.Quote
+applyNegotiatedFare merchantOpCityId quoteId quote negotiatedFare = do
+  transporterConfig <- getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = merchantOpCityId.getId}) Nothing >>= fromMaybeM (TransporterConfigDoesNotExist merchantOpCityId.getId)
+  let negotiationFareMinTolerancePct = fromMaybe 0.1 transporterConfig.negotiationFareMinTolerancePct
+      negotiationFareMaxTolerancePct = fromMaybe 0.1 transporterConfig.negotiationFareMaxTolerancePct
+      currentFare = quote.estimatedFare
+      minAcceptable = currentFare * (1 - realToFrac negotiationFareMinTolerancePct)
+      maxAcceptable = currentFare * (1 + realToFrac negotiationFareMaxTolerancePct)
+  unless (negotiatedFare >= minAcceptable && negotiatedFare <= maxAcceptable) $ -- We can discuss on this comparisation logic
```
0utkarsh, line 250: Multiple selects are allowed? In that case it will keep applying the tolerance on previously selected value, not the first time selected one

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1452,7 +1415,9 @@ TransporterConfig:
     forceDirectCalling: Maybe Bool
     allowNonFleetDriverLiveMap: Maybe Bool
     allowInvalidRcDeletionOnReplacement: Maybe Bool # When True, allows deleting an invalid RC and its driver/fleet associations when a replacement RC is submitted
-    metricsDistanceBucketsKm: "Maybe [Int]" # BPP funnel distance buckets: upper edges in KILOMETRES, ascending, max 4, e.g. {5,12,30}. Nothing = defaults (5/12/30 km, matching ClickHouse trip_distance_bin). See SharedLogic.MetricsLabels.distanceBucketEdges for how a config is validated and when it is rejected.
+    enableScheduledCategorySignal: Maybe Bool # City-level gate for the MSIL-style Quote-based (static/scheduled) /select capability
```
0utkarsh: I was thinking to keep it at BAP level but fine for now

0utkarsh (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs
FCJ6202: Didn't get this

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Select.hs
```diff
@@ -50,19 +56,57 @@ select ::
...
-  L.setOptionLocal TxnIdKey transactionId
   Utils.withTransactionIdLogTag transactionId $ do
     logTagInfo "SelectV2 API Flow" "Reached"
-    dSelectReq <- ACL.buildSelectReqV2 subscriber reqV2
+    dSelectReq' <- ACL.buildSelectReqV2 subscriber reqV2
+    merchant <- QMerch.findById transporterId >>= fromMaybeM (MerchantNotFound transporterId.getId)
+    city <- Utils.getContextCity reqV2.selectReqContext
+    moc <- CQMOC.findByMerchantIdAndCity transporterId city >>= fromMaybeM (InvalidRequest $ "Operating City " <> show city <> " not supported or not found")
+    transporterConfig <- getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = moc.id.getId}) Nothing >>= fromMaybeM (TransporterConfigDoesNotExist moc.id.getId)
+    let isPilotMerchant = fromMaybe False transporterConfig.enableScheduledCategorySignal
```
RitikaHanish: Let's use proper name @FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Select.hs
```diff
@@ -50,19 +56,57 @@ select ::
...
 
-    Redis.whenWithLockRedis (selectLockKey dSelectReq.messageId) 60 $ do
-      (merchant, searchRequest, estimates) <- DSelect.validateRequest transporterId dSelectReq
-      fork "select request processing" $ do
-        Redis.whenWithLockRedis (selectProcessingLockKey dSelectReq.messageId) 60 $
-          DSelect.handler merchant dSelectReq searchRequest estimates
-      fork "select received pushing ondc logs" do
-        void $ pushLogs "select" (toJSON reqV2) merchant.id.getId "MOBILITY"
+    if isPilotMerchant
```
RitikaHanish: Let's use proper name @FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/ACL/OnSupport.hs
```diff
@@ -0,0 +1,77 @@
...
+    HasFlowEnv m r '["nwAddress" ::: BaseUrl]
+  ) =>
+  DM.Merchant ->
+  DRB.Booking ->
+  Maybe Text ->
+  Maybe Spec.Descriptor ->
+  m Spec.OnSupportReq
+buildOnSupportMessageV2 merchant booking mbMessageId mbDescriptor = do
+  msgId <- generateGUID
+  let bppId = getShortId merchant.subscriberId
```
RitikaHanish, line 46: bppId is subscriberId not shortID @FCJ6202

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
```diff
@@ -148,11 +148,10 @@ buildSearchReqRaw messageId bapSubscriberId bapSubscriberUrl req context actualB
         toSpecialLocationId = Id <$> toSpecialLocationId_,
         emailDomain = emailDomain_,
         businessEmailDomain = businessEmailDomain_,
+        isExplicitlyScheduled = Nothing, -- only the MSIL layer (Beckn.OnDemand.Transformer.MSIL.Search.msilParser) sets this
```
RitikaHanish: Can we give more meaningful name?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs
```diff
@@ -178,3 +201,61 @@ addNammaTags tagData sReq = do
...
+    throwError $ NegotiatedFareNotAcceptable quoteId.getId
+  let negotiationDelta = negotiatedFare - currentFare
+      updatedFareParams = quote.fareParams {DFareParams.baseFare = quote.fareParams.baseFare + negotiationDelta} -- Need to discuss that should we store negotiationDelta information in other field at fare params to keep this information how much we negotiate with bap for this transaction?
```
RitikaHanish: yes it needs to be separate field, you can't club it in base fare

RitikaHanish (changes requested):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Init.hs
FCJ6202: Already explained in-person. Because they aren't passing the fulfillment type from init, tripCategory always becomes a dynamic offer, and as a result, our fulfillmentId becomes DriverQuoteId instead of QuoteId. That's why we need to correct this fulfillment ID in the case of schedule ride

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Confirm.hs
FCJ6202: Yes, because I already tested the flow with Workbench. It should pass with every BAP request as long as they are following ONDC rules

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
FCJ6202: What if they want to set a decimal percentage?

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/ACL/OnSupport.hs
FCJ6202: > let bppId = getShortId merchant.subscriberId

I am calling merchant.subscriberId only. 
Screenshot 2026-09-01 at 3 04 42 PM
You can check the type of the subscriberId. So, to get the ID, we need to call the shortId function, which will give me the raw bppId

Backend/lib/beckn-spec/src/BecknV2/OnDemand/Types.hs
FCJ6202: No. It is text. Only decripterCode is an ENUM. 

image

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
FCJ6202: We can only allowed int value. Need to add validation also which should be 0 to 100

FCJ6202: We already did this changes in other pr

Backend/lib/beckn-spec/src/BecknV2/OnDemand/Types.hs
FCJ6202: Resolved

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/Search.hs
FCJ6202: We can change it to isSchedule. Or we can send the descriptor code only here

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs
FCJ6202: We will check where we should store this negotiateDelta value in the fare params table. or we can add new column only for this value to store in fare params

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Common.hs
```diff
@@ -565,9 +567,38 @@ mkStopsOUS booking ride rideOtp mEndOtp =
...
+-- | ONDC v2.1.0 fulfillments.agent.person.creds. MEMBERSHIP_TIER is only sent for
+-- cities piloting the scheduled-category signal (see TransporterConfig.enableOndcScheduledRideSupport).
+-- ACHIEVEMENTS/BADGES are intentionally omitted -- no domain source for them yet.
+mkAgentCreds :: UTCTime -> Bool -> SP.Person -> Maybe DDriverStats.DriverStats -> [Spec.Cred]
+mkAgentCreds now isOndcScheduledRideSupportEnabled driver mbDriverStats =
+  catMaybes
+    [ if isOndcScheduledRideSupportEnabled
+        then Just Spec.Cred {Spec.credType = Just "MEMBERSHIP_TIER", Spec.credId = Just "Verified"}
+        else Nothing,
+      Just Spec.Cred {Spec.credType = Just "ACTIVE_SINCE", Spec.credId = Just (activeSinceDuration now driver.createdAt)},
```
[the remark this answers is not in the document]
FCJ6202: But it is fine beacause It isn't break the existing system. We just send extra information to BAP, which will be backward compatible with the ondc oldest version

RitikaHanish (approved):
Create Refactor PR for this.

0utkarsh:
## Review — request changes

Stack-reviewed together with #16580 and #16595 (findings placed on the PR that introduces them). Good movement since the last review round: prod migrations are now present, `negotiatedFareDelta` replaced the `baseFare` mutation, and the fetch-gate-apply pattern got centralized in the follow-up PR. The findings below survive at the stack head (`46329d7`).

### Blockers

**1. Negotiation: anchor walks on repeated selects** — `Domain/Action/Beckn/Select.hs` (`applyNegotiatedFare`)
Validation reads `currentFare = quote.estimatedFare`, and acceptance runs `QQuote.updateEstimatedFare`. A re-select (new `message_id` — legal, and the Redis lock is message-id-scoped so it doesn't dedupe) re-anchors the ±tolerance around the *previously accepted bid*: ₹500 → bid ₹450 (accepted) → bid ₹405 (within 10% of 450, accepted) → … the fare ratchets arbitrarily far from the original. Two concurrent selects with different message ids also race the write (last-write-wins).
**Fix**: never overwrite `estimatedFare` — keep the original as the immutable anchor, store the accepted bid/delta separately, and read `estimatedFare + delta` at consumers. (Also note `negotiatedFareDelta` currently records only the *last* step, not the cumulative delta.)

**2. Negotiation: breakup ≠ price now, and the negotiated fare is silently dropped at ride end** — `Beckn/OnDemand/Transformer/MSIL/OnSelect.hs` (`mkQuoteBreakupFromQuote`)
The breakup is built from `mkFareParamsBreakups(quote.fareParams)`, which has no line for `negotiatedFareDelta` — so on_select/on_init/on_confirm carry price ₹450 with breakup items summing to ₹500 (ONDC Workbench NACKs breakup-total ≠ price). Worse, the end-of-ride fare recomputation rebuilds fareParams with `negotiatedFareDelta = Nothing`, so the final on_update reverts to the **un-negotiated** amount — the rider bids ₹450, sees on_confirm at ₹450, and is billed ₹500.
**Fix**: emit the delta as its own breakup line and thread `negotiatedFareDelta` through the ride-end fare-params rebuild.

**3. Quote-stripping for external BAPs removed unconditionally** — `SharedLogic/SearchRequestProcessing.hs:59`
`dSearchRes = dSearchResWithQuotes` deletes the `isValueAddNP` filter with no pilot gate. Every existing non-value-add BAP, in every city, now receives static Quote items in on_search (special-zone/rental/intercity/scheduled searches) it was deliberately never shown — and can proceed through select/init on a Quote id as far as booking creation before the non-pilot confirm gate NACKs, leaving orphan bookings. Live-network behavior change for existing integrations.
**Fix**: gate the un-stripping on `enableOndcScheduledRideSupport`.

**4. "Payment Params not found" check dropped for all BAPs** — `Beckn/OnDemand/Utils/Init.hs:97`
The comment says "for schedule case," but the relaxation is unconditional: any BAP anywhere sending init without `payment.params` now silently books as **Cash** (empty params → `castPaymentInstrument` fallback) instead of being rejected.
**Fix**: scope the relaxation to the pilot/scheduled path.

**5. Confirm-gate bypass is broader than scheduled** — `Domain/Action/Beckn/Confirm.hs` (`validateRequest`)
The city flag short-circuits the *entire* non-value-add trip-category guard — an external BAP in a pilot city can now confirm Rental/InterCity/Delivery bookings that were previously NACKed, into flows never validated for external BAPs.
**Fix**: bypass only for the static/scheduled categories this was written for.

**6. City-level gate changes the dialect for every BAP in the city** — `SharedLogic/CallBAP.hs` (arrival reroute + `applyMsilRideAssignedOrderOverrides`)
In a pilot city the overrides apply to *all* BAPs, including the NY rider app: driver-arrival moves to on_status (which the rider app doesn't consume → riders stop seeing "driver arrived"), and breakup lines with titles outside the remap table are **deleted**, not passed through. Enabling the flag anywhere NY operates breaks arrival UX and fare display.
**Fix**: gate per-BAP (bap_metadata / whitelist) or at minimum make the remap pass-through-by-default; until then the pilot city must be whitelist-locked to a single BAP, and that guarantee should be stated in this PR's description.

**7. Safety alerts can be silently dropped in pilot cities** — `Beckn/ACL/OnSupport.hs:51`
`buildOnSupportMessageV2` hard-fails on `becknConfig.onSupportTTLSec`; the migration adds the column with no default/backfill; pilot-city safety alerts (route deviation, stoppage) route through this builder inside a fork — NULL TTL → InternalError swallowed → alert never delivered.
**Fix**: default the TTL in code or backfill in the migration.

### Majors

**8. Dynamic on_select still misses the fulfillment-id override** — `SharedLogic/CallBAP.hs` (`sendDriverOffer`): `fixMsg` composes only ItemCompliance + BreakupTitles; the comment says the fulfillment-id pin (driverQuote.id → estimate id) is included, but it isn't — pilot-city instant rides fail Workbench's id-consistency check exactly as the comment predicts.

**9. Hardcoded `Enums.CAB` in the on_search BPP_TERMS lookup** — `API/Beckn/Search.hs` (carried into the stack head's `OndcScheduledRide/OnSearch.hs`): a pilot city with no CAB beckn_config row throws "Beckn Config not found" inside the search fork — on_search is never sent for *any* search in that city. Resolve per vehicle category like every other flow.

### Minors
- No booked-quote guard in `validateQuoteSelect` (a select replayed after confirm re-negotiates a consumed quote and emits an on_select contradicting the order) — one guard query.
- `/select` pilot branch does merchant/moc/config/quote lookups before the message-id dedup lock; `/search` fetches transporter config before the `setNxExpire` dedup — duplicates pay full cost. Move inside.
- `"SCHEDULED_TRIP"` compared as a raw string in the parser while the emit side already uses `show Enums.SCHEDULED_TRIP` — use the existing enum (and add `SCHEDULED_RENTAL` to it).
- The pilot-gate fetch (getOneConfig + fromMaybe False) is copy-pasted at ~9 sites — extract one `isOndcScheduledRideSupportEnabled` predicate (also makes the per-BAP gate of finding 6 a one-place change).
- `mkFulfillmentV2` / `mkFulfillmentV2SoftUpdate` duplicate the config-fetch + personCreds block and put a `TransporterConfigDoesNotExist` throw on a hot path that previously needed no config — pass the flag/creds in from callers that already hold the config.

RitikaHanish (changes requested):
Rebase this.

PR #16134, opened 2026-08-06
Backend: feat - Whatsapp ride booking

sathikabegum-nammayatri:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added WhatsApp support for ride booking, fare quotes, saved locations, tracking, cancellation, SOS, driver contact, and safety actions.
  * Added regular and flexible ride flows with pickup and destination confirmation.
  * Added multilingual conversations in English, Hindi, Gujarati, Kannada, Tamil, and Telugu.
  * Added secure webhook integration and configurable merchant bot settings.
  * Added ride-status notifications for driver assignment, arrival, start, completion, and cancellation.
* **Bug Fixes**
  * Improved handling of duplicate, invalid, unsupported, or unconfigured messages.


Backend/dhall-configs/dev/rider-app.dhall
```diff
@@ -545,4 +545,10 @@ in  { esqDBCfg
     , bapHostRedirectMap = [] : List { mapKey : Text, mapValue : Optional Text }
     , xyneWebhookSigningSecret = "<XYNE_WEBHOOK_SIGNING_SECRET>"
     , xyneWebhookBearerToken = "<XYNE_WEBHOOK_BEARER_TOKEN>"
+    , metaBotEnabled = True
+    , metaAllowedPhones = [ "9361176218", "9343922922", "6374989436" ]
+    , metaTrackerEnabled = True
```
kavya-shree-s: Added these in merchant service table

kavya-shree-s:
## PR Review: #16134 — Backend: feat - Whatsapp ride booking

**Files**: 97 | **+13,889 −0** | **Commits**: 1

**CI Status: PASS** — but see [C6]; rider-app itself was never compiled.

---

### CRITICAL Issues

**[C4] `search'` argument order bug — the new param landed in the wrong slot at the HTTP call site**

- File: `Backend/app/rider-platform/rider-app/Main/src/API/UI/Search.hs:233`

`search'` binds `… mbEnableSyncSearch mbIsWhatsappRequest mbHasPasses` (line 239), but the HTTP handler calls it with `… mbEnableSyncSearch mbHasPasses Nothing`. All three are `Maybe Bool`, so this compiles silently.

Two consequences:

1. Every organic app search writes `search_request.is_whatsapp_request = <client's hasPasses>`, which then propagates to `booking.is_whatsapp_request` via `SharedLogic/Confirm.hs:461`. The new analytics column is wrong from day one.
2. **Regression to existing functionality**: `mbHasPasses` is always `Nothing` inside `search'`, so line 270 `multiModalSearch … mbHasPasses` → `DMC.postMultimodalInitiateSimpl … mbHasPasses` never sees the client's `hasPasses=true`. `hasPasses` is a real client query param (`API/UI/Search.hs:182,199`) and pass eligibility shipped recently (031d803b0, 205158b5e).

Fix:

```haskell
-- line 233
… mbEnableSyncSearch Nothing mbHasPasses
```

`WhatsappBot/Adapter/Backend.hs:runSearch` is correct — the bug is only at line 233.

**QA should test**: app search with `hasPasses=true` still returns pass-eligible multimodal options; `search_request.is_whatsapp_request` is null for app searches and true only for WhatsApp-originated ones.

---

**[C6] rider-app was never built by CI on this PR**

Passing checks are HLint ×2, the **Dashboard** repl check, the whatsapp-bot golden suite, check-conditions, and CodeRabbit. `.github/workflows/cabal-pr-repl-check-rider.yaml` is `workflow_dispatch:`-only and did not run.

This PR adds ~1,300 lines under `rider-app/Main/src/WhatsappBot/`, rewires `API.hs` / `App.hs` / `Environment.hs`, changes a 14-argument signature across 6 call sites, and depends on `Kernel.External.Meta.*` from shared-kernel with `flake.lock` unchanged. (Those modules do exist at the pinned rev `10df93c8` — `Api/Config/Error/Flow/Types/Webhook.hs` — so it should resolve. But nothing compiled it, and the project is `-Werror`.)

Please dispatch the rider repl check before merge.

---

C1 (API backward compat), C2 (no destructive DDL), C3 (no Redis pattern matching) and C5 (PII encryption) are all clean. `appSecret` / `verifyToken` / `accessToken` are correctly `EncryptedField 'AsEncrypted Text`, and `"!SecondaryKey"` is properly quoted per CLAUDE.md rule 11.

---

### IMPORTANT Issues

**[I2] Migration comments cite three files that do not exist, and `0049` has no backfill**

`dev/feature-migrations/0049-…sql` references `0047-meta-config-access-token-backfill.sql` (0047 is actually `whatsapp-bot-translations-seed.sql`) and `1562-meta-config-access-token-not-null.sql`. `src/Tools/Meta.hs:24-30` additionally cites `1561-meta-config-rename-and-access-token.sql`. None exist — ddl-migrations jumps 1558 → 1563.

Real impact: `meta_config` is created empty (no seed INSERT anywhere in the PR) while `0049` deletes the `Meta_CloudApi` row that currently holds the outbound credentials. Someone must hand-insert a row whose three secret columns are Passetto ciphertext before anything works. Either fix the comments and document the runbook, or hold `0049` until provisioning is done.

**[I3] Unauthenticated GET handshake does a full table scan plus one Passetto `decrypt` per row**

- File: `src/Domain/Action/UI/MetaWhatsappWebhook.hs:78-88`

`GET /webhook/whatsapp` is public and unrate-limited; each request runs `QMWC.findAll` then `decrypt` per row until a match. The "not a hot path" comment holds for legitimate traffic, not for someone hammering it. Suggest rate-limiting the route, or matching on a hash rather than decrypt-and-compare.

**[I4]** shared-kernel already ships `Kernel.External.Meta.Webhook.metaVerifyChallenge` — constant-time, identical semantics. `getWebhookChallenge` reimplements it inline.

**[I3] Tracker fan-out: every pod × every merchant re-reads one global ride set**

- Files: `src/WhatsappBot/Adapter/Registry.hs:19`, `Adapter/Tracker.hs:19-35`, `Adapter/Env.hs:167`

`ridesSetKey = "wab:activerides"` is global. `buildTrackerDeps` filters by `merchantLabel` only *after* `ridesFor` has done `SMEMBERS` plus one `GET` per booking id. So each tick costs `pods × merchants × totalActiveRides` Redis GETs, most discarded — every `metaTrackerPollMs` (3000 ms in dev dhall). There is no distributed lock; the code comment acknowledges this ("for fleet-wide single execution, wrap the tick in a Redis lock").

Suggestion: merchant-scoped set keys (`wab:activerides:<label>`) plus a Redis leader lock around the tick.

**[I3] `merchantLabel` uniqueness is the only multi-tenant boundary, and it is unenforced**

`Adapter/Env.hs:154-158` states this outright: a duplicate label lets one merchant's tick push another merchant's ride from the wrong WhatsApp number. The unique index is on `phone_number_id`, not on `bot_config->>'merchantLabel'`. Either add a unique expression index, or key the registry on `phoneNumberId`, which is already unique.

**[I3] Session read-modify-write races** — `Adapter/SessionStore.hs`

`resolveSession` and `saveContext` are unlocked `GET` → `SETEX`. Inbound messages run in `fork`s, so two fast messages from one user can interleave and drop a context update mid-booking.

**[I2] Five new non-optional `AppCfg` fields**

`metaBotEnabled`, `metaTrackerEnabled`, `metaTrackerPollMs`, `metaSessionTtlSec`, `metaTrackerMaxAgeSec`. Only `dhall-configs/dev/rider-app.dhall` is updated in-repo — every out-of-repo prod/staging config must add all five or rider-app fails to parse its config at boot. Also consider shipping dev with `metaBotEnabled = False` rather than `True`.

I1 is clean: `findByPhoneNumberId` is backed by `idx_meta_config_phone_number_id`; `findAll` scans a tiny config table.

---

### Good to Have

- **[G1] NammaDSL generator version drift.** `src-read-only/Storage/Beam/SearchRequest.hs` shows −76/+77 lines but the only semantic change is `+ isWhatsappRequest`; everything else is `:: B.C f X` → `:: (B.C f X)` reparenthesization. `Storage/Queries/SearchRequestPartiesLink.hs` changed `[X]` → `([X])` with no spec change at all. Your generator is newer than the one that produced the rest of the tree, so the next person running `, run-generator` will flip it back. (I diffed the field sets — `enableKVPG` secondary keys are unchanged, so no KV regression.)
- **[G3]** `Adapter/Sender.hs:44` — `take 10 btns` silently truncates buttons. The webhook logs a warning when it drops a batch tail; do the same here.
- **[G3]** `lib/whatsapp-bot/src/WhatsappBot/Env.hs:124` — `hashPhone` is unsalted SHA-256 over a 10-digit number. The haddock calls it one-way and invokes the DbHash convention, but 10^10 is a seconds-long brute force; a Redis dump yields the numbers. The salted `getDbHash` is used elsewhere in this same PR. Not a blocker — the comment just overstates the guarantee.
- **[G1]** `wab:user:` / `wab:introsent:` have no TTL by design (`Adapter/PersonStore.hs`) — permanent per-user Redis keys with no eviction path. Fine at pilot scale; needs a bound before wide rollout.
- **[G1]** `take 1 allEvents` in the webhook drops all but the first message per envelope (TS parity, warned in logs) — real user-visible message loss if Meta batches after downtime.
- **[G1]** `.github/workflows/cabal-whatsapp-bot-tests.yaml` triggers only on `Backend/lib/whatsapp-bot/**` and `Backend/cabal.project`, so edits to the ~1,300 lines of adapters under `rider-app/Main/src/WhatsappBot/**` will not re-run the golden suite.
- **[G1]** `dev/integration-tests/collections/WhatsappRideFlow/` is added but not referenced by any runner.

No TODO/FIXME/HACK, no commented-out code, no partial functions, no `unsafePerformIO`. Code quality and haddock density in the new modules are genuinely high.

---

### Impact Summary

| Aspect | Status |
|--------|--------|
| Touches ride booking flow | **YES** — search → select → confirm → cancel, plus the `search'` bug above |
| Has DB migrations | YES (2 DDL, 2 feature; all additive) |
| API changes | YES — new public `webhook/whatsapp` routes; client-facing schema backward compatible, but `hasPasses` behavior regresses |
| Commits | 1 (squash not needed) |
| Risk level | **HIGH** |

**BECKN actions potentially affected**: search, on_search, select, on_select, init, on_init, confirm, on_confirm, cancel, on_cancel, status/on_status (via `bookingStatus` polling in the tracker).

### Recommendation

**REQUEST CHANGES** — the `search'` argument-order bug at `API/UI/Search.hs:233` silently regresses `hasPasses` for all app users and corrupts the new column; it must be fixed and the rider-app repl check dispatched before merge.

Backend/app/rider-platform/rider-app/Main/src/WhatsappBot/Adapter/Env.hs
```diff
@@ -0,0 +1,177 @@
...
+resolveMerchant cfg = do
+  merchant <- QMerchant.findById cfg.merchantId >>= fromMaybeM (MerchantDoesNotExist cfg.merchantId.getId)
+  moc <- CQMOC.findById cfg.merchantOperatingCityId >>= fromMaybeM (MerchantOperatingCityNotFound cfg.merchantOperatingCityId.getId)
+  pure (merchant.id, moc.id, mkMerchantCtx cfg.botConfig)
+
+mkClock :: Clock Flow
+mkClock = Clock {now = getCurrentTime, sleepMs = \n -> threadDelay (n * 1000)}
+
+-- | Poll constants are TS-hardcoded (L11).
+mkBotConfig :: MerchantCtx -> Map.Map SupportedLanguage LanguageStrings -> BotConfig
```
kavya-shree-s, line 107: Move hardcoded value to table

Backend/lib/whatsapp-bot/src/WhatsappBot/I18n/En.hs
kavya-shree-s, line 1: remove language files

Backend/lib/whatsapp-bot/src/WhatsappBot/Env.hs
```diff
@@ -0,0 +1,285 @@
...
+-- Instances only: LanguageStrings / LanguageInfo are dot-accessed via RDP
+-- (getField), so their selectors are never referenced by name, but the HasField
+-- instances are needed. (Naming them would trip -Wunused-imports under -Werror.)
+import WhatsappBot.I18n.Types ()
+import WhatsappBot.Types
+
+-- | Static per-conversation config: the allowlist, the resolved merchant, and
+-- the poll-loop constants (TS parity, @engine.ts@/@config.ts@). Held separate
+-- from the handles so tests can vary it without touching the effect layer.
+data BotConfig = BotConfig
```
kavya-shree-s, line 73: Check if its used, if used move it to meta config table

kavya-shree-s (changes requested):

kavya-shree-s:
## Re-review of `60f01f51b8` — and a retraction

Follow-up to my review above. Two things: the critical bug is fixed, and one of my two CRITICAL findings was wrong.

### ✅ [C4] fixed — verified

I byte-compared the full 14,981-line diff of `60f01f51b8` against the revision I reviewed. Exactly one line differs, and it is the fix:

```haskell
-- before
… mbEnableSyncSearch mbHasPasses Nothing
-- after  (API/UI/Search.hs:233)
… mbEnableSyncSearch Nothing mbHasPasses
```

Checked against `search'`'s parameter list at line 239: `Nothing` now lands in `mbIsWhatsappRequest` and `mbHasPasses` in `mbHasPasses`. Both consequences are resolved — `search_request.is_whatsapp_request` no longer picks up the client's `hasPasses`, and `hasPasses` reaches `multiModalSearch` → `postMultimodalInitiateSimpl` again.

### ❌ [C6] retracted — rider-app *is* compiled in CI

My claim that "rider-app was never built by CI" was wrong. I checked whether `dashboard-repl-check` reaches rider-app by reading only the first `dependencies:` block of `rider-dashboard/package.yaml` and concluded it did not. The generated `.cabal` is authoritative, and `rider-dashboard.cabal`'s **library** stanza (line 177) lists `rider-app` in `build-depends`.

The chain the dashboard check actually walks:

```
cabal repl dashboard-repl-check          (workflow line 136)
  └─ dashboard-repl-check  (lib)         "pull in both rider-dashboard and provider-dashboard"
       └─ rider-dashboard  (lib)
            └─ rider-app   (lib)         source-dirs: src, src-read-only
                 └─ whatsapp-bot (lib)
```

`rider-app`'s library is `src/` + `src-read-only/`, which is where all of this PR's Haskell lives — `src/WhatsappBot/**`, `src/API/UI/Search.hs`, `src/Domain/Action/UI/MetaWhatsappWebhook.hs`, `src/App.hs`, `src/Environment.hs`. Since `cabal repl` must build every dependency of its target before loading it, rider-app is fully compiled under `-Werror`, not merely typechecked.

The dashboard workflow's `paths` filter covers `Backend/app/rider-platform/**`, `Backend/lib/**` and `Backend/cabal.project` — all touched here — so it fired, and **Incremental Repl Check — Dashboard** passed in 26m3s on `60f01f51b8`.

That also explains the `workflow_dispatch:`-only state I read as a gap: b5caa0faa8 ("feat : enable the new pr workflow (dashboard)", #15859) deliberately moved the `pull_request` trigger off the rider and driver workflows onto the dashboard one, because `dashboard-repl-check` transitively covers all three. Consolidation, not an omission. Apologies for the noise.

Also correcting the header stat in my earlier comment: it is **+13,889 −110**, not `−0`. No finding depended on it.

### Revised recommendation

With C4 fixed and C6 withdrawn, there is no remaining blocker.

**APPROVE**, with these as follow-ups rather than merge blockers:

- **[I2]** `0049-remove-meta-cloudapi-merchant-service-config.sql` and `Tools/Meta.hs:24-30` cite three migration files that do not exist (`0047-meta-config-access-token-backfill.sql`, `1561-meta-config-rename-and-access-token.sql`, `1562-meta-config-access-token-not-null.sql`; ddl-migrations jumps 1558 → 1563). More importantly there is no seed INSERT for `meta_config` anywhere, while `0049` deletes the `Meta_CloudApi` row that currently holds the outbound credentials. Worth resolving before the feature is enabled in any environment, since nothing works until a row exists with Passetto ciphertext in all three secret columns.
- **[I3]** `getWebhookChallenge` (`MetaWhatsappWebhook.hs:78-94`) runs `QMWC.findAll` plus one Passetto `decrypt` per row on an unauthenticated, unrate-limited public GET — and `anyM` short-circuits, so a *wrong* token is the expensive path. Worth fixing before the route is publicly reachable. The repo already has the pattern: `PartnerOrganization.apiKeyHash` + `Tools/Auth.hs:196-198` (`getDbHash` is local, no I/O), so a `verifyTokenHash :: DbHash` column plus `findByVerifyTokenHash` makes it one local hash and one indexed lookup.

The rest — tracker fan-out over the global `wab:activerides` set, unenforced `merchantLabel` uniqueness, session read-modify-write races, NammaDSL generator drift, the G3 items — are fine as post-merge cleanups at pilot scale.

Backend/app/rider-platform/rider-app/Main/src/WhatsappBot/Adapter/Env.hs
```diff
@@ -0,0 +1,177 @@
...
+    { merchant = ctx,
+      flexiQuotePollAttempts = 10,
+      flexiQuotePollIntervalMs = 2000,
+      regularEstimatePollAttempts = 6,
+      regularEstimatePollIntervalMs = 2000,
+      driverPollAttempts = 90,
+      driverPollIntervalMs = 2000,
+      driverPollNotifyEvery = 15,
+      translations = translations
+    }
```
kavya-shree-s, line 119: **Poll tuning is hardcoded here — suggest moving it into `MetaBotCfg` (the `bot_config` jsonb).**

The seam itself is right: the engine never reads a literal, every consumer goes through `env.cfg.*` (`Flow/Booking.hs:325,331,340,353-355,469,475,531,547-549`), which is what the `Monad m`-only invariant at `Env.hs:93-101` requires. So this is purely about where the values live.

**The TS-parity argument doesn't cover this function.** The haddocks pin each value to `engine.ts` lines, which reads as "the golden suite locks these". It doesn't — `test/src/GoldenReplay.hs:599-611` builds its own `fixtureConfig` with the same eight literals. They're duplicated, not shared, so changing `mkBotConfig` moves no fixture. (Independently of the config question: worth a cross-reference comment on each site, since they can now drift silently.)

**`driverPollAttempts = 90 × 2000ms` is the one I'd not leave as a literal.** That's a 180-second poll loop inside the `fork` spawned from the webhook handler (`MetaWhatsappWebhook.hs` `processInbound` → `dispatchInbound`). Per waiting user: one green thread held for up to 3 minutes issuing ~90 `getBookingDetails` calls, each `QRB.findById` + `DBooking.bookingStatus`. N concurrent searchers = N such loops per pod, unbounded. It's exactly the knob you'd want during a driver-supply incident or a DB hot spot, and today shortening it costs a redeploy.

`regularEstimatePollAttempts` (12s) has a smaller version of the same issue: it's a fixed budget for estimates to arrive over `on_search`, while the search window is tunable per city via RiderConfig — tune one and the other doesn't follow, so estimates arriving at 13s read as "no rides found".

**Also inconsistent with the PR's own config story.** `metaTrackerPollMs`, `metaSessionTtlSec` and `metaTrackerMaxAgeSec` are dhall `AppCfg` fields (`Environment.hs:198-202`), and `MetaBotCfg` already carries per-merchant tuning including durations (`flexiRentalDurationS`, `flexiRentalDistanceM`, `flexiServiceRadiusKm`). `metaTrackerPollMs` being configurable while `driverPollIntervalMs` is a literal is hard to justify — same kind of number.

### Suggested approach

`MetaBotCfg` is the best of the three homes: per-merchant (dhall isn't), hot-editable without a redeploy, and **no migration** — it's jsonb, and generic `FromJSON` decodes missing keys on `Maybe` fields as `Nothing`, so existing rows keep working untouched.

```haskell
-- Domain/Types/Extra/MetaWebhookConfig.hs
data MetaBotCfg = MetaBotCfg
  { ...
    -- Poll tuning; Nothing = the TS-parity default applied in mkBotConfig.
    flexiQuotePollAttempts :: Maybe Int,
    flexiQuotePollIntervalMs :: Maybe Int,
    regularEstimatePollAttempts :: Maybe Int,
    regularEstimatePollIntervalMs :: Maybe Int,
    driverPollAttempts :: Maybe Int,
    driverPollIntervalMs :: Maybe Int,
    driverPollNotifyEvery :: Maybe Int
  }
```

```haskell
-- WhatsappBot/Adapter/Env.hs
mkBotConfig :: DMWC.MetaBotCfg -> MerchantCtx -> Map.Map SupportedLanguage LanguageStrings -> BotConfig
mkBotConfig botCfg ctx translations =
  BotConfig
    { merchant = ctx,
      -- Defaults are the TS-port parity values (engine.ts:1074-1078, 825-829,
      -- 890-906), kept in sync with test/src/GoldenReplay.hs:599-611.
      flexiQuotePollAttempts        = fromMaybe 10 botCfg.flexiQuotePollAttempts,
      flexiQuotePollIntervalMs      = fromMaybe 2000 botCfg.flexiQuotePollIntervalMs,
      regularEstimatePollAttempts   = fromMaybe 6 botCfg.regularEstimatePollAttempts,
      regularEstimatePollIntervalMs = fromMaybe 2000 botCfg.regularEstimatePollIntervalMs,
      driverPollAttempts            = fromMaybe 90 botCfg.driverPollAttempts,
      driverPollIntervalMs          = fromMaybe 2000 botCfg.driverPollIntervalMs,
      driverPollNotifyEvery         = fromMaybe 15 botCfg.driverPollNotifyEvery,
      translations = translations
    }
```

Both callers (`dispatchInbound`, `buildTrackerDeps`) already have the `MetaWebhookConfig` row in hand, so threading `cfg.botConfig` through is local.

If you'd rather not expand the surface now, the minimum worth doing is `driverPollAttempts` + `driverPollIntervalMs` — those two are the incident lever. The other five are fine as constants for a pilot.

Backend/dhall-configs/dev/rider-app.dhall
```diff
@@ -545,4 +545,9 @@ in  { esqDBCfg
     , bapHostRedirectMap = [] : List { mapKey : Text, mapValue : Optional Text }
     , xyneWebhookSigningSecret = "<XYNE_WEBHOOK_SIGNING_SECRET>"
     , xyneWebhookBearerToken = "<XYNE_WEBHOOK_BEARER_TOKEN>"
+    , metaBotEnabled = True
```
kavya-shree-s: Move these config to table , not config in .dhall file

Backend/dev/feature-migrations/0047-whatsapp-bot-translations-seed.sql
```diff
@@ -0,0 +1,65 @@
+-- WhatsApp bot copy, Phase 1: 6 pilot fields (flexiArrived splits into two DB
+-- keys for its empty/non-empty-OTP branches) x English/Hindi/Kannada only, as
+-- global (merchant_operating_city_id = NULL) rows so any city can override
+-- later with zero code change. Gujarati/Tamil/Telugu are intentionally NOT
+-- seeded in this phase and keep rendering the static compiled copy.
+-- {{0}}/{{1}} are positional placeholders substituted by
+-- WhatsappBot.Adapter.Translations.substitute.
+INSERT INTO atlas_app.translations (id, message_key, language, message, merchant_operating_city_id, created_at, updated_at)
```
kavya-shree-s, line 8: @sathikabegum-nammayatri Move all the keys to table and remove hard-coded translation file

PR #16166, opened 2026-08-10
feat: added in-line (event-driven) recon

bytedex:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added reconciliation for specifically selected source records.
  * Reconciliation now derives the applicable time range automatically.
  * Matching targets, including orphan targets, are handled during reconciliation.
  * Supports matching individual and grouped records.
  * Updates reconciliation entries with event-time information.
  * Synchronizes source record statuses when applicable.
  * Safely completes without processing when no source records are available.

### DIFF
--- a/Backend/lib/finance-kernel/src/Lib/Finance/Reconciliation/Runner.hs
+++ b/Backend/lib/finance-kernel/src/Lib/Finance/Reconciliation/Runner.hs
@@ -19,6 +19,9 @@ module Lib.Finance.Reconciliation.Runner
     RunOutcome (..),
     runNextChunk,
 
+    -- * Inline (event-driven) execution
+    reconcileSources,
+
     -- * Chunk planning & safety window (exported for tests)
     planChunks,
     chunkIsSafe,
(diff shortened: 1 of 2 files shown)


Backend/lib/finance-kernel/src/Lib/Finance/Reconciliation/Runner.hs
```diff
@@ -162,6 +165,46 @@ runNextChunk recipe input = do
         <> "|"
         <> show chunk.to
 
+-- ─── Inline (event-driven) execution ───────────────────────────────────────
+
+classifyInline ::
```
hkmangla: It is exactly same as processChunk function, why don't we just reimport it with this name? Why to create separate function?

hkmangla (changes requested):

Backend/lib/finance-kernel/src/Lib/Finance/Reconciliation/Runner.hs
bytedex: The primary reason was because this was handling on the basis of SourceIds. As most of the code is similar gated the SourceId level logic or Range level logic in a wrapper function which internally calls processChunk only.

hkmangla (approved):

PR #16226, opened 2026-08-13
BPP reAllocation feat adhering to ONDC specs

RitoBose:
## Type of Change

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## Description
Reallocation is now BPP-owned and keeps one order.id per transaction,
gated per-city on the new transporter_config flag `enableBppReallocation`
(default off => existing behaviour, byte-for-byte).

- config: add `enableBppReallocation` (Maybe Bool) to transporter_config;
  dashboard-tunable via config-pilot params.
- ownership: checkIfRepeatSearch allows reallocation when the BPP flag is
  on OR the BAP's isReallocationEnabled is set (was BAP-gated only).
- forced cap: reallocateStaticOffer makes system/forced (activation)
  reallocations respect searchRepeatCounter — "reallocate N times, then
  close" now applies to system reallocations too.
- single booking: under the flag, performStaticOfferReallocation reuses the
  original booking (mint a fresh quote, re-point via updateReallocationReset
  — quoteId / status=NEW / startTime on the same booking.id) instead of
  createNewBookingAndQuote, so a reallocation keeps one bookingId (= one
  ONDC order.id). Off => legacy new-booking path.
- reallocation-window signal: push BookingReallocation on_update
  (RIDE_BOOKING_REALLOCATION fulfillment state, same order.id) at
  reallocation start so the BAP greys out the freed driver and disables
  call/tracking during the window. New OnUpdate variant
  RideBookingReallocationBuildReq + transformer case +
  CallBAP.sendReallocationUpdateToBAP; not gated on isValueAddNP so
  third-party ONDC BAPs receive it.

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an optional transporter setting to enable BPP-managed ride reallocation.
  * Added support for reusing eligible existing bookings during ride reallocation.
  * Added notifications to BAPs when single-booking reallocation succeeds.
  * Added support for processing ride booking reallocation updates.
* **Bug Fixes**
  * Enforced configured limits for forced reallocation attempts.
  * Improved repeat-search eligibility checks during reallocation.


RitikaHanish (changes requested):
[C4/RIDE_FLOW] The reallocation on_update is rejected by nammayatri's own BAP 100% of the time. The feature does not work end-to-end.

The new branch deliberately drops driver details:


Utils.mkFulfillmentV2 Nothing Nothing ride booking Nothing Nothing … (Just $ show EventEnum.RIDE_CONFIRMED) …
mbDriver = Nothing and mbDriverStats = Nothing ⇒ mbDInfo = Nothing ⇒ agentContact = Nothing, personName = Nothing ([Common.hs:588-628](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Utils/Common.hs#L588-L628)).

On the BAP, RIDE_CONFIRMED is routed straight into the assigned-event parser ([OnUpdate.hs:98-101](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/rider-platform/rider-app/Main/src/Beckn/ACL/OnUpdate.hs#L98-L101)), which hard-requires exactly those two fields:


driverName        <- … (.personName)   & fromMaybeM (InvalidRequest "driverName is not present in RideAssigned Event.")
driverMobileNumber<- … (.contactPhone) & fromMaybeM (InvalidRequest "driverMobileNumber is not present in RideAssigned Event.")
[Common.hs:139-140](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/rider-platform/rider-app/Main/src/Beckn/ACL/Common.hs#L139-L140) — verified against origin/main. Every "driver removed" callback 4xxs. Zero rider-app files are in this PR.

[C4] Even with the parser fixed, ad-hoc bookings are blocked by the BAP status guard.


isAssignable booking = booking.status `elem`
  (if booking.isScheduled then [CONFIRMED, AWAITING_REASSIGNMENT, NEW, TRIP_ASSIGNED]
                          else [CONFIRMED, AWAITING_REASSIGNMENT, NEW])
[Common.hs:1574](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs#L1574). After the first assignment the BAP booking is TRIP_ASSIGNED. For a non-scheduled booking, both the driver-removed on_update and the subsequent new-driver on_update throw BookingInvalidStatus "TRIP_ASSIGNED". Scheduled rides squeak through; ad-hoc do not — which directly contradicts the flag's own comment ("BPP owns reallocation for ALL rides (ad-hoc + scheduled)").

The BAP needs a genuine "reallocating" transition (booking → AWAITING_REASSIGNMENT, detach the ride, notify the rider). Note OUBookingReallocationReq already exists on the BAP but is dead code — marked -- not used at [OnUpdate.hs:123](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/OnUpdate.hs#L123) and never produced by the parser. That's the hook this feature should be reviving.

[C4] Booking is left permanently in NEW when reallocation fails after the reset.

Both callers set the booking CANCELLED before reallocation runs ([Internal.hs:217](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/CancelRide/Internal.hs#L217), [Cancel.hs:139](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Cancel.hs#L139)). In performStaticOfferReallocation, QRB.updateReallocationReset flips it back to NEW before handleDriverSearchBatch. If initiateDriverSearchBatch throws, the Left _ fallback calls cancelRideTransactionForNonReallocation Nothing (Just estimateId) with mbNewBooking = Nothing — which never re-sets the status. Result: BPP booking stuck in NEW with a CANCELLED ride, while sendBookingCancelledUpdateToBAP tells the BAP it's cancelled. Divergent state, invisible to any cancelled-booking reconciliation. Pass Just booking on the singleBooking failure path, or reset only after the batch succeeds.

[C4/I2] One BookingCancellationReason row per booking — multi-ride reallocation silently loses driver cancellations. (This is the core "one booking, many rides" breakage you asked about.)

QBCR.upsert keys on bookingId alone, and on the second call updates only bookingId, rideId, reasonCode, additionalInfo — not source, driverId, driverCancellationLocation, driverDistToPickup, createdAt ([BookingCancellationReasonExtra.hs:38-50](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/BookingCancellationReasonExtra.hs#L38-L50)). With N rides on one booking:

Drivers 2..N's cancellations are never attributed to them. findAllCancelledByDriverId / findAllBookingIdsCancelledByDriverId feed cancellation rate, doCancellationRateBasedBlocking and cancellation penalties — every driver after the first cancels a reallocated booking for free. That is an exploitable gap, not just a metrics bug.
The surviving row carries driver 1's driverId/source with driver N's rideId — corrupt attribution for disputes and analytics.
If reallocation eventually succeeds and the trip completes, a BCR row stays attached to a COMPLETED booking; anything joining on BCR mis-classifies it as cancelled.
Key BCR on (bookingId, rideId) — or insert one row per ride — before enabling multi-ride bookings.

IMPORTANT
[I2] The dynamic-offer path is never covered, but its gate was loosened anyway. reallocateDynamicOffer passes singleBooking = False unconditionally (handleDriverSearchBatch driverSearchBatchInput booking searchTry.estimateId False False), so OneWay/Ambulance/Delivery OneWayOnDemandDynamicOffer — the main NY ad-hoc flow — never gets BPP reallocation. Yet checkIfRepeatSearch now returns True on bppReallocationAllowed || fromMaybe False isReallocationEnabled. Turning the flag on therefore forces legacy ESTIMATE_REPETITION reallocation onto BAPs that explicitly set isReallocationEnabled = False — a behaviour change for partners who opted out, shipped under a flag whose comment promises the opposite. Either wire singleBooking into the dynamic path, or scope the bppReallocationAllowed || override to the static-offer branch only.

[I2] Stale OTP and stale fulfillment id in the driver-removed on_update. mkFulfillmentV2 sets fulfillmentId = ride.id — the cancelled ride — and mkStopsOUS booking ride rideOtp embeds that ride's start OTP. The BAP keys state on a dead ride id, and the rider app can surface an OTP belonging to a driver who is gone. For a no-driver state, drop the authorization token and don't reuse the cancelled ride's id.

[G3/I1] Mutating a KV secondary key is an untested path in this codebase. Booking declares quoteId as a KV secondary key — $(enableKVPG ''BookingT ['id] [['quoteId], ['specialZoneOtpCode], ['transactionId]]) ([Beam/Booking.hs:127](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Storage/Beam/Booking.hs#L127)). updateReallocationReset is the only place in driver-app that Se.Sets one. The whole reassignment then depends on it: SharedLogic/SearchTry.hs, SendSearchRequestToDrivers.hs and Domain/Action/UI/Driver.hs all resolve the booking via QRB.findByQuoteId. If the KV connector doesn't reindex on update, the new driver's accept either resolves the wrong booking or throws BookingDoesNotExist. Please validate on a KV-enabled env, not DB-only.

[I6] sendReallocationUpdateToBAP is a near-verbatim copy of sendQuoteRepetitionUpdateToBAP ([CallBAP.hs:1249](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/CallBAP.hs#L1249)) and does wasted work: it fetches driverStats, merchantPaymentMethod, bppConfig, riderDetails and runs a Passetto decrypt on rider PII — all of which the transformer discards (it passes Nothing for driver, stats, vehicle and riderPhone). That's four extra reads plus a decrypt round-trip on the cancel hot path, per reallocation. Factor out one helper parameterised by the build req and skip the decrypt.

[I2] Duplicate entries in the scheduled-booking Redis set. The driver-cancel path (cancelRideTransaction) does not call removeBookingFromRedis — only the BAP-cancel path does ([Booking.hs:83](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Booking.hs#L83)). addScheduledBookingInRedis targetBooking now re-runs for the same booking id with startTime = max now booking.startTime. Since the sorted-set key is derived from localStartTime, an overdue scheduled booking lands under a new key while the old member survives. Previously each reallocation created a distinct booking, so this never overlapped. Add an explicit zRem of the prior member.

[I5] The "x times" cap is implicit. It rides on searchTry.searchRepeatCounter, a search-try counter, not a reallocation counter. It happens to bound the loop today because each round creates a new SearchTry, but nothing enforces that invariant. For a booking that is now long-lived and mutated in place, a dedicated reallocationCount on Booking would make the requirement explicit and auditable.

[I4] ONDC schema nit. The branch emits fulfillmentAgent = Just Agent{contact = Nothing, person = Just emptyPerson{…all Nothing}} rather than omitting the agent. TRV10 validators generally reject an empty person. fulfillmentAgent = Nothing is both spec-cleaner and gives the BAP an unambiguous absence.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -2222,14 +2225,51 @@ respondQuote (driverId, merchantId, merchantOpCityId) clientId mbBundleVersion m
...
+          mbFarePolicyForCommission <- getFarePolicyByEstOrQuoteIdWithoutFallback driverQuote'.id.getId
+          commission <- FC.calculateCommission driverQuote'.fareParams mbFarePolicyForCommission
+          cancellationCommission <- FC.calculateCancellationCommission driverQuote'.fareParams mbFarePolicyForCommission
+          let (mbPaymentCharge, mbPaymentChargeBearer) = FC.calculatePaymentCharge ((.driverWalletConfig) <$> Just transporterConfig) driverQuote'.fareParams
+          now <- getCurrentTime
+          QBE.updateReallocationResetDynamic driverQuote' commission cancellationCommission mbPaymentCharge mbPaymentChargeBearer now reusedBooking.id
+          QST.updateStatus DST.COMPLETED searchTry.id
+          QBooking.updateDqDurationToPickup reusedBooking.id sReqFD.durationToPickup
+          mFleetAssociation <- QFDA.findByDriverId driver.id True
+          uBookingPre <- QBooking.findById reusedBooking.id >>= fromMaybeM (BookingNotFound reusedBooking.id.getId)
```
dushyant-2002: Just before this, we read the booking:

uBookingPre <- QBooking.findById [reusedBooking.id](http://reusedbooking.id/) >>= fromMaybeM (BookingNotFound reusedBooking.id.getId)

Since the booking has already been updated, due to replication lag, we may still get the old booking data from the replica. This could result in the ride and ledger entry being created with the old fare in the initializeRide function below.

Suggestion :
you can use inMem booking object like this driverFeeUpdateWithPlanAndOffer in Calculate driver Fee job

PR #16264, opened 2026-08-17
feat: impl for FRFS ticketing override with pass - full/partial fare coverage; cancellation overrides; trip limits

pranavs6:
## Description

Adds **purchased-pass fare overrides** to FRFS ticketing. A pass can now cover a booking fully or partially, with limited or unlimited trips, across single-mode and multimodal journeys. Trips are metered in Redis and mirrored to Postgres, debited once the booking reaches `CONFIRMED` (i.e. after the operator has actually issued the ticket) and credited back on cancellation. Adds a per-pass cancellation allowance layered on the existing service-tier quota, and records on `frfs_recon` which pass funded a trip.

## How it works

Three moving parts:

1. **Offer** (`SharedLogic/FRFSPassOverride.hs`) — decides which passes apply to a quote and what they'd cost.
2. **Spend** — the trip is debited in `OnConfirm.onConfirm`, immediately after the booking becomes `CONFIRMED`. Not before: anything failing between a debit and the ticket — a thrown BPP call, a pod restart, an abandoned journey — used to strand the trip with nothing left to release it. Nothing is spent until the ticket exists, so an abandoned booking costs nothing.
3. **Release** — cancellation credits back exactly what was spent, guarded on the booking having reached `CONFIRMED`.

Because the debit runs after the ticket exists, it cannot refuse. Eligibility is settled earlier, where refusing is still free: the offer side only lists passes with `availableTripCount >= quantity`, `resolvePassOverride` re-checks at booking creation, and `validatePassSelections` checks trips against leg count for a whole journey. If those are ever bypassed, `spendTripForBooking` logs `ticket issued on an exhausted pass` — **worth alerting on before real volume**.

Where state lives:

| State | Source of truth | Notes |
|---|---|---|
| Remaining trips | Redis `FRFSPassOverride:availableTripCount-<purchasedPassPaymentId>` | `purchased_pass_payment.available_trip_count` is a **mirror**; Redis reseeds from it only if the key is lost. Spends write the mirror **downwards only** (`updateAvailableTripCountIfLowerById`), so two concurrent spends landing out of order cannot leave it above the true count. To adjust a balance by hand: update Postgres **then** delete the Redis key, in that order |
| "Does this rider own any pass?" | `person.has_pass_till` + Redis `FRFSPassOverride:HasPass-<personId>` | Cheap gate so riders without passes skip the lookup entirely |
| Benefit config | `pass.override_benefit_config_json` via `CachedQueries.Pass` | Redis behind a **1h in-process cache with no invalidation** |
| Which pass paid | `frfs_ticket_booking` → `frfs_recon` override columns | Joins to the `BUS_PASS` recon row |

One trip is debited **per ticket**, not per booking; `maxTicketQuantityPerOverride` caps how many tickets one booking may cover. On a multimodal journey, one trip is debited per pass-covered leg.

```mermaid
flowchart TD
    A["POST /frfs/search<br/>hasPasses, tripTime"] --> B{has_pass_till<br/>set?}
    B -- no --> C["quotes, no passes"]
    B -- yes --> D["load purchased passes<br/>→ pass_id → benefit config"]
    D --> E["filter: vehicle type, date window,<br/>tier, quantity cap, trips ≥ qty"]
    E --> F["GET /quote<br/>applicablePasses[]"]
    F --> G["POST /confirm<br/>purchasedPassPaymentId"]
    G --> H{"re-validate<br/>(resolvePassOverride)"}
    H -- not applicable --> I["400 InvalidRequest"]
    H -- ok --> J["stamp booking:<br/>override_type, overridden_amount,<br/>override_applied_entity_id"]
    J --> K{fully covered<br/>& not a journey leg?}
    K -- yes --> L["claim NEW→CONFIRMING under lock<br/>(claimBookingForConfirm) → BPP confirm"]
    K -- journey leg --> M["defer to FRFSPassConfirm:<br/>journey payment success, or all legs covered"]
    K -- partial --> N["payment order for overridden_amount<br/>→ BPP confirm after payment success"]
    L --> O["on_confirm → CONFIRMED<br/>→ debit trips (DECRBY qty)<br/>→ frfs_recon face fare + override columns"]
    M --> O
    N --> O
    O --> P{cancel or failure?}
    P -- cancel --> Q["quota check → credit trips<br/>(INCRBY qty, clamped to pass capacity)"]
    P -- failure --> R["release only if the booking<br/>reached CONFIRMED"]
```

## Database changes

All new columns are **nullable**; no backfills, no table rewrites.

### `atlas_app.pass`
| Column | Type | Null | Purpose |
|---|---|---|---|
| `frfs_price_override_applicable` | `boolean` | yes | Opts the pass into FRFS fare override |
| `override_benefit_config_json` | `jsonb` | yes | Benefit config: saving %/fixed, max trips, unlimited flag, max tickets per override |
| `frfs_cancel_limit` | `integer` | yes | Per-pass cancellation allowance; replaces the tier's count, borrows the tier's window |
| `vehicle_type` | `text` | yes | Reads as `BUS` when null (`fromTType` default) |

### `atlas_app.purchased_pass`
| Column | Type | Null | Purpose |
|---|---|---|---|
| `vehicle_type` | `text` | yes | Snapshot at purchase; reads as `BUS` when null |

### `atlas_app.purchased_pass_payment`
| Column | Type | Null | Purpose |
|---|---|---|---|
| `pass_id` | `varchar(36)` | yes | Link to `pass` — the only path from a purchase to its pass row |
| `available_trip_count` | `integer` | yes | **Mirror of the Redis counter**, rewritten on each spend/refund; not a source of truth |

### `atlas_app.frfs_ticket_booking`
| Column | Type | Null | Purpose |
|---|---|---|---|
| `override_type` | `text` | yes | `PassOverride` / `OtherOverride` |
| `overridden_amount` | `double precision` | yes | Amount still payable after the pass (`0` = fully covered) |
| `override_applied_entity_id` | `text` | yes | `purchasedPassPaymentId` |

### `atlas_app.frfs_recon`
| Column | Type | Null | Purpose |
|---|---|---|---|
| `override_type` | `text` | yes | Mirrors the booking |
| `overridden_amount` | `double precision` | yes | Mirrors the booking |
| `override_applied_entity_id` | `text` | yes | Joins to the `BUS_PASS` recon row that collected the money up front |

Money columns stay at **face fare**, not `overridden_amount`, and that is deliberate: the operator is owed the fare for a real ride whatever funded it, and the money was already collected by the `BUS_PASS` recon row when the pass was bought — 57,821 such rows already exist in production, avg fare ₹1022.60. Settling `overridden_amount` would pay ₹0 for a ride that happened. `override_applied_entity_id` is what joins the trip back to the purchase that pre-paid for it.

For a pass-covered booking the txn columns (`txn_id`, `transaction_uuid`, `transaction_ref_number`) point at the **pass purchase's charge** rather than the booking's (absent) payment, so they keep holding payment transactions.

This reasoning covers *confirm*. On **cancel** it does not hold — see Known limitations.

### Others
| Table | Column | Type | Null |
|---|---|---|---|
| `atlas_app.frfs_search` | `has_applicable_pass` | `boolean` | yes |
| `atlas_app.integrated_bpp_config` | `pass_override_applicable` | `boolean` | yes |
| `atlas_app.person` | `has_pass_till` | `date` | yes |

> **ClickHouse:** the three `frfs_recon` columns must also be added to the ClickHouse `frfs_recon` table, `frfs_recon_mv` and `frfs_recon_queue`, or they never reach analytics.

## Behaviour changes on the non-pass path

Almost everything here is gated behind `overrideAppliedEntityId` or `isFullyPassCovered`. **One change reaches bookings with no pass**, and it moves money, so it should not be discovered in review:

| Change | Effect on a non-pass booking |
|---|---|
| `FRFSUtils.updateTotalOrderValueAndSettlementAmount` divides the order value by the number of `frfs_recon` rows rather than by `fareParameters.totalQuantity` | **Settlement.** `buildReconTable` has always split the fare across `length tickets` — the tickets the operator actually issued — so the two disagreed. Where the operator issues a single ticket for a multi-quantity booking (the common case: ~1.04 rows/booking at qty 2, 1.25 at qty 5 in production) a row holding the whole fare was rewritten down to a fraction of it on every cancellation. Corrected here, for **all** bookings, not only pass-covered ones |

This is a pre-existing bug that the pass work surfaced — multi-ticket pass bookings are where `ticket_qty` and the row count diverge most — and it cannot be split out, because a multi-ticket pass cancellation depends on the corrected denominator.

Two added costs on shared paths, no behaviour change:

- `FRFSStatus` calls `confirmPassCoveredLegsOfJourney` after a successful confirm, so a non-pass booking pays one `getAllJourneyFrfsBookings` lookup that finds nothing to do. Structural: a paid leg's payment success is the only signal a sibling pass leg can wait on.
- `loadJourneyPassCandidates` runs on the journey-status polling path, adding a `getJourney` read per poll for any journey with a Bus/Metro/Subway leg, whether or not passes are configured in that city.

### Pre-existing bugs deliberately left alone

Found while building this, reverted to keep the blast radius pass-only, and filed separately:

- **Stuck `CONFIRMING` on the shared paid path.** A throw out of the BPP confirm in `FRFSStatus` skips the `FAILED` write and escapes as a 5xx. Production shows 662 bookings stranded, 97% older than a day, the largest bucket with an empty `failure_reason`; SUBWAY is ~10x BUS, consistent with CRIS instability. Recovery exists but is poll-driven, so riders who stop polling are never resolved.
- **Seat holds leak on an unrelated throw** in `confirmAndUpsertBooking`. The pass-specific throw this PR adds (`resolvePassOverride` rejecting an inapplicable pass after the hold exists) *is* handled; other throws keep the pre-existing leak until the hold TTL expires.
- **Duplicate cancel callbacks re-run non-idempotent effects** — the ticket count is decremented again and settlement rewritten on every replay.

## API changes

All additive. No existing field changes shape or meaning.

### Per endpoint

| Endpoint | New in | New out |
|---|---|---|
| `POST /frfs/search` | `hasPasses: Bool` (query)<br>`tripTime: Maybe UTCTime` (body, `FRFSSearchAPIReq`) | — |
| `GET /frfs/search/{searchId}/quote` | `hasPasses: Bool` (query)<br>`tripTime: UTCTime` (query) | `applicablePasses: [FRFSPassOptionAPIEntity]` on each `FRFSQuoteAPIRes` |
| `POST /frfs/quote/{quoteId}/confirm`<br>`POST /frfs/quote/v2/{quoteId}/confirm` | `purchasedPassPaymentId: Maybe (Id PurchasedPassPayment)` (body, `FRFSQuoteConfirmReq`) | `overrideType`, `overriddenTotalPrice`, `appliedPurchasedPassPaymentId` on `FRFSTicketBookingStatusAPIRes` |
| `GET /frfs/booking/{bookingId}/status`<br>`GET /frfs/booking/list` | — | same three fields on `FRFSTicketBookingStatusAPIRes` |
| `GET /frfs/booking/{bookingId}/canCancel/status` | — | `cancelOverrideEntityType`, `cancelOverrideEntityId`, `cancelOverrideEntityName` on `FRFSCanCancelStatus` |
| `POST /frfs/booking/{bookingId}/canCancel` | — | may now fail with `FRFS_CANCELLATION_LIMIT_REACHED` (400) |
| `POST /multimodal/{journeyId}/initiate` | `hasPasses: Bool` (query) | `applicablePasses` per leg on `JourneyInfoResp` |
| `POST /multimodal/{journeyId}/confirm` | `purchasedPassPaymentId: Maybe (Id PurchasedPassPayment)` (body, `JourneyConfirmReqElement`, per leg) | — |
| `GET /multimodal/pass/list` | — | `availableTripCount: Maybe Int`, `unlimitedTripCount: Bool` on `PurchasedPassAPIEntity`; `vehicleType: VehicleCategory` on the nested `PassAPIEntity` |
| `GET /multimodal/pass/availablePasses` | — | `vehicleType: VehicleCategory` on `PassAPIEntity` |

### New/changed response types

**`FRFSQuoteAPIRes`** — `applicablePasses: [FRFSPassOptionAPIEntity]` (new, non-null list)

**`FRFSPassOptionAPIEntity`** (new type)
| Field | Type |
|---|---|
| `purchasedPassPaymentId` | `Id PurchasedPassPayment` |
| `passId` | `Id Pass` |
| `passName` | `Maybe Text` |
| `overriddenUnitPrice` | `PriceAPIEntity` (per ticket) |
| `overriddenTotalPrice` | `PriceAPIEntity` (whole booking) |
| `availableTripCount` | `Maybe Int` |
| `unlimitedTripCount` | `Bool` |
| `maxTicketQuantityPerOverride` | `Maybe Int` |

**`FRFSTicketBookingStatusAPIRes`**
| Field | Type |
|---|---|
| `overrideType` | `Maybe OverrideType` |
| `overriddenTotalPrice` | `Maybe PriceAPIEntity` |
| `appliedPurchasedPassPaymentId` | `Maybe (Id PurchasedPassPayment)` |

**`FRFSCanCancelStatus`** (`GET /frfs/booking/{bookingId}/canCancel/status`) — adds, on top of the existing `cancellationsUsed` / `maxCancellationCount`:
| Field | Type |
|---|---|
| `cancelOverrideEntityType` | `Maybe OverrideType` |
| `cancelOverrideEntityId` | `Maybe Text` |
| `cancelOverrideEntityName` | `Maybe Text` |

These identify the pass whose allowance is in force. All stay `Nothing` when no pass quota applies, so a client can distinguish "tier quota" from "pass quota". `POST /frfs/booking/{bookingId}/canCancel` returns `FRFS_CANCELLATION_LIMIT_REACHED` (400) once the allowance is spent.

**`PassAPIEntity`** — `vehicleType: VehicleCategory` (**non-optional**)

**`PurchasedPassAPIEntity`** — `availableTripCount: Maybe Int`, `unlimitedTripCount: Bool`

> `tripsLeft` on `PurchasedPassAPIEntity` is unchanged by this PR and derives from `maxValidTrips - usedTripCount`. Since `usedTripCount` is never incremented, it reports the full allowance and will **not** track the metered count — clients should read `availableTripCount`.

## Known limitations

- A fully-covering pass is not offered on **ONDC** providers (`coverageSupported`) — the rider sees no pass and no reason why.
- `pass.frfs_cancel_limit` is read through a 1h in-process cache with no invalidation: changes need a restart, and during that window different pods enforce different limits.
- The pass cancel limit borrows the tier's **window**, so "2" means 2 per tier window, not 2 per pass.
- A pass limit is inert unless the tier has both cancellation columns set; a half-configured tier is treated as **uncapped**.
- Partial-discount coverage is implemented but has no automated test.
- **A pass-covered leg is ticketed before a Taxi leg in the same journey is paid for.** Deferring it instead would strand it: the deferred confirm only fires on FRFS payment success, and a taxi leg creates no FRFS payment order, so the leg would sit in `NEW` until `validTill` swept it to `FAILED` — no ticket and no error. Issuing early is the lesser fault until there is a trigger that can fire for a taxi-funded journey.
- **A cancelled pass-covered booking still settles the operator the face fare.** For a paid booking `totalOrderValue` subtracts the refund once the payment is `REFUND_PENDING`; a pass-covered booking has no payment row, so recon keeps the full fare even though the rider's trip was credited back.
- **Lowering `maximum_trip_count` on a live pass retroactively truncates existing balances.** `refundTrip` clamps a credit to the *current* config value, so a rider holding more trips than a newly-lowered cap loses the difference on their next cancellation, with only a `logWarning`. The clamp itself is needed — it stops a repeated release inflating a pass beyond what it was sold with — but it reads live config rather than what the pass was sold with. Treat `maximum_trip_count` as append-only in ops.
- Trip-accounting idempotency rests on Redis markers with a 24h TTL (`TripConsumed` / `TripReleased`, plus `TripRefundPending` to make a failed credit retryable without double-crediting). Beyond that window the booking-status transitions are the only guard.
- `dropPassOverrideOnFareChange` drops the override rather than recomputing it when the BPP changes the fare at `on_init`. Measured over 30 days / 3.7M bookings, `is_fare_changed` was never true, so this path has never fired in production.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added pass-aware fare options across transit search, quotes, confirmations, and multimodal journeys.
  * Eligible purchased passes can reduce or fully cover fares.
  * Added vehicle types, remaining-trip counts, unlimited-trip indicators, and pass validity tracking.
  * Booking and cancellation responses now show applied pass details, fare overrides, and related charges.

* **Bug Fixes**
  * Improved handling of pass-covered bookings, cancellations, refunds, and failed confirmations.
  * Prevented unnecessary payment creation when passes fully cover fares.
  * Improved pass-trip tracking and recovery when bookings or confirmations fail.


Backend/app/rider-platform/rider-app/Main/src/Lib/JourneyModule/Base.hs
```diff
@@ -253,17 +257,54 @@ getAllLegsInfoFromLegs ::
...
 getAllLegsInfoFromLegs personId journeyId allLegs = do
-  whenJourneyUpdateInProgress journeyId $
-    mapMaybeM (\leg -> measureLatency (getLegInfo personId True allLegs leg) ("getLegInfo leg: " <> show leg.sequenceNumber <> " mode: " <> show leg.mode)) allLegs
+  whenJourneyUpdateInProgress journeyId $ do
+    mbPassCandidates <- loadJourneyPassCandidates personId journeyId allLegs
+    mapMaybeM (\leg -> measureLatency (getLegInfo personId True allLegs mbPassCandidates leg) ("getLegInfo leg: " <> show leg.sequenceNumber <> " mode: " <> show leg.mode)) allLegs
+
+loadJourneyPassCandidates :: JL.GetStateFlow m r c => Id DPerson.Person -> Id DJourney.Journey -> [DJourneyLeg.JourneyLeg] -> m (Maybe [FRFSPassOverride.PassCandidate])
+loadJourneyPassCandidates personId journeyId legs
+  | not (any (\leg -> leg.mode `elem` [DTrip.Bus, DTrip.Metro, DTrip.Subway]) legs) = pure (Just [])
```
Vignesh-772, line 267: can this be based on the service tier instead of hard filtering the mode?

Vignesh-772 (approved):

Backend/app/rider-platform/rider-app/Main/src/Lib/JourneyModule/Base.hs
pranavs6: The mode check isn't eligibility though, just to weed out walk/taxi-only legs. Per-leg eligibility is filterCandidatesForLeg (vehicleType + validity window + trips remaining), then coversTier at pricing time. Also, serviceTierType is checked for in `passOptionsForQuote` which does 
```
coversTier applicablePass =
      maybe False (`elem` applicablePass.pass.applicableVehicleServiceTiers) mbServiceTier
```

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFS/OnConfirm.hs
```diff
@@ -210,6 +228,15 @@ onConfirm merchant booking' quoteCategories dOrder = do
...
   person <- runInReplica $ QPerson.findById booking.riderId >>= fromMaybeM (PersonNotFound booking.riderId.getId)
+  -- Debit the pass here, once the ticket exists. Everything between a debit and the ticket -- a
+  -- thrown BPP call, a pod restart, a journey the rider abandons -- used to strand the trip with
+  -- nothing left to release it. Swallowed so a metering failure can never undo a confirmed ticket.
+  --
+  -- Guarded on the PRE-update status: validateRequest does not reject an already-CONFIRMED
+  -- booking, so a replayed on_confirm re-runs this handler, and the TripConsumed marker only
+  -- dedupes for passMarkerTtl. Debit strictly on the transition into CONFIRMED.
+  unless (booking.status == Booking.CONFIRMED) $
+    void $ withTryCatch "onConfirm:spendTripForBooking" (FRFSPassOverride.spendTripForBooking person booking {Booking.status = Booking.CONFIRMED})
```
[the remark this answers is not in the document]
pranavs6: Ideally that could happen only when Redis/DB is down, and should be guarded by 0DC on partial failure, so should be fine

PR #16270, opened 2026-08-17
feat: person-entity grants table, encrypted tokenNo/vpa, bulkUpsert and ptList

Aadil-Ahmed-27:
## What this does

  Adds per-person depot (entity) grants to the unified dashboard:

  - `entity_access` join table — a person may hold many depots, scoped per merchant
  - `person.token_no` / `person.vpa` stored encrypted + hashed (same pattern as email)
  - `POST /bap/{merchant}/{city}/person/bulkUpsert` — CSV-driven upsert of PT staff
    (`bulkCreate` kept as a deprecated alias)
  - `GET /admin/person/ptList` — paginated PT-staff listing with role/depot/search filters

  ---

  ## Deliberate decisions (please read before flagging)

  **Soft-deleted depots keep their grants.** `updateEntity` is the restore path
  (`deleted=false` toggles back), so revoking on delete would silently lose data on
  restore. Instead: retired depots are filtered out of every read path (login,
  `profile`, `ptList`), and `bulkUpsert` rejects only *new* attachments to a retired
  depot — re-sending one a person already holds is a no-op, so unrelated edits to
  that person aren't blocked.

  **`entityIds` is three-state.** Absent = leave grants alone; `[]` = revoke all under
  this merchant; non-empty = replace. "Absent means unchanged" is what keeps payloads
  predating this feature behaving exactly as before. The deprecated scalar `entityId`
  is merged in, but an explicit `[]` wins over it.

  **Role: request-level `roleId`, per-row `roleName` still accepted.** The upload
  screen picks one role from a dropdown and the CSV has no role column. Per-row
  `roleName` stays because `bulkCreate` is still live and existing CSVs carry it.
  Both paths resolve to the same role and pass the same `accessibleRoles` gate.

  **`accessibleRoles` gate exempts admins.** Pre-existing policy, unchanged — mirrors
  `Roles.listV2`. Non-admins reach `bulkUpsert` via the `DASHBOARD_USER_BULK_CREATE`
  capability and *are* gated; `createPerson` has no gate because it's `DASHBOARD_ADMIN`-only.

  **`ptList` returns `tokenNo` / `vpa` in plaintext.** Intentional — it's the PT-staff
  management screen. Everywhere else passes `HideTokenNo`; only `/user/profile`
  (a person reading their own record) also shows it.

  **The `ptList` filter is written twice.** `aggregate_` nests its inner query at a
  different Beam scope than the paged select, and `MonoLocalBinds` won't generalise a
  local binding closing over the arguments — one binding cannot serve both. The two
  copies must stay in sync; a page disagreeing with its own count is worse than the
  duplication.

  **`profile` logs dangling grants instead of erroring.** `entity_access` has no FK, so
  a deleted depot can leave a dangling row. Profile is login-critical — a hard 400
  would lock the person out over one bad row. It logs and returns what resolves.

  ---

  ## Known limitations (tracked, not oversights)

  - **Person writes and grant writes aren't in one transaction.** Making them atomic
    means bypassing the KV write path for updates, which would leave Redis serving
    stale person rows — worse than the failure it fixes. `bulkUpsert` is idempotent,
    so retrying the same CSV converges.
  - **Grant creates are per-row.** The KV layer has no bulk create; raw SQL inserts
    would bypass KV on a table with a `personId` secondary key. Revocations *are*
    batched into one delete.
  - **`ptList` count is skipped only when derivable** (`offset=0` and a short page).
    A true single-pass count needs `COUNT(*) OVER ()`, which has no precedent in this
    repo — every paginated query here uses the same two-query shape.
  - **Same-batch `tokenNo` handoff is rejected.** Conflicts are snapshotted before rows
    resolve, so A-frees-token / B-claims-token in one CSV false-fails. Workaround: two
    uploads. Fix needs two-pass row resolution.
  - **The deprecated `entityId`/`entityName` scalars pick an arbitrary depot** when a
    person holds several — stable across requests (`createdAt`, then `id`), but not
    request order. Honouring request order needs a sort column on `entity_access`.

  ---

  ## Deploy notes

  1. Run migrations **before** the new binary — Beam selects `vpa_*` / `token_no_*`.
  2. `0018-super-admin-seed.sql` is hand-run and **required**: with no `SUPER_ADMIN`,
     nobody can create or modify a `DASHBOARD_ADMIN` (bare `ACCESS_DENIED`, no hint).
  3. Capability seeds add `admin.user.write → DASHBOARD_USER_BULK_CREATE`. Without it
     `Capability.enforce` fails closed and `bulkUpsert` is 403 for everyone.
  4. `capability_endpoint` PK is `(capability_id, server_name, endpoint_id)` — the
     3-column form is required for `ON CONFLICT` to resolve.


Backend/app/dashboard/Lib/src/Domain/Types/Person/API.hs
```diff
@@ -48,14 +53,58 @@ data AvailableCitiesForMerchant = AvailableCitiesForMerchant
   }
   deriving (Show, Generic, FromJSON, ToJSON, ToSchema)
 
--- mb-prefixed params to avoid shadowing Person's own `entityId` field bound by RecordWildCards.
-makePersonAPIEntity :: DecryptedPerson -> DRole.Role -> [ShortId DMerchant.Merchant] -> Maybe [AvailableCitiesForMerchant] -> Maybe (Id DEntity.Entity) -> Maybe Text -> PersonAPIEntity
-makePersonAPIEntity Person {..} personRole availableMerchants availableCitiesForMerchant mbEntityId mbEntityName =
+-- tokenNo authenticates a PT login, so callers state whether this response is allowed to carry
+-- it. Only the person's own profile does; the admin list does not (ptList is the place for that).
+data TokenNoVisibility = ShowTokenNo | HideTokenNo
```
Jayanth-Parthsarathy: Check why this is needed

Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
```diff
@@ -453,11 +460,58 @@ listPerson _ mbSearchString mbLimit mbOffset mbPersonId = do
   res <- forM personAndRoleList $ \(encPerson, role, merchantAccessList, merchantCityAccessList) -> do
     decPerson <- decrypt encPerson
     let availableCitiesForMerchant = makeAvailableCitiesForMerchant merchantAccessList merchantCityAccessList
-    pure $ DP.makePersonAPIEntity decPerson role (nub merchantAccessList) (Just availableCitiesForMerchant) Nothing Nothing
+    pure $ DP.makePersonAPIEntity decPerson role (nub merchantAccessList) (Just availableCitiesForMerchant) [] DP.HideTokenNo
```
Jayanth-Parthsarathy: Use the same qualified import name

Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
```diff
@@ -128,6 +131,51 @@ findAllByIds ::
...
+    L.findRows $
+      B.select $ do
+        access <- B.all_ (SBC.merchantAccess SBC.atlasDB)
+        person <- B.join_' (SBC.person SBC.atlasDB) (\person -> BeamMA.personId access B.==?. BeamP.id person)
+        B.guard_' (BeamMA.merchantId access B.==?. B.val_ (getId merchantId) B.&&?. B.sqlBool_ (BeamP.tokenNoHash person `B.in_` (B.val_ . Just <$> tokenHashes)))
+        pure (BeamP.tokenNoHash person, BeamP.id person)
+  case res of
+    -- A swallowed error here would read as "tokenNo is free" and let duplicates through.
+    Left err -> throwError (InternalError $ "findTokenNoConflictsForMerchant failed: " <> T.pack (show err))
+    Right rows -> pure [(h, Id pid) | (Just h, pid) <- rows]
```
Jayanth-Parthsarathy: Try to see if this can be made into kv like queries

Backend/dev/ddl-migrations/provider-dashboard/0108-add-vpa-and-token-no-encrypted.sql
```diff
@@ -0,0 +1,32 @@
...
+-- Deploy order: APPLY THIS MIGRATION BEFORE the new provider-dashboard binary
+-- ships. The binary's Beam SELECT lists `vpa_encrypted` / `vpa_hash` /
+-- `token_no_encrypted`; reading any existing row will 500 if the columns aren't
+-- present yet.
+
+ALTER TABLE atlas_dashboard.person
+  ADD COLUMN IF NOT EXISTS vpa_encrypted text;
+
+ALTER TABLE atlas_dashboard.person
+  ADD COLUMN IF NOT EXISTS vpa_hash bytea;
```
Jayanth-Parthsarathy, line 19: Check this once (compare with email_hash)

Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
```diff
@@ -333,6 +383,118 @@ findAllWithLimitOffset mbSearchString mbSearchStrDBHash mbLimit mbOffset personI
...
+      -- page already bounds the total.
+      totalCount <-
+        if offsetVal == 0 && toInteger (length rows) < limitVal
+          then pure (length rows)
+          else
+            countRes >>= \case
+              Left err -> throwError (InternalError $ "findAllPTWithLimitOffset count failed: " <> T.pack (show err))
+              Right countRows -> pure $ if null countRows then 0 else head countRows
+      pure (page, totalCount)
+  where
```
Jayanth-Parthsarathy, line 499: Try to use KV here as well

Aadil-Ahmed-27: `findAllPTWithLimitOffset` can't go through `findAllWithKV` — it needs four things the KV layer can't express:

1. `JOIN person × role` (it returns `(Person, Role)` pairs)
2. a correlated `EXISTS` on `merchant_access` — this is the tenancy scoping
3. a correlated `EXISTS` on `entity_access` — the depot filter
4. `concat_ [firstName, lastName] LIKE ?`, plus the `COUNT(*)` for `totalCount`

Sequelize's `Clause` GADT is parameterised over a single table:

```haskell
data Clause be (table :: (Type -> Type) -> Type) where
  And :: [Clause be table] -> Clause be table
  Or  :: [Clause be table] -> Clause be table
  Is  :: Column table value -> Term be value -> Clause be table
```

`Is` binds one `Column table value`, and the available `Term`s are `In | Eq | Null | GreaterThan(OrEq) | LessThan(OrEq) | Like | Not` — no join, subquery or aggregate constructor, and no way to express a predicate over two columns. `findAllWithKV :: Where Postgres table -> m [a]` is single-table by construction.

The blocker that matters most is (2): dropping to KV would mean fetching persons unfiltered and doing the merchant scoping in Haskell, i.e. moving a tenancy check out of the database.

This also follows the existing pattern — `findAllWithLimitOffset` in the same file (backing `listPerson`, already on main) uses the same `getReplicaBeamConfig` → `L.runDB` → `B.select` shape for the same reason.

All the single-table access this PR adds does use KV: `Storage.Queries.EntityAccess` (`createWithKV`, `deleteWithKV`, `findAllWithKV`) and the additions to `Storage.Queries.Entity`.

Backend/app/dashboard/Lib/src/Storage/Beam/Person.hs
```diff
@@ -50,8 +50,10 @@ data PersonT f = PersonT
     language :: B.C f (Maybe KET.Language),
     secretKey :: B.C f (Maybe Text),
     is2faEnabled :: B.C f Bool,
+    tokenNoEncrypted :: B.C f (Maybe Text),
```
XyneSpaces, line 53: **[🔴 DANGEROUS] atlas_bap_dashboard.person** — adding `tokenNoEncrypted` (along with the new VPA fields below) to the shared Beam row makes rider-dashboard select columns that its migrations never create, so ordinary person reads can fail with undefined-column errors after deployment. ↳ FIX: add an idempotent rider-dashboard migration for `token_no_encrypted`, `vpa_encrypted`, and `vpa_hash` and run it before the binary, or separate the schema mappings until both tables are compatible.

XyneSpaces (commented):
## NammaYatri Reviewer — nammayatri/nammayatri PR #16270

**Head SHA:** `603c69fd154aaaff0eace0ab56badce5a3af45ed`
**Review rules:** repository-specific rules loaded from `NammaYatri Review — Consolidated`.
**Reviewed:** 49 files, every supplied diff hunk · **Findings:** 1 blocking, 0 major, 0 minor, 0 nits

### Findings

| blocking | R3 / Migration & rolling compatibility | `Backend/app/dashboard/Lib/src/Storage/Beam/Person.hs:53` | The shared `PersonT` now selects `token_no_encrypted`, `vpa_encrypted`, and `vpa_hash` for both dashboard schemas, but this PR adds those columns only to `atlas_dashboard.person`. The rider schema gets only `entity_access`, so rider-dashboard person reads will fail with missing-column errors after the new binary is deployed. |

### Recommended reviewer

**@khuzema786** is already requested for review — selected from the current requested reviewers because the target-repository workload query shows 0 pending review requests, versus 3 for @kumararnab11 and 54 for @hkmangla.

| @khuzema786 | PR-wide dashboard/backend change | currently requested reviewer | 0 | lowest available workload |
| @kumararnab11 | PR-wide dashboard/backend change | currently requested reviewer | 3 | — |
| @hkmangla | PR-wide dashboard/backend change | currently requested reviewer | 54 | — |

Workload scope: target repository.

### 🛢️ Migration Safety — unsafe operations found

cc @qa-nammayatri

| Migration | Tier | Statement | What it breaks | Fix |
|-----------|------|-----------|----------------|-----|
| missing rider-dashboard migration | 🔴 DANGEROUS | Shared Beam mapping adds `token_no_encrypted`, `vpa_encrypted`, and `vpa_hash` for `atlas_bap_dashboard.person` without adding those columns | Any rider-dashboard query decoding `PersonT` can fail at runtime with undefined-column errors | Add an idempotent rider-dashboard migration for all three columns and apply it before deploying the binary, or keep schema-specific Beam mappings until both schemas are compatible. |

QA deploy stance: verify the repository’s affected critical flows, backward compatibility, and old-version behavior after deployment.

Backend/app/dashboard/Lib/src/Storage/Queries/Person.hs
```diff
@@ -333,6 +383,118 @@ findAllWithLimitOffset mbSearchString mbSearchStrDBHash mbLimit mbOffset personI
...
+-- Interpolated into a LIKE pattern, so unescaped %/_ act as wildcards; a bare "%" matched every row.
+escapeLikeLiteral :: Text -> Text
+escapeLikeLiteral = T.concatMap $ \c -> case c of
+  '\\' -> "\\\\"
+  '%' -> "\\%"
+  '_' -> "\\_"
+  _ -> T.singleton c
+
+-- Filter is duplicated, not shared: aggregate_ nests at a different Beam scope than the paged select, so one local binding cannot serve both. Keep the copies in sync.
+findAllPTWithLimitOffset ::
```
Jayanth-Parthsarathy, line 400: Very big query pls add docs explaining what is actually happening

Jayanth-Parthsarathy (approved):

Cyanid0 (approved):

PR #16275, opened 2026-08-17
fix(bpp-metrics): backend_version rename, distance_bucket + pooling version labels

navyaamahajan:
## What

Follow-up to the BPP funnel metrics PR — label-schema fixes reviewed and validated before broad dashboard adoption (metrics are 3 days old, so the breaking rename is cheap now and impossible later).

- **Rename** deployment `version` label → `backend_version` on all BPP funnel counters + driver response counter. Legacy 2023 metrics (`BPP_search_time`, deviation histograms) keep their old label. Existing integ Grafana panels referencing `version` on the new counters need updating.
- **`distance_bucket` label** on every funnel counter, from search-time estimated distance. Edges intentionally match ClickHouse `trip_distance_bin` (`0-5km / 5-12km / 12-30km / >30km`) so Grafana and warehouse agree by construction. This is deliberately NOT `SharedLogic.Pricing.getDistanceBin` (2km bins for pricing Redis keys — that granularity would blow up metric cardinality).
- **`pooling_logic_version` + `pooling_config_version` labels** only on allocation-side counters (`sent_to_driver`, `expired`, `driver_search_request_response_count`) — where `ensurePoolingLogicVersion` has already run. Deliberately NOT on `search_try`: INITIAL tries are created before version assignment, so the label would encode try order, not pooling.
- **Driver response counter**: Reject/Pulled paths now load the search request so labels are consistent across Accept/Reject/Pulled — without this, only accepts would carry version labels and every per-version accept rate would be biased upward. The lookup is **non-throwing** (missing record degrades labels to "unknown") and **forked off the respond hot path** — which also moves the previously-synchronous metrics-label lookups off it, so the reject path is net lighter than before.

## Notes for reviewers

- Label9 on the response counter is the prometheus-client maximum — any future label needs a metric split (documented in-code).
- `BPP_search_request_count` buckets on the BAP-sent `routeDistance` (pre-serviceability); downstream stages bucket on persisted `estimatedDistance`. Per-bucket funnel ratios should start from `search_try`.
- Optional follow-up if even a forked KV read is unwanted: stamp the three label values onto `search_request_for_driver` at ping creation.

## Validation

- All 10 counter label-tuple arities machine-checked against their `P.vector` registrations.
- Pooling-version assignment ordering verified against `ensurePoolingLogicVersion` call flow in the allocator.
- After integ deploy: confirm `pooling_*_version != "unknown"` on batch-0 `sent_to_driver`, and watch VM series growth (~×4–5 expected from distance buckets).

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Cancel.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Cancel.hs
@@ -130,7 +130,7 @@ cancel req merchant booking mbActiveSearchTry = do
     bookingCR <- buildBookingCancellationReason disToPickup currentLocation mbRide
     QBCR.upsert bookingCR
     cityLabel <- SML.getCityLabel booking.merchantOperatingCityId
-    Metrics.incrementRideCancelledCount merchant.shortId.getShortId cityLabel (show booking.vehicleServiceTier) (show bookingCR.source)
+    Metrics.incrementRideCancelledCount merchant.shortId.getShortId cityLabel (show booking.vehicleServiceTier) (show bookingCR.source) (SML.distanceBucketLabel booking.estimatedDistance)
     QRB.updateStatus booking.id SRB.CANCELLED
     when booking.isScheduled $ removeBookingFromRedis booking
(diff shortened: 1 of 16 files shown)


navyaamahajan:
Superseded by #16278 — re-raised from the correct account (Navya-28). Same diff plus review fixes.

PR #16278, opened 2026-08-17
fix(bpp-metrics): backend_version rename, distance_bucket + pooling version labels

Navya-28:
Supersedes #16275 (re-raised from the correct account).

## What

Follow-up to the BPP funnel metrics PR — label-schema fixes reviewed and validated before broad dashboard adoption (metrics are days old, so the breaking rename is cheap now and impossible later).

- **Rename** deployment `version` label → `backend_version` on all BPP funnel counters + driver response counter. Legacy 2023 metrics (`BPP_search_time`, deviation histograms) keep their old label. Existing integ Grafana panels referencing `version` on the new counters need updating.
- **`distance_bucket` label** on every funnel counter, from search-time estimated distance. Edges intentionally match ClickHouse `trip_distance_bin` (`0-5km / 5-12km / 12-30km / >30km`) so Grafana and warehouse agree by construction. Deliberately NOT `SharedLogic.Pricing.getDistanceBin` (2km bins for pricing Redis keys — that granularity would blow up metric cardinality).
- **`pooling_logic_version` + `pooling_config_version` labels** only on allocation-side counters (`sent_to_driver`, `expired`, `driver_search_request_response_count`) — where `ensurePoolingLogicVersion` has already run. Deliberately NOT on `search_try`: INITIAL tries are created before version assignment, so the label would encode try order, not pooling.
- **Driver response counter**: Reject/Pulled paths now load the search request so labels are consistent across Accept/Reject/Pulled — without this, only accepts would carry version labels and every per-version accept rate would be biased upward. The lookup is **non-throwing** (missing record degrades labels to "unknown") and **forked off the respond hot path** — which also moves the previously-synchronous metrics-label lookups off it, so the reject path is net lighter than before.

## Notes for reviewers

- Label9 on the response counter is the prometheus-client maximum — any future label needs a metric split (documented in-code).
- `BPP_search_request_count` buckets on the BAP-sent `routeDistance` (pre-serviceability); downstream stages bucket on persisted `estimatedDistance`. Per-bucket funnel ratios should start from `search_try`.
- Optional follow-up: stamp the three label values onto `search_request_for_driver` at ping creation to remove even the forked lookup.

## Validation

- All 10 counter label-tuple arities machine-checked against their `P.vector` registrations.
- Pooling-version assignment ordering verified against `ensurePoolingLogicVersion` call flow in the allocator.
- After integ deploy: confirm `pooling_*_version != "unknown"` on batch-0 `sent_to_driver`, and watch VM series growth (~×4–5 expected from distance buckets).

## Summary by CodeRabbit

* **New Features**
  * Added configurable distance-based buckets across ride, booking, cancellation, and search metrics.
  * Added pooling logic and configuration labels to allocation-funnel metrics.
  * Expanded driver response metrics with search-funnel context.

* **Improvements**
  * Improved visibility into search requests, driver responses, bookings, rides, and cancellations.
  * Added validation, safeguards, and fallback handling for distance-bucket configuration.
  * Improved cancellation charge and dues handling, preserving outstanding dues when appropriate.


Navya-28:
@coderabbitai review

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1970,8 +1970,13 @@ respondQuote (driverId, merchantId, merchantOpCityId) clientId mbBundleVersion m
...
       when transporterConfig.analyticsConfig.enableFleetOperatorDashboardAnalytics $ Analytics.updateOperatorAnalyticsAcceptationTotalRequestAndPassedCount driverId transporterConfig False False True False
       QSRD.updateDriverResponse (Just Reject) Inactive req.notificationSource req.renderedAt req.respondedAt sReqFD.id
-      (merchantLabel, cityLabel) <- SML.getMetricsLabels merchantId merchantOpCityId
-      Metrics.incrementDriverResponseCounter merchantLabel cityLabel (show sReqFD.vehicleServiceTier) (show sReqFD.batchNumber) (show req.response)
+      -- Forked off the respond hot path (same idiom as the special-zone hop below): the
+      -- label lookup adds a KV read that must not cost the driver any latency, and metrics
+      -- must never alter the respond flow (a lookup miss degrades labels to "unknown").
+      fork "driver-response-metrics" $ do
+        (merchantLabel, cityLabel) <- SML.getMetricsLabels merchantId merchantOpCityId
+        mbSearchReqForMetrics <- QSR.findById searchTry.requestId
```
piyushKumar-1: we have mSReqFD already fetched above, the data we are trying to find below
(`searchReqFunnelLabels mbSearchReqForMetrics`) i.e. distance, config versions are all there in mSReqFD, so we can avoid this extra -> `mbSearchReqForMetrics <- QSR.findById searchTry.requestId`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/MetricsLabels.hs
```diff
@@ -37,3 +38,35 @@ getCityLabel :: (CacheFlow m r, EsqDBFlow m r) => Id DMOC.MerchantOperatingCity
...
   pure $ maybe merchantOpCityId.getId (show . (.city)) mbCity
+
+-- | Distance bucket label for funnel counters. Edges deliberately match the analytics
+-- stack's trip_distance_bin (ClickHouse) so Grafana and warehouse views agree by
+-- construction. "unknown" when no estimate exists — never drop a metric over a label.
+-- NOT the same as SharedLogic.Pricing.getDistanceBin (fine-grained 2km bins for
+-- dynamic-pricing Redis keys) — that granularity would blow up metric cardinality.
+distanceBucketLabel :: Maybe Meters -> Text
+distanceBucketLabel Nothing = "unknown"
+distanceBucketLabel (Just d)
```
piyushKumar-1: agar ye configurable ho jate to maza hi aa jata......

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
navyaamahajan: Good catch — fixed in a96193c.

You're right that `sReqFD` already has all three (`tripEstimatedDistance`, `poolingLogicVersion`, `poolingConfigVersion`), so the `QSR.findById` is gone from both Reject and Pulled — and with it the fork I'd added only to hide that lookup's latency. The Accept path now reads from `sReqFD` too, so all three responses label from one source.

One extra reason this is better than my original: at fan-out the ping row is stamped with `dpwRes.poolingLogicVersion <|> searchReq.poolingLogicVersion` (SendSearchRequestToDrivers.hs:397), i.e. the version the pool **actually ran with** for that batch. `searchReq.poolingLogicVersion` is only the pre-pool snapshot, so refetching it could have mislabelled a batch during a rollout. Net: −15 lines, one less KV read per response, and more accurate labels.

Navya-28: Good catch — fixed in a96193c.

You're right that `sReqFD` already has all three (`tripEstimatedDistance`, `poolingLogicVersion`, `poolingConfigVersion`), so the `QSR.findById` is gone from both Reject and Pulled — and with it the fork I'd added only to hide that lookup's latency. Accept now reads from `sReqFD` too, so all three responses label from one source.

One extra reason this is better than my original: at fan-out the ping row is stamped with `dpwRes.poolingLogicVersion <|> searchReq.poolingLogicVersion` (SendSearchRequestToDrivers.hs:397) — the version the pool **actually ran with** for that batch. `searchReq.poolingLogicVersion` is only the pre-pool snapshot, so refetching it could have mislabelled a batch mid-rollout. Net: −15 lines, one less KV read per driver response, and more accurate labels.

PR #16279, opened 2026-08-17
feat(bpp-metrics): driver supply gauges — online, receiving, accepting, on-ride

Navya-28:
Supersedes #16277 (re-raised from the correct account).

## What

Four **distinct-driver** supply metrics per merchant × operating city, filling the gap that no supply-side signal exists in VictoriaMetrics today:

| Metric | Meaning | Freshness |
|---|---|---|
| `BPP_drivers_online` | on-duty drivers (`active` = ONLINE/SILENT mode, incl. on-ride — same population the dispatch filter uses) | 60s, from read replica |
| `BPP_drivers_receiving_searches` | distinct drivers sent ≥1 search request | per completed 10-min window |
| `BPP_drivers_accepting_searches` | distinct drivers accepting ≥1 request | per completed 10-min window |
| `BPP_drivers_on_ride` | distinct drivers assigned ≥1 ride (scheduled bookings excluded) | per completed 10-min window |

The driver-grain ratio (accepting ÷ receiving) is a new signal: it distinguishes "all drivers accept less" from "fewer drivers do all the accepting", which ping-grain accept rates cannot see.

## Why gauges + Redis sets (not counters)

These are **entity counts, not event counts** — Prometheus counters count events and cannot dedup the same driver across 40 pings. Call sites `sAddExp` the driver id into a 10-min-bucketed Redis set (TTL 2.5 windows, reset per event); a publisher loop forked at app boot (same pattern as kafka-consumers' startup loops) sets the gauges. Gauges are **set from source of truth, never inc/dec on events** — event-driven gauges drift on pod restarts and split across pods.

## Robustness properties (each verified against kernel/source precedent)

- **Recording can never break a business flow** — kernel `sAddExp` catches and logs all Redis errors internally.
- **Count-query errors propagate** — the publisher's per-tick exception guard holds the last good gauge value; coercing to 0 would fake a supply collapse during a replica blip.
- **Missed ticks lose no windows** — the publisher emits every completed-but-unpublished window (capped 3 back, with a warning log); the strictly-increasing bucket range also guards against backward clock steps.
- **No deployment-version label on gauges** — every pod publishes; a version label would double-count supply during rollouts.
- **Scheduled bookings excluded from on-ride** — mirrors the `updateOnRide` guard in the same function.

## Dashboards

Every pod exports its own series: **always aggregate `max by (merchant, city)` — never `sum` across instances.**

## Ops note — migration 0854

Partial index `idx_driver_information_city_active ON driver_information (merchant_operating_city_id) WHERE active = true`, created **CONCURRENTLY** (hot write table; runner applies via autocommit psql, per existing migrations 0822/0830/0847). If a concurrent build fails it leaves an INVALID index — drop and re-run. Pre-creating manually on prod is safe (`IF NOT EXISTS`).

Known limitation (documented in-code): legacy `driver_information` rows with NULL `merchant_operating_city_id` are excluded from the online count, consistent with most per-city queries.

## Validation after integ deploy

- `BPP_drivers_online` appears within ~60s of boot and roughly tracks known supply
- windowed gauges appear after the first full 10-min window
- `max by (merchant, city)` of accepting ÷ receiving produces a sane driver-grain accept ratio
- follow-up candidate: add `sCard` to shared-kernel to make window reads O(1)

## Summary by CodeRabbit

* **New Features**
  * Added driver supply metrics for online, receiving requests, accepting rides, and on-ride drivers.
  * Metrics are tracked by merchant and operating city in rolling time windows.
  * Added monitoring support for driver activity and supply levels.
* **Performance**
  * Improved active-driver lookups by operating city to support metric reporting.

### DIFF
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
@@ -556,6 +556,7 @@ library
       SharedLogic.DriverPool.DriverPoolMigrations
       SharedLogic.DriverPool.LTSDataSync
       SharedLogic.DriverPool.Types
+      SharedLogic.DriverSupplyCounter
       SharedLogic.DriverSupplyMetrics
       SharedLogic.DynamicPricing
       SharedLogic.EventTracking
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/dynamic-offer-driver-app.cabal
@@ -556,6 +556,7 @@ library
       SharedLogic.DriverPool.DriverPoolMigrations
       SharedLogic.DriverPool.LTSDataSync
       SharedLogic.DriverPool.Types
+      SharedLogic.DriverSupplyMetrics
       SharedLogic.DynamicPricing
       SharedLogic.EventTracking
       SharedLogic.External.LocationTrackingService.API.DriverBlockTill
@@ -1008,6 +1009,7 @@ library
       Tools.Metrics.ARDUBPPMetrics.Types
       Tools.Metrics.DriverSearchRequestResponseMetrics
       Tools.Metrics.DriverSearchRequestResponseMetrics.Types
+      Tools.Metrics.DriverSupplyMetrics.Types
       Tools.Metrics.SendSearchRequestToDriverMetrics
       Tools.Metrics.SendSearchRequestToDriverMetrics.Types
       Tools.Notifications
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/App.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/App.hs
@@ -66,6 +66,7 @@ import Network.Wai.Handler.Warp
     setInstallShutdownHandler,
     setPort,
   )
+import qualified SharedLogic.DriverSupplyMetrics as DSM
 import Storage.Beam.SystemConfigs ()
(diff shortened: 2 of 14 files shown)


Navya-28:
@coderabbitai review

Navya-28:
Moving to draft while I act on a design-premise review (three independent passes over the existing codebase, asking *should this machinery exist* rather than *is it correct*).

**Fixed in the latest commit:** the windowed gauges now carry `vehicle_service_tier`, matching every other BPP funnel metric. Supply shortages are tier-specific and a city total blends auto/cab/bike, so a collapse in one tier was hiding behind a flat line. Redis members became `"<tier>|<driverId>"`, so per-tier distinct counts come from the same single read — no extra Redis traffic. `BPP_drivers_online` stays city-only because `driver_information` has no tier column; its help text now states the actual predicate instead of claiming a mode filter the query never applied.

**Considered and rejected — reading supply from ClickHouse instead.** `atlas_kafka.driver_supply_snapshot` does carry online/on-ride per city, and control-center already queries it. But it is too slow to serve a live gauge: control-center races that exact query against a **5s timeout and returns 0** on expiry (`snapshotRepository.ts:576-584`), and it was pulled out of the master-conversion initial load for the same reason. Direct probes from my side timed out at 100s. `driver_information` is KV-disabled, so the indexed replica count is both fresher and cheaper.

**Still open — publisher placement.** It is currently a forever-loop forked at boot, so it runs on *every API pod*: N×pods duplicate replica queries per minute, and it forces the "never sum across instances" caveat. The house pattern for periodic work is a self-rescheduling Allocator job (`SupplyDemandRatio`, `CongestionChargeAvg` are structural twins), which would run once cluster-wide and make the window/interval runtime-tunable via job data. Leaning toward that; flagging before I rework it in case there's a reason the scheduler is the wrong host here.

PR #16285, opened 2026-08-18
feat: msil invoice flow

Yukesh-R:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added invoice PDF generation with optional S3 storage and reuse of existing PDFs.
  * Invoice listings, ride records, and driver ride details can include invoice links and IDs.
  * Completed rides and cancelled bookings can include invoice documents in status updates.
  * Added GST-compliant metadata, B2C QR codes, and configurable invoice footer and signature details.
  * Invoice records now retain rendered data, PDF references, place-of-supply information, and QR data.

* **Bug Fixes**
  * Zero-value wallet cancellations now generate invoices without unnecessary financial transfers.


RitikaHanish (changes requested):
Every new download path serves a corrupt file — S3 stores base64 text, URLs advertise application/pdf.
storeInvoicePdf calls S3.put path pdfBase64. S3.put renders Text verbatim as UTF-8 under application/octet-stream ([S3/Types.hs:42-46](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/lib/beckn-services/src/AWS/S3/Types.hs#L42-L46)) — so the S3 object is the base64 ASCII string, not PDF bytes.

The existing getFinanceInvoicePdf endpoint round-trips fine (S3.get returns the same base64, served in JSON). But all three new consumers hand out a pre-signed URL to that object:

pdfUrl in invoice list (driver UI + dashboard)
invoicePdfUrl in DriverRideRes
ONDC documents[].url with mime_type = "application/pdf"
A browser or the BAP fetching that URL gets an undecodable blob. Use S3.putRaw path (base64-decoded bytes) "application/pdf" ([S3/Types.hs:206](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/lib/beckn-services/src/AWS/S3/Types.hs#L206)) and adjust the read path to re-encode.

RitikaHanish:
@Yukesh-R Please check these comments.

[C-3] Synchronous wkhtmltopdf render on the BECKN /status ACK path.
[Status.hs:77](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Status.hs#L77) calls InvoiceDocument.getInvoiceDocumentUrl booking → generateAndStoreInvoicePdf → renderInvoicePdfBase64 → generateFinanceInvoicePdf, which forks wkhtmltopdf with a 60-second timeout ([InvoicePdf.hs:53-70](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/lib/beckn-services/src/Tools/InvoicePdf.hs#L53-L70)), plus an S3 PUT and a SigV4 presign.

DStatus.handler runs inline before the ACK — [API/Beckn/Status.hs:66](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Status.hs#L66) is above the Callback.withCallback fork. ONDC gateways NACK on multi-second ACK latency. And on a status burst you fork one wkhtmltopdf (~100MB+ RSS) and one qrencode per request → OOM risk. withTryCatch prevents a throw but not the wall clock.

This should be materialised out-of-band (scheduler job on ride-end / cancellation), with getInvoiceDocumentUrl presigning only when pdfS3Path is already set — same read-only shape as getBookingInvoicePdfUrl.

IMPORTANT Issues
[I-1] N+1 on hot list endpoints, and generateDownloadUrl is not cheap.
The comment on mkInvoicePdfUrl says "stays cheap per list item (SigV4 signing is local)". It isn't — generateUrl' calls Amazonka.newEnv Amazonka.discover on every invocation ([S3/Flow.hs:301-317](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/lib/beckn-services/src/AWS/S3/Flow.hs#L301-L317)). That builds a fresh Env + HTTP Manager and walks the credential chain, which on EKS/EC2 means an IMDS/container-endpoint round trip per call.

Applied per-item in three places:

listDriverRides → buildDriverRideResItem → getBookingInvoicePdfUrl — plus a QFInvoice.findById per ride. This is one of the hottest driver-app endpoints; you've added two round trips per row.
getSubscriptionInvoices — per invoice
getFinanceManagementInvoiceList — per invoice, and dashboard pages can be large
Every prior generateDownloadUrl caller in the codebase is single-item or in a job; this PR is the first to loop it. Either batch-fetch invoices and hoist the Env, or return invoiceId only and presign on a separate detail/download call.

[I-2] documents: null is emitted, not omitted.
RideCompletedOrder derives ToJSON from plain defaultOptions and bookingCancelledOrderJSONOptions only sets fieldLabelModifier — neither sets omitNothingFields. So the -- omitted when Nothing comment is wrong; you'll ship "documents": null on every on_status. A strict ONDC schema check on an array-typed property will reject that. (payment: null already ships today, so gateways may tolerate it — but don't rely on that for a new field.)

[I-3] Cancellation-invoice gating change is a ride-flow behaviour change with no test note.
[CancelRide/Internal.hs:337](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/CancelRide/Internal.hs#L337) drops && totalCancellation > 0 from the guard, so every cancellation now creates an invoice when wallet/prepaid is enabled. That's a large new row volume on finance_invoice and a new ₹0-invoice code path through the ledger. QA should test: driver-cancel and rider-cancel at ₹0 fee, then at non-zero fee, verifying no stray ledger entries at ₹0 and that GST tax transactions are absent.

[I-4] B2C QR is shipped structurally incomplete — empty-string sentinels in a GST artifact.
[B2CQRCode.hs:70-72](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Finance/B2CQRCode.hs#L70-L72) hardcodes upiId = "", bankAcc = "", ifsc = "", and gstin falls back to "". A GST B2C dynamic QR without payee details isn't compliant, and the empty-string sentinel violates the Maybe-over-sentinel convention. Either source these from merchant config in this PR or gate QR generation off until they exist — persisting a non-compliant QR to unsigned_qr_code is worse than persisting nothing, because it later reads as done.

Also fmtMoney goes HighPrecMoney → realToFrac → Double → printf "%.2f". Round-half-even on a Double for a tax document risks a QR total that disagrees with the invoice total by a paisa. Round on the HighPrecMoney (Rational) before display.

[I-5] renderedInvoiceJson is written on every invoice and read nowhere.
[Service.hs](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/lib/finance-kernel/src/Lib/Finance/Invoice/Service.hs) stamps a ~20-field JSON snapshot on every invoice creation — two per completed ride (customer + driver), now plus one per cancellation. Nothing in the PR reads it, and it mostly duplicates columns already on the row. On a table growing at ride volume that's meaningful jsonb bloat with no archival story. Either wire the ONDC/system-comms consumer that justifies it, or drop it and reconstruct from the row.

[I-6] ~800 lines of generated-file churn from a mismatched NammaDSL version.
Domain/Types/TransporterConfig.hs regenerates deriving (Generic, Show, ToJSON, ...) as deriving (Generic, (Show), (ToJSON), ...), and the dashboard endpoint files regenerate Maybe Text as Maybe (Text). Storage/Beam/TransporterConfig.hs is 328+/328− of pure reordering. Regenerate with the pinned generator so the diff shows only the real field additions.

PR #16329, opened 2026-08-20
frfs/fix: use live journey-leg bus number for trip-start notification (prodHotPush-BAP)

Jayanth-Parthsarathy:
## Summary
- Same fix as #16319, targeted at `prodHotPush-BAP`.
- `notifyBusTripStartedForTrip` now reads the vehicle number from `journeyLeg.finalBoardedBusNumber` instead of `booking.vehicleNumber`, which is frozen at booking-confirm time and goes stale after an operator bus swap.
- `finalBoardedBusNumber` is seeded at leg creation from `booking.vehicleNumber` and kept fresh independently by the waybill-details fan-out and the QR-boarding-confirm flow, so no other files need to change for this to work correctly.

## Test plan
- [ ] `cabal build all` (rider-app)
- [ ] Trigger a GIMS trip-start on a waybill with a confirmed shuttle/premium-tier booking; confirm the push/WhatsApp notification carries the correct current vehicle number


Jayanth-Parthsarathy:
Superseded — rebranched from the current prodHotPush-BAP tip after it was force-updated; replaced by a fresh PR.

PR #16342, opened 2026-08-20
backend/feat: migrate the metro seller (ONDC:TRV11) into rider-app

Cyanid0:
Brings `public-transport-bpp`'s **seller** half into the Haskell backend: the routes a
buyer app calls when *we sell* metro tickets, for Chennai (CMRL) and Kochi (KMRL).
rider-app already spoke this protocol as a **buyer** — this is the other half.

Adds `search`, `select`, `init`, `confirm`, `status`, `cancel`, `issue`, `issue_status`,
`info` and `receiver_recon` plus their callbacks; the ONDC:NTS10 recon loop settles
against `frfs_recon` and answers signed; operator adapters for CMRL v1/v2 (CDAC) and
KMRL (Axis gateway, JWE-in-JWS over mutual TLS).

## Configuration this PR requires — nothing works until it is seeded

All seller configuration lives on **`integrated_bpp_config`**. Nothing was added to
`beckn_config`, which the whole rider-app shares.

`integrated_bpp_config.operator_config` (jsonb, per operator row) is **required** — the
seller NACKs every `search`/`select`/`init`/`confirm` while it is NULL, by design, so a
half-filled blob fails loudly rather than borrowing another operator's numbers:

```json
{
  "businessTermsUrl": "…", "courtJurisdiction": "…",
  "cancellation": { "termsUrl": "…", "isAllowed": true },
  "maxPaidAreaMinutes": 20,
  "oneWayTicketLimit": 6, "roundTripTicketLimit": 6,
  "ticketValidityLabel": "…", "ticketValidityDuration": "PT30M",
  "recon":      { "domain": "ONDC:NTS10", "coreVersion": "1.0.0", "ttl": "PT300S" },
  "catalog":    { "brandName": "Triffy Metro", "brandLogoUrl": "…" },
  "quoteCache": { "ttlSeconds": 300, "heldTtlSeconds": 1800 },
  "defaultOrderIdPrefix": "TRF",
  "sellerEntityInfo":  { "gst": …, "pan": … },
  "serviceableBapIds": { "<cityCode>": ["<bapId>"] }
}
```

Also required:

| where | what | why |
|---|---|---|
| `integrated_bpp_config.ondc_registry_public_key` / `.ondc_encryption_private_key` | columns | ONDC `on_subscribe` challenge; kept as columns so they can become `EncryptedField` later |
| `config_json → ticketTypeIds` (CMRLV2) | `"GROUP": 104` | multi-passenger app bookings throw without it |
| `config_json` (KMRL) | `metroType`, `defaultTransactionIdPrefix` | `metroType` 1 = rail; Kochi water metro will differ |

**`sellerEntityInfo` carries real GST and PAN.** Seed per environment; never commit it.

**Per-buyer prefixes are no longer derived.** `bapOrderIdPrefixes` and
`bapTransactionIdPrefixes` key on the exact `bap_id`. Unnamed buyers fall back to the
configured default, so ids stay unique but stop identifying the buyer — seed them if KMRL
settlement reconciliation reads that prefix for attribution.

## How this was verified

Not against fixtures. Real buyer requests were pulled from the **live Go service's own
pod logs** (pre-prod and prod), replayed at rider-app, and every callback diffed field
by field against the one the live service actually sent for that same request.

Buyers seen in the wild: Uber, WayQ, ftdev, Rapido — each registered locally under its
*real* `unique_key_id`, so the `keyId` we sign with matches what they send.

Missing fields per callback, measured that way:

| callback | missing |
|---|---|
| `on_search` (discovery + priced) | **0** |
| `on_select` | **0** |
| `on_init` | **0** |
| `on_confirm` | **0** |
| `on_status` | 3 — `billing`, which those requests don't carry |

Fares are checked against the **real operator**: both sides priced by live CDAC agree on
all four journeys tested.

KMRL's JWE-in-JWS was verified against the Go libraries the live service pins
(`lestrrat-go/jwx v1.2.26`, `go-jose v2.6.0`): headers match byte for byte, and a token
built the way `Crypto.hs` builds it round-trips through the Go service's own decrypt path.
Sabotaging raw-DEFLATE, the AAD binding, or `zip:DEF` each turns it red, so the check
isn't vacuous.

## Known gaps — this is not yet cutover-ready

- **ONDC registry identity is the cutover blocker.** ONDC requires `bpp_id` and `bpp_uri`
  to share a domain, and `subscriber_url` is a path under `subscriber_id`. So the seller
  must answer on the *registered* host — repointing `metrobox.triffy.in`'s ingress at
  rider-app avoids a re-registration; moving to a rider-app hostname means a new
  subscriber, new keys and a fresh `on_subscribe` challenge per operator.
- **The operator prefix is now mandatory.** Go double-mounts the seller routes at both
  `/metro/seller/*` and `/{operator}/metro/seller/*`; we only serve the second. Any BAP
  registered against a non-prefixed `bpp_uri` 404s at cutover.
- **Four callbacks are not ported**: `on_info`, `on_issue_status`, `on_settle`,
  `on_report`. `on_settle` has a spec type but no route, so settlement callbacks 404 —
  and the caller there is the settlement agency, not a buyer.
- `on_status` does not echo `billing`; needs the block persisted on the booking.
- **GROUP is now gated on journey type** — a return booking for 2+ passengers used to send
  CDAC product 104 and silently drop the return leg. Fixed, but **not yet confirmed that
  CDAC accepts RJT with `grp_Size > 1`**; a `getFare` A/B settles it without issuing.
- Mutual-TLS clients are **pinned to TLS 1.2**: offering 1.3 fails against a Go server
  with `bad record MAC`. A deliberate downgrade, marked temporary in the source.
- `cancel`, `issue`, `info`, `receiver_recon` have **zero production traffic** in the
  retained window, so they stay fixture-verified only.
- `beckn_config.payment_params_json` is still the `xyz@upi` placeholder, and
  `integrated_bpp_config.provider_name` is empty.
- **KMRL book/status/cancel are implemented and reachable but not driven end to end.**
  Pre-prod's client certificate (`CN=Nammayatri-triffy`) expired 15 Aug 2026, so the
  gateway rejects the handshake at alert 40 and no live call completes. Prod is unaffected
  — it loads a separate credential valid to Feb 2027. Renewing pre-prod's cert is the
  prerequisite for exercising these paths against the real operator.

## Summary by CodeRabbit

* **New Features**
  * Added seller-side metro ticketing support for Search, Select, Init, Confirm, Status, Cancel, and issue-management workflows.
  * Added seller callbacks for catalogs, orders, ticket status, cancellations, seller information, and settlement reconciliation.
  * Added Kochi Metro integration for fares, station data, operating hours, bookings, cancellations, and ticket status.
  * Added seller identity and settlement configuration support.
* **Bug Fixes**
  * Improved duplicate request handling and ticket-status mapping.
  * Enhanced authentication, encryption, certificate validation, and external-provider error handling.
* **Tests**
  * Added coverage for metro encryption, orders, settlement payloads, and secure HTTP connections.

## Summary by CodeRabbit

* **New Features**
  * Added seller-side FRFS metro APIs for search, selection, booking, status, cancellation, information, and issue management.
  * Added Kochi Metro support for fares, stations, bookings, cancellations, status checks, and secure communication.
  * Added ONDC seller onboarding, registration details, site verification, operating hours, journey-specific fares, and quote caching.
  * Added reconciliation and settlement processing with duplicate-request protection.

* **Bug Fixes**
  * Improved metro authentication, response handling, station data parsing, ticket status updates, and provider-name fallback.
  * Added monitoring for external provider failures.


Backend/app/rider-platform/rider-app/Main/src-read-only/Domain/Types/IntegratedBPPConfig.hs
```diff
@@ -24,6 +24,9 @@ data IntegratedBPPConfig = IntegratedBPPConfig
     isTicketValidOnMultipleRoutes :: Kernel.Prelude.Maybe Kernel.Prelude.Bool,
     merchantId :: Kernel.Types.Id.Id Domain.Types.Merchant.Merchant,
     merchantOperatingCityId :: Kernel.Types.Id.Id Domain.Types.MerchantOperatingCity.MerchantOperatingCity,
+    ondcEncryptionPrivateKey :: Kernel.Prelude.Maybe Kernel.Prelude.Text,
```
XyneSpaces, line 27: **[major] R1 — secrets at rest** — `ondcEncryptionPrivateKey` (and `ondcRegistryPublicKey`) is introduced as plaintext `Maybe Text` storage. The PR description explicitly defers conversion to `EncryptedField`, but now is the cheap point to do it: converting after real keys are seeded requires a live data migration over private material. DB readers, dumps, replicas, and analytics copies would expose the private key used to decrypt ONDC `on_subscribe` challenges, enabling subscriber impersonation during onboarding. Please apply the repository's encrypted-field pattern now (as used for KMRL secrets) across the domain/Beam mapping and migration.

.github/workflows/cabal-main-push.yaml
```diff
@@ -294,7 +295,7 @@ jobs:
     name: Analyze Cabal Build Times
     needs: [cabal-build, check-conditions]
     if: needs.check-conditions.outputs.should-proceed == 'true'
-    runs-on: x86_64_main_push
+    runs-on: x86_64_prodHotPush-Common_pus
```
XyneSpaces, line 298: **[major] CI correctness** — `x86_64_prodHotPush-Common_pus` is missing the trailing `h`; the other two replacements in this PR use `x86_64_prodHotPush-Common_push`. Because this PR also adds its branch to the push trigger, the Analyze Cabal Build Times job can remain queued without a matching runner. Change this label to `x86_64_prodHotPush-Common_push`.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/Status.hs
```diff
@@ -0,0 +1,140 @@
...
+reportOrder :: Text -> DBC.BecknConfig -> Spec.StatusReq -> Flow (Either StatusFailure (SpecEnums.OrderStatus, OnConfirmACL.ConfirmedOrder))
+reportOrder operator becknConfig req = runExceptT $ do
+  let orderId = req.statusReqMessage.statusReqMessageOrderId
+  booking <-
+    lift (QBooking.findByBppOrderId (Just orderId))
+      >>= maybe (throwE (OrderNotFound orderId)) pure
+  unless (Common.isSellerRider booking.riderId) $
+    throwE (OrderNotFound orderId)
+  tickets <- lift (QTicket.findAllByTicketBookingId booking.id)
+  integratedBPPConfig <-
```
XyneSpaces, line 84: **[minor] R6** — `mapM (refreshTicket integratedBPPConfig) tickets` performs remote operator status calls strictly in series (up to six per booking under the configured ticket limits). One slow KMRL/CMRL response therefore delays the whole `on_status`, and every poll pays the sum of all round-trip times. Use bounded concurrency and/or avoid refreshing statuses that cannot transition further.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/IGM.hs
```diff
@@ -0,0 +1,224 @@
...
+  let ctx = req.context
+  transactionId <- ctx.contextTransactionId & fromMaybeM (InvalidRequest "TransactionId missing on issue context")
+  messageId <- ctx.contextMessageId & fromMaybeM (InvalidRequest "MessageId missing on issue context")
+  bapId <- ctx.contextBapId & fromMaybeM (InvalidRequest "BapId missing on issue context")
+  bapUriText <- ctx.contextBapUri & fromMaybeM (InvalidRequest "BapUri missing on issue context")
+  bapUri <- parseBaseUrl bapUriText
+  dIssue <- IssueACL.buildIssueReq req
+  seller <- sellerContext operator
+  recordIssue seller dIssue >>= \case
+    Left failure -> logWarning $ "FRFS seller issue rejected: " <> failureMessage failure
```
XyneSpaces, line 101: **[minor] R7** — when `recordIssue` returns `Left`, the request has already been ACKed but this branch only logs, so the buyer receives no `on_issue` rejection and waits out the callback TTL. Other seller flows in this PR build error callbacks on validation failures. Build an `on_issue` error response (unknown order, unsupported transition, missing IGM config) or NACK before acknowledgment so the buyer can render a terminal failure.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/ReceiverRecon.hs
```diff
@@ -0,0 +1,157 @@
...
+reconcileOrder operator oid order =
+  case settlementAmount order of
+    Nothing -> unanswered oid "no readable settlement amount"
+    Just amount ->
+      Redis.whenWithLockRedisAndReturnValue (lockKey oid) 10 (withLock amount) >>= \case
+        Right result -> pure result
+        Left _ -> unanswered oid "another delivery is applying this order"
+  where
+    withLock amount = do
+      mbRow <- QRecon.findByPrimaryKey (Id (Common.sellerReconId oid) :: Id DRecon.FRFSRecon)
```
XyneSpaces, line 91: **[minor] R2** — the recon row is fetched by `Common.sellerReconId oid`, derived only from the collector-supplied order ID, and `applyToRow` then mutates it without checking that it belongs to the operator/merchant resolved for this endpoint. The `Nothing` branch below handles unknown rows, but a found row belonging to another CMRL/KMRL merchant is still mutable through a misrouted payload. Pass the resolved merchant into reconciliation and verify ownership, or include the operator/merchant in the recon key so foreign orders miss.

Backend/app/rider-platform/rider-app/Main/src/ExternalBPP/ExternalAPI/Metro/KMRL/StationList.hs
```diff
@@ -0,0 +1,73 @@
...
+    toCMRLStation sequenceNo station =
+      CMRLStationList.Station
+        { id = 0,
+          lineId = kochiLineId,
+          stationId = station.stationId,
+          code = "",
+          name = station.displayName,
+          taName = Nothing,
+          address = "",
+          latitude = fromMaybe 0.0 station.latitude,
```
XyneSpaces, line 50: **[nit] R2** — defaulting a missing coordinate to `0.0` publishes `(0,0)` (Null Island) as if it were a real station location; the non-optional target type then prevents downstream code from distinguishing missing data. Prefer filtering stations without coordinates, rejecting the malformed operator response, or using an explicit absence representation that the catalog builder can skip.

XyneSpaces (commented):
## NammaYatri Reviewer — nammayatri/nammayatri PR #16342

**Head SHA:** `47f6a8ff875e56b9f01eb8eef31c0b72b6dbe669`
**Verdict:** Changes requested
**Review rules:** repository-specific rules loaded from `NammaYatri Review — Consolidated` (mined rulebook, Phase-3 backend checklist, and mined PR-comment rules).
**Reviewed:** 34 entries returned by the PR-files API, plus branch-side retrieval of FRFSSeller domain/ACL/shared-logic modules, KMRL/CMRL integrations, NTS10 types, and migration SQLs · **Findings:** 0 blocking, 2 major, 3 minor, 3 nits

### Findings

| major | R1 | `src-read-only/Domain/Types/IntegratedBPPConfig.hs:27` | ONDC onboarding keys are stored as plaintext fields; encrypt them before real keys are seeded. |
| major | CI correctness | `.github/workflows/cabal-main-push.yaml:298` | Runner label ends in `Common_pus`, so the build-times job cannot acquire the intended runner. |
| minor | R7 | `src/Domain/Action/Beckn/FRFSSeller/IGM.hs:101` | Rejected `issue` requests only log; the buyer receives no `on_issue` error callback. |
| minor | R6 | `src/Domain/Action/Beckn/FRFSSeller/Status.hs:84` | Ticket status is refreshed through sequential per-ticket operator calls. |
| minor | R2 | `src/Domain/Action/Beckn/FRFSSeller/ReceiverRecon.hs:91` | Settlement lookup is keyed by request-supplied order ID without checking operator/merchant ownership before mutation. |
| nit | R2 | `src/ExternalBPP/ExternalAPI/Metro/KMRL/StationList.hs:50` | Missing coordinates become `(0,0)` and can publish a Null Island station. |
| nit | R10 | `src/Beckn/ACL/FRFSSeller/OnReceiverRecon.hs:27` | Missing `bap_id` becomes an empty collector ID in the callback. |
| nit | R10 | `src/Beckn/ACL/FRFSSeller/OnReceiverRecon.hs:67` | Timestamp formatting hardcodes `.000` instead of preserving/formatting actual fractional seconds. |

Existing CodeRabbit/hlint findings were re-read before posting; unchanged issues already raised there were not duplicated.

### Coverage limitation

The PR-files API returned only its first 34 entries and exposed no pagination control. The page is demonstrably incomplete: `Domain/Action/Beckn/FRFSSeller/Search.hs` is new on this branch but absent from the page, and the changed `migrations-read-only/rider-app/{merchant,integrated_bpp_config}.sql` files are also absent. I retrieved and reviewed missing implementation and migration files directly from the branch, but test files mentioned in the PR description could not be enumerated from that API page and therefore could not be independently reviewed line-by-line. Migration SQL inspected here is additive/nullable; no unsafe migration operation is being reported.

The branch is also behind `main` (notably DDL migrations 1565–1566 and feature migrations 0055–0059), so rebase before merge and re-run migration conflict checks.

### Recommended reviewer

**@Hari-Krishna-Dupaguntla** is recommended and is already requested on this PR — tagging for attention based on direct FRFS familiarity and lower pending workload.

| @Hari-Krishna-Dupaguntla | FRFS booking/init and offer flows (`#15803`, `#16726`) | Existing requested reviewer; recent FRFS author | 7 | Lower queue; active in FRFS today |
| @Vignesh-772 | FRFS/transit backend (`#16523`, `#15803`) | Existing requested reviewer | 10 | — |

Workload scope: **organization-wide** open PRs with the candidate still requested for review. `CODEOWNERS` was not found at `.github/CODEOWNERS`, root `CODEOWNERS`, or `Backend/CODEOWNERS`.

No duplicate reviewer request was issued because both candidates are already requested on GitHub.

XyneSpaces (commented):
## NammaYatri Reviewer — nammayatri/nammayatri PR #16342

**Head SHA:** `3c44021387f0cfa012ae3e99a9ba1fe7243c2c87`
**Review rules:** Repository-specific rules loaded from `NammaYatri Review — Consolidated`.
**Reviewed:** 86 files reported by the PR; all returned changed hunks were inspected. **Findings:** 2 blocking, 3 major, 0 minor, 0 nits.

### Findings

| blocking | Concurrency / idempotency | `Backend/app/rider-platform/rider-app/Main/src/API/Beckn/FRFSSeller/Handler.hs:60` | Redis failure is converted into a process-anyway path, so confirm/issue work can run without a durable dedupe claim and duplicate financial/operator side effects can occur. Fail closed with a retryable NACK, or persist an in-flight claim. Existing CodeRabbit inline comment covers this exact issue. |
| blocking | API correctness | `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/Cancel.hs:165` | Technical cancellation checks for reason `"0"`, while the TRV11 technical-cancellation code is `"000"`; compliant technical cancels will be rejected/misclassified. Use the canonical spec value. Existing CodeRabbit inline comment covers this exact issue. |
| major | API/state consistency | `Backend/app/rider-platform/rider-app/Main/src/Beckn/ACL/FRFSSeller/OnSearch.hs:268` | Search emits the provider stop code instead of the canonical stop identifier used by later seller flows, so a search result can produce a stop that select/init cannot resolve. Resolve/map to the canonical station id. Existing CodeRabbit inline comment covers this exact issue. |
| major | Financial correctness | `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/Confirm.hs` | Partial operator issuance is accepted, but booking/reconciliation metadata can retain the requested quantity rather than the number of tickets actually issued. Reject/compensate partial issuance or persist the issued count consistently. Existing inline review coverage requests rejecting incomplete issuance. |
| major | Idempotency | `Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/FRFSSeller/IGM.hs` | `openIssue` creates a deterministic issue row without an idempotent read/upsert guard; retried issue requests can hit a duplicate-key failure instead of returning the existing issue. Existing CodeRabbit inline comment covers this exact issue. |

### Recommended reviewer

**@Vignesh-772** is already requested for review — the PR affects Beckn/FRFS seller flows, metro operator adapters, and settlement/reconciliation; the existing reviewer request is the strongest available ownership signal.

| @Vignesh-772 | FRFS seller handlers, Beckn ACLs, metro adapters, recon | Already requested reviewer | unavailable | Highest-risk backend/Beckn scope |
| @Hari-Krishna-Dupaguntla | Same affected subsystem | Already requested reviewer | unavailable | Existing reviewer request |

Workload scope: unavailable. Organization-wide pending-review workload could not be calculated with the available GitHub operations; no workload number is claimed.

The selected reviewer is already requested; no additional reviewer-assignment operation was available in this runtime.

### Top blocking issue

The incorrect technical-cancellation code (`"0"` vs `"000"`) can reject valid buyer cancellations, while the fail-open dedupe path can duplicate paid operator actions. Fix both before merge.

Existing CodeRabbit inline comments already cover the exact changed lines for the first three findings; no duplicate inline comments were posted by this reviewer.

PR #16346, opened 2026-08-20
feat(bpp-metrics): driver supply gauges — online, receiving, accepting, on-ride

Navya-28:
Supersedes #16279 — moved off my fork so CI runs without per-push approval.

## What

Two **driver-supply level metrics** per merchant × operating city, filling the gap that no supply-side signal exists in VictoriaMetrics today:

| Metric | Meaning |
|---|---|
| `BPP_drivers_online` | dispatch-eligible drivers (`active` = ONLINE/SILENT, **includes** on-ride — same population the dispatch filter uses) |
| `BPP_drivers_on_ride` | drivers currently on a ride |

Free supply = `BPP_drivers_online − BPP_drivers_on_ride`.

## How

Both are plain **Redis counters**, maintained by the events that change the underlying state:

- **online** — `incr`/`decr` from the six writers of `driver_information.active`
- **on-ride** — `incr` at ride start, `decr` at completion, and at cancellation **only when the ride was `INPROGRESS`** (cancelling a ride that never started must not decrement something that was never counted)

A forked publisher copies both counters into gauges once a minute. It counts nothing — a Prometheus scrape can only read process memory, so something has to move the value across.

## Review changes since the first version

**Counters, not sets of driver ids** (@piyushKumar-1). A set large enough to hold every online driver lands on one Redis shard and grows with the fleet. `pinged` and `accepted` were dropped entirely — they're monotonic, so the existing `BPP_*` counters already answer them. Neither remaining metric needs set semantics: a driver can be online, or on a ride, only once. That removed the window bucketing, member encoding and per-tuple publishing along with them.

**No Postgres on this path.** The merchant/city list is resolved once at boot rather than per tick. The counters seed from **ClickHouse**, and only when a key is absent — in practice once per city per Redis lifetime. The partial index and its migration are gone, since nothing queries PG by `(city, active)` any more.

**Seeding is single-flighted and additive.** A Redis lock ensures one pod establishes the baseline; it uses `incrby`, not `set`, because a concurrent `incr`/`decr` between the ClickHouse read and the write is a real state change that `set` would silently discard. This addresses the CodeRabbit finding about the seed racing counter updates.

**`deactivateRC` takes the row when the caller has it.** `linkRCStatus` already loads `driver_information`, so it passes it down; the other three call sites have no row for that driver and pass `Nothing`. The counter needs both the row's city and `active` — neither is on `Person`, and the city must be the driver's own, not the request's operating city.

## Verified against prod ClickHouse, not assumed

`active` and `on_ride` are `Nullable(String)` with **mixed casing**, not booleans — the current-state distinct set is `FALSE / False / TRUE / True / false`, no NULLs. Matching a single spelling undercounts badly and unevenly: 1,685 instead of 39,334 in one city (**96% low**), 325 instead of 22,292 in another (**98.5% low**). The seed queries match every truthy spelling and were checked against prod — they agree exactly with an independent per-city count.

⚠️ **Pre-existing issue this surfaced, not fixed here:** `enabled` in the same ClickHouse mirror is declared `C f Bool` against the same mixed-casing column, so `countEnabledByDriverIds` and `getOnlineDriverCountByDriverIds` are likely undercounting in production today.

## Dashboards

**Always aggregate `max by (merchant, city)` — never `sum`.** Every pod publishes its own copy; summing multiplies by pod count.

No `backend_version` label, deliberately: during a rollout two versions would publish the same truth under different label sets, so any `sum()` would double-count supply.

No `vehicle_service_tier` either — `driver_information` carries no tier (a driver's tiers come from their vehicle and preferences), so both gauges are city-wide by necessity. Per-tier volumes are already on the `BPP_*` counters.

## Known trade-off

With no periodic reconcile (per review), a missed `incr`/`decr` persists until the key is deleted — which re-seeds from ClickHouse on the next tick. That's the deliberate cost of keeping a database off the tick path.

## Summary by CodeRabbit

* **New Features**
  * Added city-level live tracking for online and on-ride driver supply.
  * Supply metrics now initialize from existing driver data and publish ongoing updates.
  * Driver availability remains accurate when drivers change status, migrate cities, register vehicles, or are removed.
  * Ride starts, completions, and cancellations update on-ride supply counts automatically.

* **Bug Fixes**
  * Improved metric consistency during driver and ride lifecycle changes.
  * Metric storage failures no longer interrupt core driver or ride operations.

### DIFF
--- a/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
+++ b/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
@@ -73,6 +73,7 @@ import qualified Lib.Yudhishthira.Types as LYT
 import qualified Processor.RideEvents.InternalHelpers as IH
 import qualified "dynamic-offer-driver-app" SharedLogic.Analytics as Analytics
 import qualified "dynamic-offer-driver-app" SharedLogic.BehaviourManagement.ConsequenceDispatcher as BehaviorDispatch
+import qualified "dynamic-offer-driver-app" SharedLogic.DriverSupplyCounter as DSC
 import qualified "dynamic-offer-driver-app" SharedLogic.External.LocationTrackingService.Types as LT
 import qualified "dynamic-offer-driver-app" SharedLogic.FleetVehicleStats as FVS
 import "dynamic-offer-driver-app" SharedLogic.Reminder.Helper (checkAndCreateRemindersForRidesThreshold)
@@ -482,7 +483,15 @@ handleDriverCityMigration ev = withRideAndBooking ev $ \ride booking -> do
 
         -- --- operational tables: sync directly, no gating. Document tables are
         -- intentionally excluded -- nothing reads their city column. ---
+        -- A city move relocates an online driver between supply buckets without touching
+        -- `active`, so the counter needs an explicit move. Read before the write, and
+        -- keyed on driver_information's own city -- that is the bucket it was counted in.
+        mbDriverInfo <- QDriverInfo.findById (cast ride.driverId)
         QDriverInfo.updateMerchantIdAndCityIdByDriverId ride.driverId booking.providerId targetOpCityId
+        whenJust mbDriverInfo $ \di ->
+          when di.active $ do
+            DSC.recordDriverActiveChange di.merchantOperatingCityId True False
+            DSC.recordDriverActiveChange (Just targetOpCityId) False True
         QVehicleExtra.updateMerchantIdAndCityIdByDriverId ride.driverId booking.providerId (Just targetOpCityId.getId)
         QDailyStats.updateMerchantIdAndCityIdByDriverId (Just booking.providerId) (Just targetOpCityId) ride.driverId
         QDriverReferral.updateMerchantIdAndCityIdByDriverId (Just booking.providerId) (Just targetOpCityId) ride.driverId
--- a/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
+++ b/Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
@@ -73,7 +73,6 @@ import qualified Lib.Yudhishthira.Types as LYT
 import qualified Processor.RideEvents.InternalHelpers as IH
 import qualified "dynamic-offer-driver-app" SharedLogic.Analytics as Analytics
 import qualified "dynamic-offer-driver-app" SharedLogic.BehaviourManagement.ConsequenceDispatcher as BehaviorDispatch
-import qualified "dynamic-offer-driver-app" SharedLogic.DriverSupplyCounter as DSC
 import qualified "dynamic-offer-driver-app" SharedLogic.External.LocationTrackingService.Types as LT
 import qualified "dynamic-offer-driver-app" SharedLogic.FleetVehicleStats as FVS
 import "dynamic-offer-driver-app" SharedLogic.Reminder.Helper (checkAndCreateRemindersForRidesThreshold)
@@ -483,15 +482,7 @@ handleDriverCityMigration ev = withRideAndBooking ev $ \ride booking -> do
 
         -- --- operational tables: sync directly, no gating. Document tables are
         -- intentionally excluded -- nothing reads their city column. ---
-        -- A city move relocates an online driver between supply buckets without touching
-        -- `active`, so the counter needs an explicit move. Read before the write, and
-        -- keyed on driver_information's own city -- that is the bucket it was counted in.
-        mbDriverInfo <- QDriverInfo.findById (cast ride.driverId)
         QDriverInfo.updateMerchantIdAndCityIdByDriverId ride.driverId booking.providerId targetOpCityId
-        whenJust mbDriverInfo $ \di ->
-          when di.active $ do
-            DSC.recordDriverActiveChange di.merchantOperatingCityId True False
-            DSC.recordDriverActiveChange (Just targetOpCityId) False True
         QVehicleExtra.updateMerchantIdAndCityIdByDriverId ride.driverId booking.providerId (Just targetOpCityId.getId)
         QDailyStats.updateMerchantIdAndCityIdByDriverId (Just booking.providerId) (Just targetOpCityId) ride.driverId
         QDriverReferral.updateMerchantIdAndCityIdByDriverId (Just booking.providerId) (Just targetOpCityId) ride.driverId
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/App.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/App.hs
@@ -68,6 +68,7 @@ import Network.Wai.Handler.Warp
     setInstallShutdownHandler,
     setPort,
   )
+import qualified SharedLogic.DriverSupplyMetrics as DSM
 import Storage.Beam.SystemConfigs ()
 import qualified Storage.CachedQueries.Merchant as Storage
 import System.Environment (lookupEnv)
@@ -141,6 +142,7 @@ runDynamicOfferDriverApp' appCfg = do
         L.setOption KBT.Tables kvConfigs
         _ <- liftIO $ createCAC appCfg
         initCityMaps
+        fork "driver supply metrics publisher" DSM.runDriverSupplyMetricsPublisher
         allProviders <-
           try Storage.loadAllProviders
             >>= handleLeft @SomeException exitLoadAllProvidersFailure "Exception thrown: "
--- a/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
+++ b/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
@@ -271,7 +271,6 @@ import qualified SharedLogic.DriverOnboarding.OnboardingFlags.Types as SOnboardi
 import qualified SharedLogic.DriverOnboarding.Status as SStatus
 import SharedLogic.DriverPool as DP
 import qualified SharedLogic.DriverSupplyCounter as DSC
-import qualified SharedLogic.DriverSupplyMetrics as DSM
 import qualified SharedLogic.EventTracking as ET
 import qualified SharedLogic.External.LocationTrackingService.Flow as LTF
 import qualified SharedLogic.External.LocationTrackingService.Types as LT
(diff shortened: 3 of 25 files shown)


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverInformationExtra.hs
```diff
@@ -550,12 +581,18 @@ countEnabledByDriverIds driverIds =
...
+  -- A city move relocates an online driver between supply buckets without touching `active`.
+  mbDriverInfo <- findById (cast driverId)
   updateOneWithKV
     [ Se.Set BeamDI.merchantId (Just $ getId merchantId),
       Se.Set BeamDI.merchantOperatingCityId (Just $ getId merchantOperatingCityId),
       Se.Set BeamDI.updatedAt now
     ]
     [Se.Is BeamDI.driverId (Se.Eq $ getId driverId)]
+  whenJust mbDriverInfo $ \di ->
+    when di.active $ do
```
piyushKumar-1: do this in `handleDriverCityMigration` instead of doing an extra query in `updateMerchantIdAndCityIdByDriverId`

Backend/dev/ddl-migrations/dynamic-offer-driver-app/0855-driver-supply-metrics-index.sql
```diff
@@ -0,0 +1,8 @@
+-- Partial index for the driver-supply metrics publisher: per-city count of active drivers,
+-- executed on the read replica every 60s per operating city.
+-- CONCURRENTLY: driver_information is one of the hottest write tables (every online/offline
+-- toggle); a plain build would block all writes for the duration. The runner applies these
+-- files via plain psql (autocommit), so CONCURRENTLY is valid here — house style per
+-- migrations 0822/0830/0839/0847. If a concurrent build fails it leaves an INVALID index:
+-- drop and re-run.
+create index concurrently if not exists idx_driver_information_city_active on atlas_driver_offer_bpp.driver_information (merchant_operating_city_id) where active = true;
```
piyushKumar-1: 🔪

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverSupplyMetrics.hs
```diff
@@ -0,0 +1,220 @@
...
+          -- are empty anyway); the strictly-increasing range also guards against a
+          -- backward clock step republishing an already-expired bucket as 0.
+          bucketsToPublish
+            | lastPublishedBucket < 0 = [prevBucket]
+            | otherwise = [max (lastPublishedBucket + 1) (prevBucket - 2) .. prevBucket]
+      when (length bucketsToPublish > 1) $
+        logWarning $ "publishing " <> show (length bucketsToPublish) <> " windows at once - earlier ticks were missed"
+      -- Release the lock if this tick dies partway, so the next one can retry.
+      res <- try @_ @SomeException $ do
+        providers <- CQM.loadAllProviders
```
piyushKumar-1: these can be fetched at bootup and send as args to this thread instead of fetching here on every tick

piyushKumar-1: same for cities query below

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/DriverInformationExtra.hs
```diff
@@ -528,6 +532,33 @@ updateOnlineDurationRefreshedAt (Id driverId) onlineDurationRefreshedAt = do
...
+-- idx_driver_information_city_active (ddl-migration 0855).
+-- Assumes driver_information stays KV-DISABLED (Postgres-direct writes, per
+-- kv_configs) — if it is ever KV-enabled this count becomes drainer-stale.
+-- Legacy rows with NULL merchant_operating_city_id (pre-city-column) are
+-- deliberately excluded, matching the plain-equality convention of most
+-- per-city queries in this file. Errors PROPAGATE by design: the metrics
+-- publisher catches per-tick and holds the last good gauge value — coercing
+-- to 0 here would fake a supply collapse during a replica blip.
+countOnlineByCity :: (MonadFlow m, EsqDBFlow m r, CacheFlow m r) => Id DMOC.MerchantOperatingCity -> m Int
+countOnlineByCity merchantOpCityId = do
```
piyushKumar-1: lets not do this on DB 😅 , CHK should be fine for it, and only if the key not found we should do this. (essentially once in a lifetime)

Backend/app/kafka-consumers/src/Processor/RideEvents/Handlers.hs
```diff
@@ -482,7 +483,15 @@ handleDriverCityMigration ev = withRideAndBooking ev $ \ride booking -> do
...
         -- --- operational tables: sync directly, no gating. Document tables are
         -- intentionally excluded -- nothing reads their city column. ---
+        -- A city move relocates an online driver between supply buckets without touching
+        -- `active`, so the counter needs an explicit move. Read before the write, and
+        -- keyed on driver_information's own city -- that is the bucket it was counted in.
+        mbDriverInfo <- QDriverInfo.findById (cast ride.driverId)
         QDriverInfo.updateMerchantIdAndCityIdByDriverId ride.driverId booking.providerId targetOpCityId
+        whenJust mbDriverInfo $ \di ->
+          when di.active $ do
+            DSC.recordDriverActiveChange di.merchantOperatingCityId True False
```
piyushKumar-1: this is available: driverPerson.merchantOperatingCityId, no need to fetch `mbDriverInfo <- QDriverInfo.findById (cast ride.driverId)`

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverOnboarding/VehicleRegistrationCertificate.hs
```diff
@@ -781,7 +782,11 @@ deactivateRC isTaxiBoothRequest transporterConfig rc driverId = do
   removeVehicle isTaxiBoothRequest driverId
   DAQuery.deactivateRCForDriver False driverId rc.id
   now <- getCurrentTime
+  -- Bypasses updateDriverModeAndFlowStatus, so the supply count is maintained here too.
+  mbDriverInfo <- DIQuery.findById (cast driverId)
   DIQuery.updateActivityWithDriverFlowStatus False (Just DCommon.OFFLINE) (Just DDFS.OFFLINE) Nothing (Just now) (cast driverId)
+  whenJust mbDriverInfo $ \driverInfo ->
+    DSC.recordDriverActiveChange driverInfo.merchantOperatingCityId driverInfo.active False
```
piyushKumar-1: same here, instead of fetching driverInfo for `merchantOperatingCityId`, pass it to this function `deactivateRC` from the caller, they would have it already

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Driver.hs
```diff
@@ -1447,7 +1449,12 @@ deleteDriver admin driverId = do
   unless (driver.merchantId == admin.merchantId || driver.role == SP.DRIVER) $ throwError Unauthorized
   -- this function uses tokens from db, so should be called before transaction
   Auth.clearDriverSession driverId
+  -- Deleting the row drops the driver out of the supply count without ever clearing
+  -- `active`, so the count is corrected here.
+  mbDriverInfo <- QDriverInformation.findById (cast driverId)
   QDriverInformation.deleteById (cast driverId)
+  whenJust mbDriverInfo $ \driverInfo ->
+    DSC.recordDriverActiveChange driverInfo.merchantOperatingCityId driverInfo.active False
```
piyushKumar-1: same here, driver would have it

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverSupplyMetrics.hs
```diff
@@ -0,0 +1,214 @@
...
+recordSupplyEvent :: (Redis.HedisFlow m r, MonadFlow m) => Text -> Id DMOC.MerchantOperatingCity -> [([Text], Id DP.Person)] -> m ()
+recordSupplyEvent kind cityId entries =
+  unless (null entries) $ do
+    now <- getCurrentTime
+    Redis.sAddExp (supplyKey kind cityId.getId (currentBucket now)) (map supplyMember entries) windowTtlSecs
+
+-- | The whole batch shares one search request, so its funnel labels are passed once;
+-- the tier varies per driver.
+recordDriversPinged :: (Redis.HedisFlow m r, MonadFlow m) => Id DMOC.MerchantOperatingCity -> (Text, Text, Text) -> [(Text, Id DP.Person)] -> m ()
+recordDriversPinged cityId (distanceBucket, poolingLogicV, poolingConfigV) entries =
```
piyushKumar-1: this pinged and recordDriverAccepted we can get from the other metrics already, not required here.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/DriverSupplyMetrics.hs
```diff
@@ -0,0 +1,214 @@
...
+-- tiers, buckets or pooling versions. Dimension values are metric labels, which never
+-- contain "|", and driver ids are UUIDs, so the last segment is unambiguous.
+supplyMember :: ([Text], Id DP.Person) -> Text
+supplyMember (dims, driverId) = T.intercalate "|" (dims <> [driverId.getId])
+
+recordSupplyEvent :: (Redis.HedisFlow m r, MonadFlow m) => Text -> Id DMOC.MerchantOperatingCity -> [([Text], Id DP.Person)] -> m ()
+recordSupplyEvent kind cityId entries =
+  unless (null entries) $ do
+    now <- getCurrentTime
+    Redis.sAddExp (supplyKey kind cityId.getId (currentBucket now)) (map supplyMember entries) windowTtlSecs
```
piyushKumar-1: think of a better data store/approach for Pinged, Accepted and onRide, something similar to online drivers as current one will store a lot of driverIds in one redis shard (can bloat one node if number of drivers get very high)

onRide -> this is simple, every rideStart incr a key, on ride cancel/complete decr. 
Pinged -> i think below is not required for that, as its only increasing metric so the current counter metric we have should be suffice.
Accepted -> same this is also only increasing metric, so current counter should be fine.

(so we can remove the pinged and accepted metric from here, just do the onRide and online)

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/App.hs
```diff
@@ -140,6 +140,7 @@ runDynamicOfferDriverApp' appCfg = do
         L.setOption KBT.Tables kvConfigs
         _ <- liftIO $ createCAC appCfg
         initCityMaps
+        fork "driver supply metrics publisher" DSM.runDriverSupplyMetricsPublisher
```
piyushKumar-1: this can be done one line below and allProviders can be sent as a param

PR #16463, opened 2026-08-26
Rider app/feat/whatsapp i18n db migration

sathikabegum-nammayatri:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * WhatsApp bot language names and native names now support configured translations.
  * Translation text supports database-managed overrides and dynamic placeholder substitution.
  * Gujarati, Hindi, Kannada, Tamil, and Telugu localization is now supported through the translation system.
  * Merchants can configure ride-type order and control which options appear directly before “More options.”

* **Bug Fixes**
  * Missing translations now return a clear, standardized error.
  * Translation lookups consistently prioritize city-specific and requested-language content before English fallback.
  * Ride-type buttons now correctly honor configured ordering and display limits.


kavya-shree-s (commented):
## PR Review — automated review pass

**CI**: all 7 checks green (incl. WhatsApp Bot Golden Suite) · **Commits**: 1 · **Risk**: HIGH

---

### CRITICAL

**[C-A] One missing translation row silently takes the whole bot down — for every user, in every language**
`WhatsappBot/Adapter/Translations.hs:103`, `WhatsappBot/Adapter/Env.hs:129,175,150-153`

`getTranslationsMap` eagerly builds **all 6 languages × 112 keys** on *every* inbound message and *every* tracker tick. `resolveField` now throws `WhatsappBotTranslationNotFound` instead of falling back, and `dispatchInbound` catches an env-build failure and **only logs it** — there is no sender yet, so the user gets no reply at all.

Net effect: a single missing *Telugu* row breaks the bot for *English* users too, silently, with only a log line. The old static fallback was the floor under this; deleting `Gu/Hi/Kn/Ta/Te.hs` removes it entirely. The seed data is complete today (verified below), so this isn't broken *now* — but the blast radius of any future copy mistake is total and invisible.

*Suggestion*: resolve lazily for the requested language only (plus `languageName`/`nativeLanguageName` for the chooser), and/or keep `En.hs` as a logged-loudly fallback rather than dying.

**[C-B] Deploy ordering is now load-bearing, and nothing enforces it**
`Backend/dev/feature-migrations/0050-*.sql`, `0051-*.sql`

If this code rolls out before 0050/0051 are applied, the bot goes fully silent per C-A. Note `wa_bot_moreRideTypes` exists **only in 0051** and is required by `buildLanguageStringsM` for all 6 languages including English — 0050 alone is not enough. Please state the migration-first requirement in the description / release notes.

**[C-C] A shared query was switched from KV to raw Postgres, affecting 4 unrelated rider-app flows**
`Storage/Queries/TranslationsExtra.hs:34-50`

`findOneWithKV` → `findOneWithDb` on all 3 fallback levels. This function is not bot-specific — via `CachedQueries`, its callers are:

- `Domain/Action/UI/Pass.hs:800-802, 866-868` (pass name / benefit / description)
- `SharedLogic/Offer.hs:530`
- `Domain/Action/UI/Quote.hs:204`
- `Domain/Action/UI/CancellationReasons.hs:59`

The table is KV-enabled: `enableKVPG ''TranslationsT ['id] [['language], ['messageKey]]` (`src-read-only/Storage/Beam/Translations.hs:32`). Any row written through the app's KV path that hasn't drained to Postgres will now read as **missing** in those four flows.

The diagnosis in the comment (SQL-seeded rows never populate the KV secondary index) is correct, but the fix is applied one layer too wide. *Suggestion*: add a bot-scoped `...MessageKeyLanguageFromDb` variant and leave the shared function on KV.

---

### IMPORTANT

**[I-A] An empty/mistyped `rideTypesOrder` silently books Flexi**
`WhatsappBot/Adapter/Env.hs` (`parseRideType`), `Flow/Booking.hs:94-100`

Unrecognized ids are dropped, so `["quickride"]` or `[]` yields `rideTypesOrder = []` → `flexiOffered = regularOffered = False` → `promptForBookingEntry` falls to `let rt = if regularOffered merchant then Regular else Flexi` = **Flexi**. The merchant offers nothing, the user gets booked into Flexi. The old code defaulted an unknown `rideMode` to `RideModeBoth` (chooser shown), so this is a regression on bad config. *Suggestion*: log and treat an empty offered-list as unavailable, or reject at `mkMerchantCtx`.

**[I-B] Two-ride-type merchants now need an extra tap to see the second option**
`Flow/Booking.hs` (`splitRideTypes`)

The rule is `min(len - 1, 2)` direct. With today's real config (`[Flexi, Regular]`) that is **1 direct button + "More options"** — but WhatsApp allows 3 reply buttons, so both would fit with no More at all. Same at 3 types (2 direct + More instead of 3 direct). The `-1` reservation should only apply when `len > 3`. This is locked in by the new unit tests, so it may be deliberate — worth confirming with product, because it costs every user an extra tap on the main booking path.

**[I-C] The chooser gate wasn't generalized along with the rest**
`Flow/Booking.hs:94` — `promptForBookingEntry` still gates on `flexiOffered merchant && regularOffered merchant`. With the new ordered list it should be `length merchant.rideTypesOrder > 1`. Equivalent today, but the stated motivation for this PR is room for a 3rd ride type, which would hit this line immediately.

**[I-D] Per-message translation cost went from 6 lookups to 672**
`WhatsappBot/Adapter/Translations.hs:103`

Cached at two layers (in-mem 1h + Hedis), so steady state is fine. But on a cold pod or after the in-mem TTL, one inbound message issues up to 672 Redis GETs and, on miss, up to 3 Postgres queries each. Also `CachedQueries.Translations` only caches on `whenJust result` — a **negative** lookup is never cached, so a permanently-missing key re-runs 3 Postgres queries on every request. *Suggestion*: one `findAllBy` per (city, language) assembled in memory, or cache the whole assembled map.

**[I-E] The `more_ride_types` handler is unguarded**
`WhatsappBot/Engine.hs:140` — no state check and no `save`, unlike `pickup_confirm` on the very next line. A stale button tap later in the conversation re-shows ride-type buttons out of context.

**[I-F] `MetaBotCfg` jsonb field renamed with no data migration**
`Domain/Types/Extra/MetaWebhookConfig.hs:33` — `rideMode :: Text` → `rideTypesOrder :: [Text]`; both required, no `Maybe`, no default. Any existing `meta_config.bot_config` row will fail to decode. The in-file comment asserts there are no production rows for this table — if that still holds this is fine, but dev/QA/UAT rows will break and no `UPDATE` is shipped. Please confirm in the description.

---

### Good to have

- **[G1]** `substitute` (`I18n/Build.hs:31`) rescans already-substituted text: `foldl'` replaces `{{0}}` then `{{1}}`, so an argument whose value contains a literal `{{1}}` gets substituted itself. Pre-existing, but it now backs 112 fields instead of 6 — several interpolating error text (`setupFailed`, `cancelFailed`, `error`). A single-pass replace closes it.
- **[G2]** `test/resources/i18n/non-english.json` is a frozen export with nothing checking it against the SQL. I diffed them: **all 560 values match the migrations exactly** today — but the next copy edit in SQL will pass CI unnoticed. A small CI comparison would keep them honest.
- **[G3]** The PR description is an untouched template — no description, no motivation, no test evidence, every checkbox unticked (including "modifies the database schema"), for a change that deletes 5 modules and makes all bot copy a hard DB dependency.
- **[G4]** `"➕ More options"` is duplicated between `I18n/En.hs:118` and `0051-*.sql`.

---

### Verified clean

Seed completeness: **112 keys required** by `buildLanguageStringsM`, **112 seeded** across 0047+0050+0051, **672 rows = 112 × 6 languages, zero gaps, zero duplicate `(key, language)` pairs**. Both `ON CONFLICT (message_key, language) WHERE merchant_operating_city_id IS NULL` clauses match the existing partial unique index `uq_translations_global`, and `uq_translations_city` covers the city-scoped lookup — no new index needed. No Redis pattern matching, no partial functions outside test code, no TODO/FIXME, no PII fields, no API spec or `src-read-only/` changes.

The shared `buildLanguageStringsM` with a full explicit record construction (so `-Wmissing-fields` under `-Werror` proves nothing was skipped) is a genuinely good way to stop prod and tests drifting.

---

### QA should test

1. Full booking flow in each of Gujarati / Tamil / Telugu (previously unseeded for the 6 pilot fields — this is the newly fixed path).
2. Language chooser → "More languages": names now come from the DB, not `En.hs`.
3. A merchant offering both Flexi and Regular: confirm 1 direct button + "More options" is acceptable (see I-B), and that "More options" reveals the second type.
4. Deploy 0050+0051 **before** the code, then verify the bot replies; separately confirm behaviour when a row is deleted (expect: total silence — see C-A).
5. Regression pass on Pass details, Offers, Quote, and cancellation-reason copy — all four moved off the KV read path (C-C).

---

Well-structured refactor with genuinely complete and verified seed data. Three things worth resolving before merge: the shared `findOneWithKV` → `findOneWithDb` switch leaking into four unrelated flows (C-C), an empty `rideTypesOrder` silently booking Flexi (I-A), and removing every fallback meaning one bad row silently kills the bot for all languages at once (C-A).

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

kavya-shree-s:
@sathikabegum-nammayatri Please resolve conflicts

kavya-shree-s (approved):

PR #16495, opened 2026-08-28
backend/feat: add lean_flow system_configs toggle for heavy non-core …

Yashwanth1906:
…features

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added centralized controls for enabling or excluding driver and rider features, including dynamic pricing, demand hotspots, leaderboards, driver coins, route suggestions, and scheduled processing.
  * Feature settings are cached for responsive access and can be refreshed when configuration changes.

* **Bug Fixes**
  * Prevented duplicate geohash-area labels for the same city and location.
  * Excluded features now stop related updates and background processing while scheduled jobs continue safely.


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/SystemConfigs/LeanFlow.hs
```diff
@@ -0,0 +1,57 @@
...
+  let isStale = maybe True (\lastUpdated -> diffUTCTime now lastUpdated > leanFlowCacheTtl) mbLastUpdated
+  if isStale
+    then do
+      mbConfig <- decodeFromText' @LeanFlowConfig <$> KSQS.findById leanFlowConfigId
+      case mbConfig of
+        Just config -> do
+          L.setOption LeanFlowOption config
+          L.setOption LeanFlowLastUpdatedTime now
+          pure (Just config)
+        Nothing -> L.getOption LeanFlowOption
```
sidsethu: **[CRITICAL] The cache never memoizes a miss, so a missing or malformed config turns every feature check into a Postgres round-trip.**

The `Nothing` branch returns the stale option without ever setting `LeanFlowLastUpdatedTime`. `isStale` therefore stays `True` forever and **every subsequent call re-queries the DB**.

That is expensive here: `KSQS.findById` uses `findOneWithDb`, which bypasses KV and hits Postgres directly, and on a miss it also increments the `system_configs_find_failed_*` metric — so you get alert noise on top of the load.

Blast radius, because of where the checks now sit:
- `SharedLogic/FarePolicy.hs` — `getFullFarePolicy` is invoked via `mapConcurrently` over every resolved fare product, plus once for the base variant, plus once in `getAllFarePoliciesProduct`: roughly **N+2 concurrent Postgres reads per search** on the driver app's hottest path.
- `kafka-consumers` — **9 reads per `RideEndedEvent`**.
- Also per driver quote (`buildDriverQuote`) and per search try (`buildSearchTry`).

Two realistic triggers: the `lean_flow` row not existing yet (any window between deploying this code and ops applying the migration), and any JSON that fails to decode — which is permanent, not transient.

Suggested fix — cache the negative result for the same TTL:

```haskell
        Nothing -> do
          L.setOption LeanFlowLastUpdatedTime now
          L.getOption LeanFlowOption
```

Same applies to the identical rider-app copy of this file.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Types/Extra/LeanFlow.hs
```diff
@@ -0,0 +1,43 @@
...
+  | DYNAMIC_PRICING
+  deriving stock (Eq, Show, Read, Generic)
+  deriving anyclass (ToJSON, FromJSON)
+
+data LeanFlowConfig = LeanFlowConfig
+  { enabled :: Bool,
+    featuresExcluded :: [LeanFlowFeature]
+  }
+  deriving stock (Eq, Show, Generic)
+  deriving anyclass (ToJSON, FromJSON)
```
sidsethu: **[CRITICAL] Strict decoding means one bad string silently disables the entire kill-switch.**

`LeanFlowConfig` derives generic `FromJSON` with a required `enabled :: Bool` and a required `featuresExcluded :: [LeanFlowFeature]`. The whole document fails to decode if:

- someone writes `{"enabled": true}` and omits `featuresExcluded`, or
- `featuresExcluded` contains a feature name this deployed binary doesn't know — a typo, or a feature added in a newer build that's mid-rolling-deploy.

The result is `Nothing`, which means **every flag reverts to "not excluded"** and simultaneously triggers the unbounded DB reads flagged in `Storage/CachedQueries/SystemConfigs/LeanFlow.hs`.

This is the worst failure mode for a mechanism whose whole purpose is to be reached for during an incident: it fails silently, it fails globally, and it fails toward *more* load rather than less.

Suggestion: make `featuresExcluded` lenient — e.g. `Maybe [Text]` resolved against the enum with unknown names logged and skipped, or at minimum add a catch-all constructor and a `logError` when decoding fails so the failure is visible.

(To be clear, the fail-open default in `isFeatureExcluded` — `maybe False` — is the right call. An unreadable config *should not* disable features. It's the all-or-nothing parse that's the problem.)

Backend/dev/ddl-migrations/rider-app/1565-add-lean-flow-config.sql
sidsethu, line 1: **[CRITICAL] Migration index collides with `main` and will trip the repo's own pre-commit hook.**

`origin/main` already has `1564-add-is-external-order-to-payment-order.sql`, and on the driver side `origin/main` already has **two** `0861-*` files (`0861-add-is-external-order-to-payment-order.sql`, `0861-geohash-area-unique-index.sql`) — so `0861-add-lean-flow-config.sql` would make three.

The `overlapping-migrations` hook in `Backend/nix/pre-commit.nix` fails when the count of unique 4-character prefixes doesn't equal the file count in a service directory, so both of these need renumbering. The PR is currently in a `CONFLICTING` state anyway, so this can be folded into the rebase.

For the record, I checked whether the `INSERT` violates the `ddl-migrations-no-dml` hook — **it does not**. That hook explicitly allowlists DML targeting `system_configs`, so the file is in the right directory.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/SystemConfigs/LeanFlow.hs
```diff
@@ -0,0 +1,57 @@
...
+import Kernel.Prelude
+import qualified Kernel.Storage.Queries.SystemConfigs as KSQS
+import Kernel.Utils.Common
+import Storage.Beam.SystemConfigs ()
+
+data LeanFlowOption = LeanFlowOption
+  deriving stock (Generic, Typeable, Show, Eq)
+  deriving anyclass (ToJSON, FromJSON)
+
+instance OptionEntity LeanFlowOption LeanFlowConfig
```
sidsethu: **[IMPORTANT] The rider and driver `LeanFlowOption` types produce an identical EulerHS cache key.**

EulerHS derives the option key as:

```haskell
mkOptionKey :: forall k v. OptionEntity k v => k -> Text
mkOptionKey k = show (typeRep @k) <> decodeUtf8 (BSL.toStrict $ encode k)
```

`Type.Reflection`'s `Show` prints the **unqualified** type name, so both apps' options key on `"LeanFlowOption"` — the module path doesn't disambiguate them. `GetOption` then does `unsafeCoerce` on whatever it finds in the map.

Today this is latent, because only the driver-app module is actually invoked in any single process. But `kafka-consumers` already links **both** `rider-app` and `dynamic-offer-driver-app` (see `kafka-consumers.cabal`, and the same is true of `provider-dashboard`, `lib-dashboard`, and `producer`). The moment any process calls both, a rider `LeanFlowConfig` gets coerced into a driver one and the excluded-feature list is reinterpreted by constructor index — wrong flags, silently, with no error.

Cheap fix: give the two option types distinct names, e.g. `RiderLeanFlowOption` / `DriverLeanFlowOption` (and likewise for `LeanFlowLastUpdatedTime`).

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Scheduler/Jobs/Chakras.hs
```diff
@@ -58,65 +60,82 @@ runDailyJob ::
...
+    Event.runKaalChakraUpdateTagsJob kaalChakraHandle LYT.Monthly jobInfo.jobData
+
+rescheduleIfChakraExcluded :: ChakraJobs m r => LYT.Chakra -> Maybe UTCTime -> m ExecutionResult -> m ExecutionResult
+rescheduleIfChakraExcluded chakra startTime action = do
+  excluded <- CQLF.isFeatureExcluded NAMMA_TAG_CHAKRA
+  if excluded
+    then do
+      now <- getCurrentTime
+      pure $ ReSchedule $ Event.getNextChakraTime startTime now chakra
+    else action
```
sidsethu: **[IMPORTANT] Flipping the Chakra flag mid-run leaves batch state dirty, and excluding an UpdateTag job breaks the job chain.**

The normal cadence isn't a self-rescheduling job — it's a chain (`Lib/Yudhishthira/Event/KaalChakra/Jobs.hs`): the fetch job `Complete`s and chains an UpdateTag job, which `Complete`s and chains the *next* fetch job. Returning `ReSchedule` here departs from that in two ways:

1. **Dirty batch state.** If a fetch job is mid-batch (`LYT.Continue`) when the flag goes on, this path returns `ReSchedule` without calling `clearEventData` / `decrChakraBatchNumber`. The partial batch-number key survives in Redis, so on re-enable the job resumes from a stale batch number and may skip users.

2. **Broken chain for UpdateTag jobs.** When an UpdateTag job is excluded, `createFetchUserDataJob` never runs, so the next cycle's fetch job is never created. The lingering UpdateTag job later runs against a stale `eventId` whose event data has since been cleared/expired, and only self-heals via its `Failed` branch.

For the UpdateTag variants specifically, mirroring the normal `Completed` path would be cleaner — chain the next fetch job and return `Complete` rather than keeping a stale UpdateTag job alive.

Minor, related: `jobData.startTime` is fixed and never advances under exclusion, so once `startTime + chakraDays` is in the past, `getNextChakraTime`'s `max`/`nextDiffTime` clamp lands within the next 24h. A **Monthly or Quarterly chakra will wake up daily** while excluded. Each wake is cheap, so impact is low, but it isn't what the code reads like it does.

Same comments apply to the driver-app copy in `SharedLogic/KaalChakra/Chakras.hs`.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/FarePolicy.hs
```diff
@@ -274,7 +281,12 @@ getFareSettlementTypeForSpecialZone (Just specialZoneId) = do
...
 getFullFarePolicy mbFromLocation mbToLocation mbFromLocGeohash mbToLocGeohash mbDistance mbDuration txnId mbBookingStartTime mbBaseVaraintCarPrice mbAppDynamicLogicVersion mbSpecialLocName mbSpecialZoneId fareProduct _configsInExperimentVersions dpInputsList mbTransporterConfig mbPreResolvedVSTItem = do
-  transporterConfig <- maybe (getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = fareProduct.merchantOperatingCityId.getId}) Nothing >>= fromMaybeM (TransporterConfigNotFound fareProduct.merchantOperatingCityId.getId)) pure mbTransporterConfig
+  rawTransporterConfig <- maybe (getOneConfig (TransporterConfigDimensions {merchantOperatingCityId = fareProduct.merchantOperatingCityId.getId}) Nothing >>= fromMaybeM (TransporterConfigNotFound fareProduct.merchantOperatingCityId.getId)) pure mbTransporterConfig
+  dynamicPricingExcluded <- CQLF.isFeatureExcluded DYNAMIC_PRICING
+  let transporterConfig =
+        if dynamicPricingExcluded
+          then rawTransporterConfig {isDynamicPricingQARCalEnabled = Just False, isMLBasedDynamicPricingEnabled = False}
+          else rawTransporterConfig
```
sidsethu: **[IMPORTANT] `DYNAMIC_PRICING` exclusion is redundant here and incomplete overall.**

**Redundant:** `getAllFarePoliciesProduct` (line ~205) already builds the overridden `transporterConfig` and passes it down as `Just transporterConfig`. This function then re-checks the flag and re-applies the identical override on every call. Since `getFullFarePolicy` runs via `mapConcurrently` over every fare product, that's the per-search fan-out driving the DB issue flagged in `Storage/CachedQueries/SystemConfigs/LeanFlow.hs`. Pass the already-adjusted config through instead of re-deriving it.

**Incomplete:** forcing `isDynamicPricingQARCalEnabled`/`isMLBasedDynamicPricingEnabled` off doesn't cover everything — congestion and supply-demand multipliers are read from Redis and still get applied.

Which surfaces a cross-feature interaction worth calling out explicitly in the PR description / runbook: excluding `CONGESTION_CHARGE` or `SUPPLY_DEMAND` on their own stops the jobs that *refresh* those Redis keys. Once the existing TTLs expire, the pricing inputs silently disappear and fares change. **Excluding a "non-core" background job can move real fares** — that needs to be understood before anyone flips these in prod.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/UI/Search.hs
```diff
@@ -712,8 +714,9 @@ search personId req bundleVersion clientVersion clientConfigVersion_ mbRnVersion
     updateRideSearchHotSpot :: SearchRequestFlow m r => DPerson.Person -> SearchReqLocation -> Merchant -> Maybe Bool -> Maybe Bool -> m ()
     updateRideSearchHotSpot person origin merchant isSourceManuallyMoved isSpecialLocation = do
       fork "ride search geohash frequencyUpdater" $ do
+        hotSpotExcluded <- CQLF.isFeatureExcluded HOTSPOT
         HotSpotConfig {..} <- getOneConfig (HotSpotConfigDimensions {merchantOperatingCityId = "", merchantId = merchant.id.getId}) Nothing >>= fromMaybeM (InternalError "config not found for merchant")
-        when (shouldSaveSearchHotSpot && shouldTakeHotSpot) do
+        when (shouldSaveSearchHotSpot && shouldTakeHotSpot && not hotSpotExcluded) do
```
sidsethu: **[IMPORTANT] Config lookup still runs before the exclusion guard** (agreeing with CodeRabbit's still-open point here).

`getOneConfig` executes on line 718 regardless of `hotSpotExcluded`, and can throw `InternalError "config not found for merchant"` even when `HOTSPOT` is excluded. It's inside a `fork`, so there's no customer-facing impact — but it defeats the purpose of a lean-mode flag, which is to *skip* the work.

```haskell
        hotSpotExcluded <- CQLF.isFeatureExcluded HOTSPOT
        unless hotSpotExcluded $ do
          HotSpotConfig {..} <- getOneConfig (...) >>= fromMaybeM (InternalError "config not found for merchant")
          when (shouldSaveSearchHotSpot && shouldTakeHotSpot) do
            ...
```

The same shape appears in `Domain/Action/UI/LeaderBoard.hs` (lines 77, 102, 117): `getOneConfig` runs and can throw `InternalError "Leaderboard configs not present"` before the exclusion check is consulted. Worth moving those lookups under the guard too — though note the leaderboard ones are on a request path, so an excluded-but-misconfigured city returns a 500 rather than the intended "Leaderboard Not Available".

Backend/app/kafka-consumers/src/Processor/RideEvents/Processor.hs
```diff
@@ -27,26 +29,28 @@ processRideEnded event =
...
-        Right () -> pure ()
-        -- Re-throw so the entry stays pending and is retried by the transport,
-        -- exactly as before. withTryCatch has already recorded the per-handler
-        -- failure in try_exception_error_counter{ErrorContext="rs-handler:<name>"}.
-        Left e -> throwM e
+runHandler :: LeanFlowFeature -> Text -> RideEndedEvent -> (RideEndedEvent -> Flow ()) -> Flow ()
+runHandler feature name event handler =
+  withLogTag name $ do
+    excluded <- CQLF.isFeatureExcluded feature
+    unless excluded $
```
sidsethu: **[IMPORTANT] Excluded events are consumed and permanently lost — please document this.**

The exclusion check runs *before* `Idem.withIdempotency`, so an excluded event is acked and skipped with no idempotency key written and no replay path. Turning `ANALYTICS_KAFKA` off for an hour permanently loses that hour of events; re-enabling does not backfill.

That may well be the intended trade for analytics/leaderboard, but it should be stated explicitly in the PR description so nobody reaches for these flags assuming the data catches up afterwards. Worth being deliberate per feature — `FLEET_OPERATOR_STATS` and `RC_STATS_REMINDERS` are counters where a permanent gap may matter more than for `ANALYTICS_KAFKA`.

Also note this is the site with the largest per-event fan-out of the config read flagged in `Storage/CachedQueries/SystemConfigs/LeanFlow.hs` — **9 `isFeatureExcluded` calls per `RideEndedEvent`**, each a Postgres round-trip whenever the config is missing or unparseable.

Backend/app/rider-platform/rider-app/Main/src/Storage/CachedQueries/SystemConfigs/LeanFlow.hs
sidsethu, line 1: **[IMPORTANT] This file is byte-identical to the driver-app copy** — git reports the same blob hash (`a98fa8e7d9e`) for both. The two `Domain/Types/Extra/LeanFlow.hs` files differ only in the feature enum.

Worth extracting into a shared lib with the feature type as a parameter, so the caching logic (and any fix to the miss-caching and lenient-decoding issues raised elsewhere in this review) only has to be made once. As it stands, every one of those fixes has to be applied twice and can drift.

Note also that every finding filed against the driver-app copy applies verbatim here.

Minor while you're in here: the driver-app `Domain/Types/Extra/LeanFlow.hs` carries the AGPL header and the rider-app one doesn't.

sidsethu (commented):
## Review: lean_flow feature toggles

The design is sound and the default-off posture is right. Inline comments cover 3 critical and 6 important findings; summary below.

### First, on CI

The green checks here are close to meaningless: only HLint ×2, a Dashboard repl check, `check-conditions`, and CodeRabbit ran. **There is no `cabal build` check on this PR**, so compilation of rider-app, dynamic-offer-driver-app, and kafka-consumers is unverified. I hand-checked the type-level assumptions (`KaalChakraJobData`/`UpdateKaalBasedTagsData` both do have `startTime :: Maybe UTCTime`, and `ChakraJobs` does imply `EsqDBFlow` + `CacheFlow`), so it should compile — but please build locally before merging. The PR is also currently `CONFLICTING` and needs a rebase.

### Critical (3)

1. **The cache never memoizes a miss** — the `Nothing` branch in `findLeanFlowConfig` doesn't set `LeanFlowLastUpdatedTime`, so a missing or unparseable config makes *every* feature check a Postgres round-trip (`findOneWithDb` bypasses KV), plus `system_configs_find_failed_*` metric noise. That's ~N+2 concurrent reads per search via `getFullFarePolicy`, and 9 per `RideEndedEvent` in kafka.
2. **Strict decoding makes the kill-switch brittle** — a missing `featuresExcluded` key or one unknown feature name fails the whole parse, silently reverting every flag to "not excluded" *and* triggering #1.
3. **Both migration indices collide with `main`** (`1564`, and `0861` which already has two files there) — the `overlapping-migrations` pre-commit hook will fail.

#1 and #2 compound into the scenario you least want: a malformed config during an incident that both disables the kill-switch and adds N Postgres reads per search. Worth fixing together.

### Important (6)

Rider and driver `LeanFlowOption` share an identical EulerHS key (`unsafeCoerce` hazard — `kafka-consumers` already links both packages); Chakra exclusion leaves batch state dirty mid-run and breaks the UpdateTag→fetch chain; `DYNAMIC_PRICING` is re-derived redundantly per fare product and doesn't cover Redis-cached pricing inputs; hotspot/leaderboard config lookups still run before their guards; excluded kafka events are permanently lost with no backfill; and the two cached-query modules are byte-identical duplicates.

### One product question, not a code issue

**The flag is global — there's no per-merchant or per-city scoping.** Excluding `LEADERBOARD` kills it for every merchant and city in the deployment. That may be exactly right for a lean-stack deployment, but every other config in this codebase is city-scoped, so it's worth confirming this is a deliberate product decision rather than an implementation shortcut.

Two smaller things worth a look: `clearCache` is exported but never called anywhere (and works by writing a timestamp ~31 years in the past), and the `enabled` field is a double negative — `enabled: true` means features are *disabled*, which is easy to read backwards at 3am.

### Suggested QA

- A search with multiple fare products while the `lean_flow` row is **absent** — watch Postgres query count and the `system_configs_find_failed_*` metric.
- A search with `featuresExcluded` containing an unrecognised string.
- End-ride coins with `DRIVER_COINS` excluded, then re-enabled. Note `blockDriverCoinsForToday` is skipped while excluded, so a driver who should have been day-blocked won't be on re-enable.
- Toggling `NAMMA_TAG_CHAKRA` on while a chakra batch is mid-flight, then off again.

Nice work on the fail-open default in `isFeatureExcluded` — an unreadable config not disabling features is the right instinct.

Backend/dev/ddl-migrations/dynamic-offer-driver-app/0865-geohash-area-unique-index.sql
sidsethu, line 1: **[CRITICAL] This is a rename of a migration already merged to `main`, not a new file — that's a different and worse problem than the numbering collision it fixes.**

`0861-geohash-area-unique-index.sql` already exists on `main` (merged in `9e4dc269eb`). This diff renames it to `0865-...` purely to free up the `0861` prefix for the new `lean-flow-config` migration — but that migration should have been renumbered instead, not this one.

I checked how these are tracked: services boot through `Kernel.Storage.Esqueleto.Migration.migrateIfNeeded` (mobility-core), which wraps `postgresql-simple-migration` — it records one row **per filename** in `<schema>.schema_migrations` and skips whatever's already there; it does not compare file content. Any long-lived environment with `autoMigrate=true` that already applied the old filename will see `0865-geohash-area-unique-index.sql` as a brand-new migration and try to run it again.

That would fail hard: the SQL is a bare `ALTER TABLE ... ADD CONSTRAINT ... UNIQUE (...)` with no `IF NOT EXISTS` guard, so re-running it errors on "constraint already exists" rather than no-op'ing.

Fix: revert this rename and bump the *new* lean-flow-config migration's own number instead (e.g. keep it at a free index like `0866`), leaving every already-merged filename untouched. The same logic applies to the rider-app side if any pre-existing file was touched there — from the diff it looks like only the driver-app `0861` file was renamed, but worth double-checking.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide.hs
```diff
@@ -664,96 +664,100 @@ endRideHandler handle@ServiceHandle {..} rideId req = do
...
-              expirationPeriod
-              thresholdConfig.timeDiffFromUtc
-              timeBoundReferenceUtc
-            -- Journey# drivers skip legacy/incentive coin awards on OTP rides too.
-            unless (SLJourney.hasJourneyTag driverTag) $
-              DC.driverCoinsEvent driverId mbDriver booking.providerId booking.merchantOperatingCityId (DCT.EndRide (isJust booking.disabilityTag) (booking.coinsRewardedOnGoldTierRide) updRide metroRideType DCT.OTPRideTrip) (Just ride.id.getId) ride.vehicleVariant (Just booking.vehicleServiceTier) (Just booking.configInExperimentVersions) booking.area
+      driverCoinsExcluded <- CQLF.isFeatureExcluded DRIVER_COINS
+      unless driverCoinsExcluded $ do
+        when shouldFlagRiderForRepeatCustomerFraud $ QRiderDetails.flagRiderForCoinZero booking.riderId
+        QRide.incrementDriverRiderRideCountForDay (cast driverId) booking.riderId
```
sidsethu, line 670: **[IMPORTANT] Gating the write but not the read splits a counter in half, and the result leaks into `rideTags` regardless of `DRIVER_COINS`.**

`incrementDriverRiderRideCountForDay` (line 670, the write) is now skipped whenever `DRIVER_COINS` is excluded. But the read it feeds — `QRide.countPriorCompletedRidesWithSameCustomer` at line 641 — runs unconditionally, every ride end, regardless of the flag. That read produces `priorRidesSameCustomer`, which computes `shouldBlockCoinsForSameRiderFlow` (line 643), which is passed into `Y.EndRideTagData` and baked into `updRide.rideTags` — unconditionally, outside this fork entirely.

Concretely: once `DRIVER_COINS` is excluded, the "rides with this same rider" counter stops incrementing but is still read every ride. So:
- While excluded, the count is frozen at its pre-exclusion value — `shouldBlockCoinsForSameRiderFlow` can never newly trip for rides that happen during the exclusion window (false negatives).
- For `slidingWindow7Days` after DRIVER_COINS is re-enabled, the counter has a real gap (missing days read as zero activity), so the same undercounting continues for a week after re-enabling.

This isn't confined to coins — `rideTags` is consumed well beyond this file (`SharedLogic/Ride.hs`, `Domain/Action/Dashboard/Ride.hs`, `Storage/Queries/RideExtra.hs`, `Storage/Clickhouse/BppTransactionJoin.hs`). Toggling a "lean flow" flag meant to shed non-core background load ends up quietly degrading a tag that other, unrelated logic reads.

Given `flagRiderForCoinZero` and `incrementDriverRiderRideCountForDay` are indeed coin-specific bookkeeping (confirmed both are only called from this one call site), gating them behind `DRIVER_COINS` is the right call — but the read at line 641 needs to move under the same guard (or be split into a coins-only sub-computation) so the counter's write and read halves are gated together instead of only the write.

sidsethu (commented):
## Re-review of the update

Good round of fixes. Confirmed resolved from the last pass:

- ✅ **Cache never memoized a miss** — `LastUpdatedTime` is now set unconditionally on every stale refresh, so a missing/unparseable config now costs one DB read per 60s instead of one per call.
- ✅ **Strict decoding** — the custom `FromJSON` instance now defaults `enabled`/`featuresExcluded` and drops unrecognized feature strings individually instead of failing the whole parse.
- ✅ **EulerHS option-key collision** — `LeanFlowOption` renamed to `DriverLeanFlowOption`/`RiderLeanFlowOption` (and the paired `LastUpdatedTime` types), so the shared-process coercion hazard is gone.
- ✅ **Config lookups running ahead of their exclusion guards** — both `Search.hs` (hotspot) and `LeaderBoard.hs` (all three cadences) now check the flag before calling `getOneConfig`.
- ✅ **Redundant `DYNAMIC_PRICING` re-derivation in `getFullFarePolicy`** — now reuses the caller's already-adjusted config via the `Just alreadyResolved` branch instead of re-fetching and re-checking the flag per fare product.
- ✅ **Chakra batch-state corruption** — resolved by removing `NAMMA_TAG_CHAKRA` gating entirely from both rider and driver Chakra jobs rather than trying to patch the reschedule path.
- ✅ **Kafka permanent event loss** — scope narrowed a lot: only `LEADERBOARD`, `REFERRAL`, and `FLEET_OPERATOR_STATS` are still gated via the new `runHandler`/`runHandlerAlways` split; the rest (analytics, interpolation, GPS toll, RC stats, notifications, city migration) now always run. The residual loss-with-no-backfill for those three is worth a one-line callout in the description, but it's a much smaller surface than before.

### New in this update (2 findings, inline)

1. **[Critical]** The migration-numbering fix renamed an *already-merged-to-main* file (`0861-geohash-area-unique-index.sql` → `0865-...`) instead of renumbering only the new lean-flow migration. I checked how these are tracked: `postgresql-simple-migration` keys applied migrations by filename in `schema_migrations`, and the underlying `ALTER TABLE ... ADD CONSTRAINT` isn't idempotent (no `IF NOT EXISTS`). Any long-lived `autoMigrate=true` environment that already ran the old filename will re-run this and hard-fail on "constraint already exists." Please revert that rename and bump the new migration's own number instead.
2. **[Important]** In `EndRide.hs`, the `DRIVER_COINS` guard now wraps the *write* side of the same-rider ride counter (`incrementDriverRiderRideCountForDay`) but not the *read* side (`countPriorCompletedRidesWithSameCustomer`, still unconditional). That read feeds `shouldBlockCoinsForSameRiderFlow`, which is baked into `rideTags` unconditionally — a field consumed well outside the coins feature. Net effect: excluding `DRIVER_COINS` silently freezes that counter while still reading it, producing stale fraud-tag data during the exclusion window and for the following 7-day sliding window after re-enabling.

### Still worth a look (not blocking)

- `FarePolicy.hs`: excluding `DYNAMIC_PRICING` still doesn't touch the Redis-cached congestion/supply-demand multipliers that `CONGESTION_CHARGE`/`SUPPLY_DEMAND` exclusion would stop refreshing — worth a line in the runbook so nobody's surprised fares move when a "non-core" job is excluded.
- `LeaderBoard.hs`: the three "Not Available" messages are now worded slightly differently per cadence, and the monthly one has a double space (`"Leaderboard Not Available  Now"`) — minor, cosmetic.
- The lenient `FromJSON` silently drops unrecognized `featuresExcluded` entries with no log line — fine for fail-safe behavior, but a typo'd feature name in the config will be invisible.

Nice work tightening the cache and decode behavior — that combination was the scariest part of the original submission and it's solid now.

PR #16506, opened 2026-08-29
chore-adding-isStruck-Boolean

Yashwanth1906:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added internal administration endpoints for managing rides and bookings, including synchronization, cancellation, and ending rides.
  - Added booking status tracking to identify bookings requiring follow-up.
  - Added batch updates when bookings are synchronized or cancelled.

- **Improvements**
  - Enhanced auditing for administrative actions.
  - Added new bus-approaching notification thresholds and updated stop-crossed messaging.


Backend/app/rider-platform/rider-app/Main/spec/Storage/Booking.yaml
```diff
@@ -204,6 +204,7 @@ Booking:
     hasStops: Maybe Bool
     riderId: Id Person
     isPetRide: Bool
+    isStucked: Bool  # ops marker for the stuck-booking cron; set in batch by dashboard booking sync / stuck-cancel
```
sidsethu, line 207: **Blocker: the generated migration for `is_stucked` is missing.**

The field was added to both Booking specs, but `Backend/dev/migrations-read-only/{rider-app,dynamic-offer-driver-app}/booking.sql` isn't in this PR — the DDL was hand-written into the legacy `ddl-migrations/` tree instead. Note that `migrations-read-only/*/ride.sql` *is* regenerated here, so the generator ran but `booking.sql` didn't get committed.

Every other spec-driven Booking column (`is_auto_accepted`, `preference_match_score`, `payment_charge`, ...) lands in `migrations-read-only/<app>/booking.sql`. Two consequences:

1. The generator-diff CI job regenerates from the spec and will report `booking.sql` dirty.
2. Any environment provisioned from `migrations-read-only` gets no `is_stucked` column, while the regenerated `Storage/Beam/Booking.hs` now SELECTs it -> `column booking.is_stucked does not exist` on **every** booking read in both rider-app and driver-app, not just the new flows.

Please re-run the generator and commit `booking.sql` for both apps.

Backend/app/dashboard/provider-dashboard/src/API/ProviderPlatform/DynamicOfferDriver/InternalAdmin.hs
```diff
@@ -0,0 +1,157 @@
...
+-- EVERY merchant on the instance; the target merchant comes from the
+-- @merchantId@/@city@ path segments. The operation bodies themselves are the
+-- exact same implementations the RBAC-guarded routes use
+-- (@Domain.Action.ProviderPlatform.Management.*@).
+--
+-- Mounted under @\/bpp\/driver-offer\/internal\/admin\/...@, next to the other
+-- internal services. NEVER expose this through a public ingress — it is meant
+-- for cluster-internal automation (stuck-booking cron jobs, ops tooling)
+-- reached over the dashboard's in-cluster service DNS.
+type API =
```
sidsethu, line 38: **Blocker: this widens the blast radius of a key that is already distributed.**

`internalAuthAPIKey` previously only gated the read-only `internal/auth` fleet-owner lookup, and it has already been handed to an out-of-repo caller. This route tree now makes that same key authorize mass ride-end, ride-cancel and cancel-all-stuck for **every merchant on the instance**:

- `checkMerchant` only asserts the merchant exists and serves `DRIVER_OFFER_BPP_MANAGEMENT` — the token is not bound to any merchant.
- `POST .../internal/admin/<any merchant>/<any city>/booking/cancel/allStuck` with `bookingIds: []` reaches `postBookingCancelAllStuck`, whose `findStuckBookings merchant merchantOpCity [] now` returns **all** stuck bookings and cancels them merchant-wide.
- The dev value is the hardcoded `ae288466-2add-11ee-be56-0242ac120002`, shared verbatim across the provider/rider/safety dhall.
- These routes are also emitted into the app's own `/openapi` and `/swagger`, so they are discoverable from the same host.

The header comment says "NEVER expose this through a public ingress", but nothing in the code enforces it. At minimum this needs its own key rather than reusing `internalAuthAPIKey`; ideally also per-merchant scoping and a rate limit on the cancel paths.

Backend/app/dashboard/provider-dashboard/src/API/ProviderPlatform/DynamicOfferDriver/InternalAdmin.hs
```diff
@@ -0,0 +1,157 @@
...
+type RideInternalAPI =
+  "ride"
+    :> ( RideEndInternalAPI
+           :<|> RideCancelInternalAPI
+           :<|> RideSyncInternalAPI
+       )
+
+type RideEndInternalAPI =
+  "end"
+    :> QueryParam "requestorId" Text
```
sidsethu: **Blocker: caller-supplied `requestorId` is spoofable and reaches the finance ledger.**

`postRideEndInternal` passes `Just $ fromMaybe defaultRequestorId mbRequestorId` into `DRide.postRideEndMultipleInternal`, which forwards it to the BPP where `ActorInfo.withDashboardMbPersonIdActorInfo ((Id @DP.Person) <$> mbRequestorId)` stamps it onto ledger/finance records. So `POST .../ride/end?requestorId=<a real dashboard person uuid>` makes both the transaction row and the ledger read as if that operator ended the rides.

Also inconsistent: `ride/sync`, `booking/sync` and `booking/cancel/allStuck` hardcode `defaultRequestorId`; only `ride/end` and `ride/cancel` accept the override. If the caller identity genuinely needs to vary, derive it from the credential rather than an unauthenticated query param.

Backend/app/dashboard/provider-dashboard/src/Domain/Action/ProviderPlatform/Management/Ride.hs
```diff
@@ -59,6 +64,38 @@ buildManagementServerTransaction ::
...
+  Maybe (Id DM.Merchant) ->
+  Maybe request ->
+  m DT.Transaction
+buildRideTransaction endpoint mbRequestor mbMerchantId request = do
+  uid <- generateGUID
+  now <- getCurrentTime
+  pure
+    DT.Transaction
+      { id = uid,
+        requestorId = Id <$> mbRequestor,
```
sidsethu, line 88: **Blocker: `validateId` was dropped, and `transaction.requestor_id` is `character(36)`.**

`buildRideTransaction` is a copy of `SharedLogic.Transaction.buildTransaction` minus its `validateId` guard — the one whose comment reads *"we need to validate id length to avoid internal error while storing transaction"*. That guard exists for exactly this hazard, and the value is now caller-controlled via the `requestorId` query param.

`POST .../internal/admin/NY_PARTNER/Bangalore/ride/end?requestorId=<40-char string>`: `verifyApiKey` passes, the rides are actually ended downstream, then `T.withResponseTransactionStoring` runs `QT.create` and Postgres raises `value too long for type character(36)` -> **500 returned for an operation that already succeeded, with no audit row**. `?requestorId=` (empty) silently stores 36 blanks.

The same omission is in all five sibling copies (`ProviderPlatform/Management/Booking.hs`, `RiderPlatform/Management/{Ride,Booking}.hs` here, and both in rider-dashboard). Rather than re-adding the guard in six places, consider widening `buildTransaction` to take `Maybe (Id DP.Person)` + `Maybe (Id DM.Merchant)` instead of `Maybe ApiTokenInfo`, or adding a `buildServiceTransaction` beside it in `Backend/app/dashboard/Lib/src/SharedLogic/Transaction.hs`.

sidsethu (commented):
Review of the new internal-admin surface and the `is_stucked` marker. Posting the **4 blockers** inline; there are further correctness and duplication findings (one-way `is_stucked` latch set on healthy bookings, rider-app stuck-cancel not setting the marker at all, `INTERNAL_ADMIN_API` transaction rows being unreachable through the dashboard's person INNER JOIN, six copies of the transaction builder) that I can add if useful.

Backend/app/dashboard/provider-dashboard/src/API/ProviderPlatform/DynamicOfferDriver/InternalAdmin.hs
sidsethu: **Correction to the failure scenario above.**

I wrote that `booking/cancel/allStuck` with `bookingIds: []` returns all stuck bookings and cancels them merchant-wide. That is wrong. `findStuckBookings` applies `Se.Is BeamB.id (Se.In (getId <$> bookingIds))` unconditionally in both apps (`dynamic-offer-driver-app/.../BookingExtra.hs:241`, `rider-app/.../BookingExtra.hs:378`), so an empty list matches nothing rather than everything — and callers always passing explicit ids is the intended design. Apologies for the noise.

So the reachable surface is narrower than I described: exercising these routes requires knowing booking/ride ids.

The rest of the comment stands, and I would still treat it as blocking:

- `internal/auth` requires the api-key **and** a valid person token (`Auth.verifyPerson`), is read-only, and is self-scoped to the token holder's own fleet owners. `internal/admin` has no `token` header — the api-key is the only credential, and it authorizes writes. Everyone already holding the key gains ride-end/ride-cancel the moment this merges, with no re-issue or re-scoping.
- `merchantId`/`city` are path parameters and `checkMerchant` never binds the key to a merchant, so a single holder can act on any merchant on the instance.
- The literal `ae288466-...` is inline in `provider-dashboard.dhall:172`, `rider-dashboard.dhall:161` and `safety-dashboard.dhall:148`, and is absent from `secrets/provider-dashboard.dhall` unlike other credentials — compare `mock-registry.dhall:25`, which sources it as `sec.internalAuthApiKey`. There is no rotation path that does not require coordinating every holder at once.
- `throwError (InvalidRequest ...)` returns 400 rather than 401, so guessing against these routes reads as ordinary client errors to monitoring that alerts on 401 spikes.

A separate key for this route tree, sourced from the secrets file, addresses the first and third points without changing the design.

sidsethu (approved):
Approving.

Verified at `2fe32fe` that three of the four blockers from my earlier review are addressed:

- `migrations-read-only/{rider-app,dynamic-offer-driver-app}/booking.sql` are regenerated and committed, so the spec and the generated DDL agree.
- The `requestorId` query param is gone from the internal routes — the requestor tag is no longer caller-controlled.
- `T.validateRequestorId` was added to `SharedLogic/Transaction.hs` and is applied in all six transaction builders.

The open thread on reusing `internalAuthAPIKey` for this route tree is **not gating this merge** — flagging it for a follow-up rather than holding the PR.

Smaller items also worth a follow-up, none blocking:

- Rider-app `postBookingCancelAllStuck` does not set `is_stucked`, while the BPP twin does — so the marker is only written on the rider side by the sync path.
- Nothing ever sets `is_stucked` back to `False`, so a booking that is cleared and later becomes stuck again keeps the old value.
- The new `API/ProviderPlatform/DynamicOfferDriver/InternalAdmin.hs` is missing the AGPL copyright header that its siblings carry, including the `RiderPlatform/InternalAdmin.hs` added in this same PR.

PR #16580, opened 2026-09-03
backend/feat: implement negotiation fare tolerance parameters in fare…

FCJ6202:
… policy and transporter config

This commit introduces negotiation fare tolerance parameters to the fare policy and transporter configuration. Key changes include:
- Added `negotiationFareMinTolerancePct` and `negotiationFareMaxTolerancePct` fields to various data structures, including `FarePolicy`, `FullFarePolicy`, and `TransporterConfig`.
- Updated the logic in the `applyNegotiatedFare` function to utilize the new tolerance parameters from the quote's fare policy instead of the transporter config.
- Created SQL migrations to add the new columns to the database schema for fare policy and transporter config.
- Adjusted related CSV handling to accommodate the new negotiation fare tolerance fields.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Fare policies now support optional minimum and maximum negotiation tolerance percentages.
  - Fare policy CSV export and import include negotiation tolerance values, with validation limiting entries to 0–100%.
  - Negotiated-fare validation uses tolerance settings from the applicable fare policy, defaulting to 10% when not configured.

- **Changes**
  - Negotiation tolerance settings are no longer managed through transporter configuration.


0utkarsh:
## Review — request changes

Stack-reviewed with #16109 and #16595. Moving the tolerance to fare policy (per-vehicle-tier bands instead of city-wide) is the right call, and the CSV upsert path is wired correctly (values thread through to the policy, 0–100 validated). Two findings:

**1. Any dashboard V2 edit silently wipes CSV-configured tolerances** — `Domain/Action/Dashboard/Management/FarePolicyV2.hs` (`fromApiPolicy`) + `Storage/Queries/FarePolicy.hs` (`update'`)
`fromApiPolicy` hardcodes `negotiationFareMinTolerancePct = Nothing` / `negotiationFareMaxTolerancePct = Nothing` (the FPV2 API type has no such fields), the V2 update path recreates the policy from it (delete + create), and `update'` unconditionally `Se.Set`s both columns. So: ops uploads a CSV with 2/5% bands → later anyone edits an unrelated field of that policy via the dashboard → both columns go NULL → the next /select accepts bids in the fallback ±10% band the merchant configured to reject. The FPV2 diff view doesn't surface the loss either, since the API type doesn't know the fields.
**Fix**: add both fields to the FPV2 API type so they round-trip like every other fare-policy field.

**2. No backfill from the transporter_config values this replaces**
#16109 stored tolerances on transporter_config (Double fraction); this PR reads only `quote.farePolicy.negotiationFare*TolerancePct` (Int percent), and the new fare_policy columns start NULL — so any environment that set a non-default band under the previous schema silently reverts to ±10%. Add a backfill (transporter_config → that city's fare_policy rows) to the migration, or state explicitly in the PR that no environment ever set the old columns.

Nit: the two copy-pasted 0–100 `whenJust` validation blocks in `Merchant.hs` collapse into one `validatePct name` helper.

dushyant-2002 (dismissed):

RitikaHanish (commented):
CRITICAL Issues
[C1] Dashboard V2 fare-policy edit silently wipes the negotiation tolerance — data loss
[FarePolicyV2.hs:1133-1134](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/FarePolicyV2.hs#L1133-L1134) hardcodes both fields to Nothing, and [FarePolicyV2.hs:290-292](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/FarePolicyV2.hs#L290-L292) does delete old.id → create newPolicy. So the only way to set the value is CSV import, and any subsequent unrelated edit through the V2 dashboard destroys it — the band silently reverts to ±10% with no error. Two config surfaces write the same row and only one knows the field exists.
→ Carry the old values through ExistingPolicyCtx, or add them to Common.FPV2Policy so the API is complete.

[C2] Duplicate DDL migrations — the migration run will fail
The branch carries byte-identical pairs:

0865-add-negotiated-fare-delta-to-fare-parameters.sql and 0870-add-negotiated-fare-delta-to-fare-parameters.sql
0866-…-transporter-config.sql and 0869-…-transporter-config.sql
The base tip already has 0865/0866; this branch re-adds them at 0869/0870. All use bare ADD COLUMN with no IF NOT EXISTS, so the second to run errors with column already exists. Delete the duplicates during the rebase.

[C3] DDL still creates the transporter_config columns this PR deletes from the spec
0866/0869 add negotiation_fare_{min,max}_tolerance_pct double precision DEFAULT 0.1 to transporter_config, while the PR removes those fields from Merchant.yaml and migrations-read-only/transporter_config.sql. Two consequences:

Prod gets two permanently dead columns.
The schema ends up with identically named columns carrying different units — transporter_config.…_pct = 0.1 (a fraction) vs fare_policy.…_pct = 10 (a percent). That is a landmine for the next person.
Also there's no backfill from transporter_config → fare_policy, so any city already tuned off the default reverts to ±10%. Low real impact if the field never left this feature branch — please confirm before merge.

IMPORTANT Issues
[I1] Negotiation is not idempotent — repeated /select compounds the discount
[Select.hs:243-260](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs#L243-L260) anchors the band to quote.estimatedFare, then overwrites quote.estimatedFare with the negotiated value. BECKN /select is repeatable (retries, BAP re-negotiation) and nothing guards against it. Five successive selects at the lower bound → 0.9^5 ≈ 59% of the original fare, each step individually "within ±10%". negotiatedFareDelta is also overwritten relative to the already-negotiated fare, so it can no longer reconstruct the original.
→ Anchor to the original: originalFare = quote.estimatedFare - fromMaybe 0 quote.fareParams.negotiatedFareDelta, and always compute the delta against that. Pre-existing from the base branch, but this PR owns the function now.

[I2] Stale comment — [Select.hs:222-226](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs#L222-L226) documents validateQuoteSelect as "Pure read -- no DB writes", but it calls updateFareParameters and updateEstimatedFare. Move the write out of validation or fix the comment.

[I3] Maybe Int is the wrong type for a percentage here — siblings in the same record use Double (businessDiscountPercentage, personalDiscountPercentage, [FarePolicy.hs:78-79](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Types/FarePolicy.hs#L78-L79)). Int can't express 7.5% and there's no reason to be stricter than the neighbours. Prefer Maybe Double with a double precision column.

[I4] Validation lives only in the CSV path — applyNegotiatedFare trusts the row as-is. A direct DB write of min = 100 makes minAcceptable = 0, i.e. a free ride, accepted. Clamp at read time. Also: is 100 a sane cap for a min tolerance? A business-level cap plus a min <= max sanity check would be better than the current independent 0–100 checks.

[I5] NegotiatedFareNotAcceptable tells the BAP nothing — no acceptable range in the error, so the BAP can only guess-and-retry. Include the band (or at minimum log it).

Good to Have
[G1] 0865/0866/0867 already collide with files on main (0865-add-scheduled-booking-transaction-id-to-issue-report.sql, 0865-continuous-batching.sql, 0866-add-scheduling-charge-to-fare-policy.sql, 0866-geohash-area-unique-index.sql). Harmless here (independent ADD COLUMNs), but renumber above main's max on rebase.

[G2] farePolicyCSVHeader gained two trailing columns, changing export shape. Import is correctly tolerant (<|> pure ""); flag for any downstream consumer pinned to a fixed column count.

[G3] Resolve or drop the -- We can discuss on this comparisation logic comment (also a typo) before merge.

PR #16595, opened 2026-09-03
backend/feat: refactor MSIL transformers to consolidate common functi…

FCJ6202:
…onality

This commit introduces a new module, Beckn.OnDemand.Utils.MSIL.Common, to centralize common overrides and helpers used across various MSIL Layer 2 transformers (Search, OnSearch, OnSelect, OnInit, OnConfirm, OnStatus). Key changes include:
- Replaced multiple specific utility modules with a single common module to reduce indirection and improve maintainability.
- Updated transformers to utilize the new common functions for handling category codes, fulfillment states, and other shared logic.
- Removed obsolete modules related to MSIL category, fulfillment state, and fulfillment type, streamlining the codebase.
- Enhanced the handling of BAP_TERMS and fulfillment types in the transformers to align with ONDC v2.1.0 specifications.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added standardized ONDC scheduled-ride processing across search, select, init, confirm, and status flows.
  * Scheduled rides now apply consistent compliance formatting for categories, fares, fulfillment details, breakup items, vehicle information, terms, and route data.
  * Added configurable minimum and maximum fare negotiation tolerance fields.

* **Bug Fixes**
  * Improved fulfillment ID handling to preserve consistent identifiers throughout scheduled-ride flows.
  * Improved scheduled-ride status and update messages for more consistent ONDC compliance.


0utkarsh:
## Review — approve after cleanups

Stack-reviewed with #16109 and #16580. The consolidation is genuinely good: verified function-by-function against all ten deleted `MSIL/*` modules — every override survives into `OndcScheduledRide/Common.hs` with composition order, gating, and per-call-site variations preserved, and the `MSIL` → `OndcScheduledRide` rename is the right naming (it's a dialect, not a customer). The per-endpoint patch chains folding into single builders and `applyOndcScheduledRideOrderOverridesIfEnabled` are exactly the extraction the previous structure needed.

Cleanups before merge:

1. **Hand-edited `src-read-only/` files** (CLAUDE.md rule 1): `Storage/Queries/BecknConfig.hs` and `Storage/Beam/BecknConfig.hs` carry paren-only churn with no spec change in this PR — and the two files' paren changes go in opposite directions, so at least one is a hand edit. Regenerate with `, run-generator --apply-hint` and commit that output (the root cause is #16109 committing non-hint generator output; fixing it there ends the churn).

2. **Extract the pilot-gate predicate**: this PR centralized the *apply* side but the gate fetch (`getOneConfig` + `fromMaybe False .enableOndcScheduledRideSupport`) is still copy-pasted at ~9 sites (Search/Select/Init/Status handlers, Confirm.validateRequest, three CallBAP senders, twice in Utils/Common). One `isOndcScheduledRideSupportEnabled :: Id MerchantOperatingCity -> m Bool` beside the apply helper — this also makes the likely move to a per-BAP gate a one-place change.

3. **`OnStatus.hs` hand-composes a 4-stage override chain** that duplicates a subset of Common's pipeline, and OnConfirm/OnInit must each remember to prepend `overrideOrderBreakupTitles` outside the "full order patch" they call — fold these into the `applyOn*` functions in Common.hs so the next compliance fix can't miss a site.

4. **Stray payment-lib hunk** (`Lib/Payment/Domain/Action.hs`, `PInterface.udf2` qualification): compile-neutral disambiguation, but nothing else here touches the payment lib — move it to its own commit or add a note explaining why it rides along.

5. **Migration renumbering** (0867→0866 + deletion of the #16109 transporter-config DDL): fine for a stack that lands together, but environments tracking applied DDL by filename can get confused — prefer keeping numbers stable within a stack.

Minor: `withBapTerms`'s `otherwise = \tagGroups -> tagGroups` is just `id`; `applyOrderCategoryAndFulfillmentStateOverrides` is a single-caller wrapper whose doc points at a CallBAP call site that doesn't exist — inline it or fix the comment.

RitikaHanish (changes requested):
CRITICAL
[C2] Generated files changed with no spec change, and the changes are formatting regressions. [Storage/Beam/BecknConfig.hs:31](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Storage/Beam/BecknConfig.hs#L31) and [Storage/Queries/BecknConfig.hs:27](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Storage/Queries/BecknConfig.hs#L27) go backwards: m [X] → m ([X]), fmap showBaseUrl x → (fmap showBaseUrl) x, (B.C f ...) → B.C f .... No spec/Storage/*.yaml changed in this PR. This is NammaDSL generator-version drift picked up incidentally — revert both files.

[C2] Dead + unit-conflicting schema left behind by the migration cleanup. The PR deletes duplicate 0866-add-scheduled-category-signal-and-negotiation-tolerance-to-transporter-config.sql and renumbers 0867-add-negotiation-tolerance-to-fare-policy → 0866 (resolving a real collision with 0867-geohash-area-unique-index). That part is correct. But the surviving [0869](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/dev/ddl-migrations/dynamic-offer-driver-app/0869-add-scheduled-category-signal-and-negotiation-tolerance-to-transporter-config.sql) still adds:


negotiation_fare_min_tolerance_pct double precision DEFAULT 0.1  -- transporter_config
while [0866](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/dev/ddl-migrations/dynamic-offer-driver-app/0866-add-negotiation-tolerance-to-fare-policy.sql) adds negotiation_fare_min_tolerance_pct integer on fare_policy — and [Select.hs:247](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Select.hs#L247) reads only the fare_policy one. So: two columns, identical names, different units (0.1 fraction vs 10 percent), one set entirely dead. Drop those two lines from 0869 now, while the branch is unshipped — after it ships that's a DROP COLUMN.

Also worth confirming: is renumbering 0867 → 0866 safe given your migration runner? If any environment has already recorded 0866 as applied, the renamed file silently never runs.

IMPORTANT
[I2] Silent removal of a /confirm input validation, not in the PR description. [API/Beckn/Confirm.hs](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/API/Beckn/Confirm.hs#L91) drops CQMOC.findByMerchantIdAndCity transporterId city >>= fromMaybeM (InvalidRequest "Operating City … not supported or not found"). The pilot gate now keys on booking.merchantOperatingCityId instead of the context city. Treating the booking as authoritative is defensible (and removes a real gate/gate disagreement with callOnConfirm), but a /confirm carrying an unserviceable context city is no longer rejected. Call it out in the description, or keep the check.

[I6] Two opposite config-access patterns introduced in the same PR. sendOnConfirmToBAP/buildOnConfirmMessage now take TransporterConfig as a threaded parameter, while applyOndcScheduledRideOrderOverridesIfEnabled fetches it internally via getOneConfig. getOneConfig is cached, so the threading buys almost nothing and permanently couples a SharedLogic signature to what the caller happens to hold. Pick one — I'd drop the parameter and fetch inside, matching the helper.

[I2] Deleted comments that encoded non-obvious constraints. Most of the −1373 is comment removal and most of it is fine (genuinely redundant prose). These four are not:

MSIL/Init.hs documented why correctFulfillmentId lives in its own module — an import cycle (Domain.Action.Beckn.Init ← SharedLogic.CallBAP). Now that everything else merged into Common.hs, the next person to "finish the consolidation" will reintroduce it.
sendDriverArrivalUpdateToBAP: "Do NOT dual-send on both endpoints — the BAP has no way to dedupe the same event on two APIs." That's an idempotency invariant, not commentary.
sendDriverOffer: the note that Beckn.ACL.OnSelect.mkFulfillmentV2 sets fulfillment.id to driverQuote.id instead of driverQuote.estimateId, which ONDC Workbench NACKs. The code still doesn't fix this — deleting the comment deletes the only record of a known open bug.
The staged promotion plan for the NEW → RIDE_CONFIRMED override (the one override that touches a call live for every merchant today).
[G2] Common.hs exports 27 names; 9 are module-internal (addTagGroup, overrideOrderFulfillmentState, overrideFulfillmentType, patchOrderRouteInfo, patchOrderTags, patchOrderVehicleEnergyType, applyOrderCategoryAndFulfillmentStateOverrides, overrideOrderFulfillmentId, overrideOrderStopAuthorizationStatus). Consolidating 10 modules into one is fine; exporting their entire former surface area gives up the encapsulation you just bought. Relatedly, the haddock on applyOrderCategoryAndFulfillmentStateOverrides says "the two fixes SharedLogic.CallBAP applies" — CallBAP no longer calls it.

System design — the thing worth discussing beyond this diff
One boolean now gates ~12 unrelated wire behaviors. enableOndcScheduledRideSupport controls: category renaming, fulfillment-state remap, fulfillment.type collapse, energy-type defaulting, breakup remap+drop, order.tags allow-list, ROUTE_INFO injection, fulfillment-id pinning, stop-auth flip, /confirm trip-category bypass, and two endpoint re-routings (driver-arrival on_update→on_status, safety alert on_update→on_support). The name says "scheduled ride support"; the behavior is "ONDC v2.1.0 TRV10 compliance profile." Flipping it for a new merchant changes a dozen things at once, and the two routing changes are a different risk class from the payload-shaping ones. This refactor was the moment to rename the flag and split routing from shaping — worth doing before the next merchant onboards.

Three places assert wrong data rather than omit it. overrideVehicleEnergyType defaults any unrecognized value to PETROL — an EV with a typo'd energy type is reported to the BAP as petrol. overrideFulfillmentType collapses everything non-RIDE_OTP to DELIVERY, which is precisely the lossy step that broke /init's fulfillment-id parse and now needs a compensating DB read (correctFulfillmentId) to undo. remapBreakupTitle returns Nothing for unmapped titles, silently dropping fare lines — so the breakup may no longer sum to the quoted total, which many ONDC BAPs validate. The old code carried a -- Need to discuss what should be passed on the energy-type one; that's gone now too.

dropNonConformingOrderTags is lossy by design — BPP_INVOICE_INFO and SETTLEMENT_TERMS are discarded, not relocated. Correct per the TRV10 schema, but any BAP consuming them loses data the instant the flag flips. At minimum log what was dropped.

PR #16639, opened 2026-09-07
fix: payment charge in rider invoice

karan2cmd:
## Type of Change

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Payment invoices now display customer-borne payment charges as a separate “Payment Charge” line item.
  * Pre-discount fares are adjusted to account for applicable payment charges.
  * Payment charges are consistently reflected across online, cash, discounted, refunded, tipped, and cancellation-related transactions.
  * Transactions without a payment charge continue to display and process a zero charge consistently.
  * Payment totals and ledger records now more accurately distinguish payment charges from other platform fees.


dhruv-1010 (dismissed):

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
```diff
@@ -1078,7 +1078,7 @@ TransporterConfig:
       paymentChargeBearer: Maybe PaymentChargeBearer
       paymentChargeRate: Maybe Double
       paymentChargeVat: Maybe Double
-      # Stripe connect-account maintenance charge (scheduled deduction).
+      clubPaymentChargeInEarnings: Maybe Bool
```
khuzema786: `clubPaymentChargeInEarnings` this can we rename as 
```
  clubProjectFareInRideInvoice: [ProjectFareParamsComponent],
  clubProjectFareInRefundsInvoice: [ProjectFareParamsComponent],
  clubProjectFareInEarnings: [ProjectFareParamsComponent],
  clubProjectFareInPayouts: [ProjectFareParamsComponent]
```
where, ProjectFareParamsComponent is as below:
`data ProjectFareParamsComponent = RIDE_FARE | TOLL_FARE | CANCELLATION_FARE | PARKING_CHARGE | PAYMENT_CHARGE`

So generically we can decide which components to club in respective APIs mapped based on this ProjectFareParamsComponent

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/CancellationFee.hs
```diff
@@ -89,6 +99,8 @@ settleCancellationFeeViaStripe booking ride personD cancellationBase cancellatio
             parkingCharge = 0,
             parkingChargeVat = 0,
             platformFee = 0,
+            paymentCharge = 0,
```
khuzema786, line 104: Payment Charge should not be zero ?

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs
```diff
@@ -628,7 +637,7 @@ rideAssignedReqHandler req = do
           let pickupAddress = listToMaybe $ catMaybes [booking.fromLocation.address.area, booking.fromLocation.address.street, booking.fromLocation.address.city]
               ledgerCtx = RidePaymentFinance.buildRiderFinanceCtx booking.merchantId.getId booking.merchantOperatingCityId.getId booking.estimatedFare.currency False booking.riderId.getId ride.id.getId Nothing Nothing pickupAddress
           estimatedBreakups <- traverse (buildFareBreakupV2 booking.id.getId DFareBreakup.BOOKING) (fromMaybe [] req'.fareBreakups)
-          mbLedgerInfo <- SPayment.buildLedgerInfoFromBreakups estimatedBreakups bookingDiscountAmount bookingPayoutAmount applicationFeeAmount 0 ledgerCtx
+          mbLedgerInfo <- SPayment.buildLedgerInfoFromBreakups estimatedBreakups bookingDiscountAmount bookingPayoutAmount applicationFeeAmount bookingPaymentChargeAmount 0 ledgerCtx
```
khuzema786, line 645: vat should not be 0

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs
```diff
@@ -580,7 +587,7 @@ rideAssignedReqHandler req = do
           estimatedBreakups <- traverse (buildFareBreakupV2 booking.id.getId DFareBreakup.BOOKING) (fromMaybe [] req'.fareBreakups)
           -- Online Ride Assigned branch (inside Just OnlinePaymentParameters case) → isOnline=True.
           let ledgerCtx = RidePaymentFinance.buildRiderFinanceCtx booking.merchantId.getId merchantOperatingCityId.getId booking.estimatedFare.currency True booking.riderId.getId ride.id.getId Nothing Nothing (listToMaybe $ catMaybes [booking.fromLocation.address.area, booking.fromLocation.address.street, booking.fromLocation.address.city])
-          mbLedgerInfo <- SPayment.buildLedgerInfoFromBreakups estimatedBreakups bookingDiscountAmount bookingPayoutAmount applicationFeeAmount 0 ledgerCtx
+          mbLedgerInfo <- SPayment.buildLedgerInfoFromBreakups estimatedBreakups bookingDiscountAmount bookingPayoutAmount applicationFeeAmount bookingPaymentChargeAmount 0 ledgerCtx
```
khuzema786, line 591: vat should not be 0

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs
karan2cmd: Fixed. `booking.paymentCharge` is stored VAT-inclusive, so parking the whole amount in the net field with a zero VAT was wrong.

Added `splitGrossPaymentCharge`, which reverses the rate the ride was priced at (recovered from the emitted breakup). Both ride-assigned fallback records now set `paymentCharge = bookingChargeNet` and `paymentChargeVat = bookingChargeVat`. Where no rate is available the whole amount stays as net — the total is still correct, only the VAT line is omitted.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs
karan2cmd: Fixed together with the sibling on the cash branch — same cause. `booking.paymentCharge` is VAT-inclusive, so both fallback records now split it via `splitGrossPaymentCharge` rather than setting the VAT to 0.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/CancellationFee.hs
karan2cmd: You were right that something was missing, though not in this record — these two are inert.

`cancellationLedgerInfo` goes to `makePaymentIntent` -> `upsertCoreRidePaymentLedger`, whose `newTotal` sums only the fare fields (`rideFare` .. `platformFee`). All of those are 0 here, so it takes the "Core ride payment total is zero, skipping create" branch and `buildRidePaymentInvoiceConfig` is never reached with this record. Setting them would have looked like a fix and changed nothing.

The real gap your question exposed is one function over: the cancellation invoice is built by `createPendingCancellationFeeLedger`, which wrote only the fee and its VAT. So the rider was captured `cancellationTotal + charge` while the invoice said `cancellationTotal`, and the charge had no ledger entry at all.

That is fixed in 5c35714871: it now takes the charge pair and emits two invoice lines plus two ledger legs. The legs credit `SellerRevenue`, not `OwnerLiability` — the charge is the payment intent's `applicationFeeAmount`, so it goes to the platform; crediting `OwnerLiability` would have over-credited the driver by the charge on every cancellation.

Only the Stripe path passes a non-zero pair; the cash, manual-due and carried-due paths are not captured through the gateway and pass `(0, 0)`.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -782,14 +786,15 @@ mkRidePaymentLedgerInfo ::
...
+  HighPrecMoney -> -- paymentCharge (the payment-charge share of platformFee)
   HighPrecMoney -> -- offerDiscountAmount
   HighPrecMoney -> -- cashbackPayoutAmount
   HighPrecMoney -> -- rideVatAbsorbedOnDiscount
   HighPrecMoney -> -- cancellationCharge (0 for normal ride)
   HighPrecMoney -> -- cancellationTax (0 for normal ride)
   FinanceCtx ->
   RidePaymentLedgerInfo
-mkRidePaymentLedgerInfo rideFare gstAmount tollFare tollVatAmount parkingCharge parkingChargeVat platformFee offerDiscountAmount cashbackPayoutAmount rideVatAbsorbedOnDiscount cancellationCharge cancellationTax financeCtx =
+mkRidePaymentLedgerInfo rideFare gstAmount tollFare tollVatAmount parkingCharge parkingChargeVat platformFee paymentCharge offerDiscountAmount cashbackPayoutAmount rideVatAbsorbedOnDiscount cancellationCharge cancellationTax financeCtx =
```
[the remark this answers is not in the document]
karan2cmd: Fixed. You were right and my earlier reasoning was wrong: I had assumed `platformFeeDue` aggregated across rides, but `duesResp` is built for a single ride (`rides = [DueAmountRide {rideId = ride.id}]`), so `ride.paymentCharge` is exactly the right figure.

One correction to the suggested fix though: passing the charge without also removing it from `platformFeeDue` would have double-counted it. The charge is booked inside `PlatformFee` on the ride path (there is no separate leg), and `rideFare + platformFee` cancels into `preDiscountFareLine`, so an added `Payment Charge` line would have inflated the invoice total by the charge.

It is now carved out instead — `platformFee = duesResp.platformFeeDue - debtPaymentCharge`, `paymentCharge = debtPaymentCharge`. Total unchanged, charge itemised instead of mislabelled as ride fare. VAT is 0 there because no fare breakup is loaded on that path, so there is no rate to reverse.

Backend/app/rider-platform/rider-app/Main/src/SharedLogic/Payment.hs
```diff
@@ -1405,6 +1414,7 @@ clearDuesForPerson person duesResp currency paymentMethodId = do
             0
             0
             duesResp.platformFeeDue
+            0
```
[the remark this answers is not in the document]
karan2cmd: Fixed. You were right and my earlier reasoning was wrong: I had assumed `platformFeeDue` aggregated across rides, but `duesResp` is built for a single ride (`rides = [DueAmountRide {rideId = ride.id}]`), so `ride.paymentCharge` is exactly the right figure.

One correction to the suggested fix though: passing the charge without also removing it from `platformFeeDue` would have double-counted it. The charge is booked inside `PlatformFee` on the ride path (there is no separate leg), and `rideFare + platformFee` cancels into `preDiscountFareLine`, so an added `Payment Charge` line would have inflated the invoice total by the charge.

It is now carved out instead — `platformFee = duesResp.platformFeeDue - debtPaymentCharge`, `paymentCharge = debtPaymentCharge`. Total unchanged, charge itemised instead of mislabelled as ride fare. VAT is 0 there because no fare breakup is loaded on that path, so there is no rate to reverse.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/FareCalculator.hs
```diff
@@ -1484,19 +1503,41 @@
...
   PaymentChargeResult
-finalisePaymentCharge mbDwc currentFare appliedCharge fareParams =
-  let preChargeFare = currentFare - appliedCharge
-      (adjusted, mbP, mbBearer) = applyPaymentChargeOnPayableFare mbDwc fareParams preChargeFare
+finalisePaymentCharge mbDwc currentFare appliedCharge mbDiscount fareParams =
+  let fareOnly = max 0 (currentFare - appliedCharge)
+      postDiscountBase = max 0 (fareOnly - fromMaybe 0 mbDiscount)
+      (charge, vat, mbBearer) = applyPaymentChargeOnPayableFare mbDwc postDiscountBase
+      (preCharge, preVat, _) = applyPaymentChargeOnPayableFare mbDwc fareOnly
+      customerBears = maybe False (== show DTC.PAYMENT_CUSTOMER) mbBearer
```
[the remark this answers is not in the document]
karan2cmd: Fixed — `maybe False (== show DTC.PAYMENT_CUSTOMER) mbBearer` is now `Just (show DTC.PAYMENT_CUSTOMER) == mbBearer`.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/Storage/Merchant.yaml
karan2cmd: Done in 3995508d00 — the shape is exactly as you described:

```
ProjectFareParamsComponent = RIDE_FARE | TOLL_FARE | CANCELLATION_FARE | PARKING_CHARGE | PAYMENT_CHARGE

clubProjectFareInRideInvoice    :: Maybe [ProjectFareParamsComponent]
clubProjectFareInRefundsInvoice :: Maybe [ProjectFareParamsComponent]
clubProjectFareInEarnings       :: Maybe [ProjectFareParamsComponent]
clubProjectFareInPayouts        :: Maybe [ProjectFareParamsComponent]
```

Listing a component means "show it as one row", omitting it means "show its parts separately". Nothing configured is separate, which preserves today's behaviour.

Two things worth your call before I go further:

**Only `clubProjectFareInEarnings` is read so far.** It replaces the boolean in `buildRideEarnings`. The ride-invoice, refund-invoice and payout surfaces have no clubbing logic to point at today, so those three keys are plumbed but unread — wiring them changes how each surface builds its line items, which I would rather do as its own change than fold into a merge commit.

**`RIDE_FARE` clubbing already exists on the ride invoice under a different name.** `driverInvoiceLineItemsVatInclusive` drives `clubVatInclusive` in `EndRide/Internal.hs`, collapsing ride fare and its VAT into one "(Incl. VAT)" row — which is what `RIDE_FARE` in `clubProjectFareInRideInvoice` would mean. Should the new key subsume and replace that flag, or is `clubVatInclusive` meant to stay as a separate concern? I did not want to silently duplicate it.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Domain/Types/TransporterConfig.hs
```diff
@@ -634,3 +640,5 @@
 $(Tools.Beam.UtilsTH.mkBeamInstancesForEnumAndList (''PayoutChargeBearer))
 
 $(Tools.Beam.UtilsTH.mkBeamInstancesForEnumAndList (''PayoutFeeType))
+
+$(Tools.Beam.UtilsTH.mkBeamInstancesForEnumAndList (''ProjectFareParamsComponent))
```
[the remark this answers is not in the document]
karan2cmd: Leaving this one. That line is NammaDSL output in `src-read-only/`, and the bracketed form is what the generator emits — every sibling in the same file is identical (ConnectChargeBearer, PaymentChargeBearer, PayoutChargeBearer, PayoutFeeType all use the same `mkBeamInstancesForEnumAndList (''X)` form).

Mine is flagged only because it is the changed line. Removing the brackets would make it the single odd one out in the file, and the next `, run-generator` would put them back — so it would be churn rather than a fix.

Happy to strip it if you would rather the scan be clean; say the word.

khuzema786 (approved):

PR #16641, opened 2026-09-07
backend/feat: integrate AddOn support into dynamic-offer-driver-app

FCJ6202:
Adds AddOn support across search, select, init, confirm, and the on_update/on_status pushes: a new AddOnConfig storage type describing per-tier add-on offerings, selection/validation/echo logic in SharedLogic.AddOn, AddOnData persisted on SearchTry/Quote/Booking, and threading through every ONDC-scheduled-ride pilot transformer/override so selected add-ons are echoed back on item.add_ons at every step of the flow.

## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added rider add-ons, including insurance options, to supported scheduled-ride search results.
  * Riders can select add-ons with quantity limits during booking.
  * Selected add-ons are retained through driver matching and echoed across booking updates.
  * Added support for displaying add-on quantities and negotiated fare changes in ONDC responses.

* **Validation**
  * Invalid, unavailable, duplicate, or incorrectly sized selections are rejected.

* **Data Updates**
  * Add-on selections are saved with searches, quotes, and bookings for consistent order details.


RitikaHanish (changes requested):
CRITICAL Issues
[C1] The add-on price is advertised and echoed but never charged. pricePerQuantity appears in exactly two places: [SharedLogic/AddOn.hs](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/AddOn.hs) buildSpecAddOn (display) and the generated Beam/Queries. Nothing adds pricePerQuantity × selectedQuantity into quote.estimatedFare, fareParams, or the booking fare. A rider selects RIDER_INSURANCE at ₹X, the BPP echoes it as accepted on on_confirm, and the fare is unchanged. Either this is deliberate (a free/informational opt-in — in which case pricePerQuantity, maxQuantity and selectedQuantity are decorative and should be dropped or documented as such), or it is straight revenue loss. This needs an explicit answer before merge — it's the first-principles question the whole feature rests on.

[C2] fromJust on a nullable field, in the search hot path. [Storage/CachedQueries/BecknConfig.hs](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/CachedQueries/BecknConfig.hs) — cacheMerchantIdDomainAny calls makeMerchantIdDomainAnyKey (fromJust config.merchantId). BecknConfig.merchantId is Maybe (Id Merchant) ([Domain/Types/BecknConfig.hs:43](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Domain/Types/BecknConfig.hs#L43)) and beckn_config.merchant_id is a nullable column. Banned by the repo's own rules, and trivially avoidable: findAnyByMerchantIdAndDomain already has merchantId in scope — thread it through.

[C3] findAnyByMerchantIdAndDomain makes on_search BPP_TERMS non-deterministic. beckn_config has one row per (merchant, domain, vehicle_category) — vehicle_category is NOT NULL default 'CAB'. The old call was findByMerchantIdDomainAndVehicle merchantId "MOBILITY" Enums.CAB; the new one is listToMaybe <$> findByMerchantIdAndDomain with no ORDER BY, so the terms now come from whichever row the scan returns first — then get cached under a vehicle-less key for configsExpTime, so whichever row won the race sticks for the whole TTL. All ~18 other call sites in this app use the vehicle-specific lookup. This is also unrelated to add-ons. Keep the explicit CAB lookup, or add a deterministic tiebreak.

IMPORTANT Issues
[I1] A repeat /select wipes a prior add-on selection. In applyNegotiatedFare, computeAddOnData returns [] for an empty sReq.addOns, and the write fires on isJust sReq.negotiatedFare || not (null sReq.addOns). The negotiation logic explicitly supports repeated /select on the same quote (that's what the currentFare = estimatedFare - negotiatedFareDelta anchor exists for). So: select #1 with add-ons, no bid → persisted; select #2 with a bid, no add-ons → the write fires and blanks add_on_data. /init's verifyAddOnEcho then sees existingAddOnData = [] and NACKs any echo — the booking dies. Guard the two halves of that write independently.

[I2] Price and descriptor aren't snapshotted at selection. AddOnData persists only {configId, selectedQuantity}; buildSelectedSpecAddOns re-reads pricePerQuantity/descriptorName/maxQuantity from the live config row at every echo. An ops price edit between /select and on_status means the same booking reports a different price than on_confirm did — on an insurance product. Worse, buildSelectedSpecAddOns silently drops a deleted config, so an already-confirmed booking's on_status quietly omits an add-on the rider selected. Anything priced must snapshot price + descriptor into AddOnData; the config row is the catalog, not the record of sale.

[I3] verifyAddOnEcho re-runs availability checks, letting ops break in-flight bookings. It delegates to resolveAddOnData, which re-validates enabled, city, and tier. Disabling an add-on in the dashboard therefore NACKs /init and /confirm for every rider who already selected it at /select — "Add-on is not enabled", mid-flow. The ratchet only needs identity + quantity equality against the persisted AddOnData; availability was decided at /select. Split a validation-free comparison out of resolveAddOnData.

[I4] getAddOn is an uncached, unindexed query on the on_search hot path. It runs on every pilot on_search. findAllByMerchantOpCityIdAndEnabled filters on merchantOperatingCityId, which per the repo's rules cannot be a KV secondary key — so it falls through to Postgres, and add_on_config.sql creates no index on (merchant_operating_city_id, enabled). This is a config table with a tiny row count and near-zero write rate; it should be a CachedQueries module exactly like BecknConfig sitting next to it in this same PR. At absolute minimum, add the index.

[I5] /confirm non-value-add gating was tightened for all trip categories. Before: not isOndcScheduledRideSupportEnabled && not isValueAddNP && ... — a pilot-enabled city let every trip category through for external BAPs. After: only OneWay OneWayOnDemandStaticOffer joins the allow-list. The new rule is the safer one and the comment says it's deliberate, but it is a live behavior change unrelated to add-ons: any pilot merchant currently confirming Rental, InterCity, or OneWay OneWayRideOtp as a non-value-add NP starts getting "Unserviceable trip category". Verify against what pilot BAPs actually send, and split it out.

[I6] getSelectedAddOns silently discards the selection on a multi-item order. case items of [i] -> Just i; _ -> Nothing in [OndcScheduledRide/Select.hs](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Beckn/OnDemand/Transformer/OndcScheduledRide/Select.hs). No error, no log — the rider believes they bought insurance, the BPP records none, and /init's echo check then NACKs. Throw explicitly or collect across items; silence is the worst of the three.

[I7] updateAddOnDetails is dead code on both Booking and SearchTry. Declared in both YAML queries: blocks, generated into src-read-only/, called from nowhere in src/. booking.addOnData is set once in buildBooking and never updated. Drop the declarations unless a follow-up needs them.

[I8] Every quote /select now takes a 60s Redis lock. Previously the no-negotiation path returned without locking (Nothing -> return quote). Now all /select calls enter withLockRedisAndReturnValue, including the common case that writes nothing — and lock contention surfaces as a failed request. Consider hoisting the lock inside the write guard.

Good to Have
[G1] Column type inconsistency for the same payload: quote_special_zone.add_on_data is jsonb (migration 0867), booking.add_on_data and search_try.add_on_data are json. Pick one — jsonb is the better default.
[G2] search_try.sql still ends without a newline, plus a doubled blank line in the appended block.
[G3] Commit 2 ("refactor BecknConfig queries and enhance fare calculation logic" — NEGOTIATED_FARE_DELTA breakup + findAnyByMerchantIdAndDomain) is a separate change from add-ons. Two commits is within the acceptable range, but C3 in particular deserves its own review.

PR #16651, opened 2026-09-08
backend/feat: Firebase Analytics as S2S event tracking provider (rider-app)

witcher-shailesh:
Two commits, reviewable independently:

1. `backend/fix: make local bootstrap migrations re-runnable` — pre-existing upstream breakage, unrelated to this feature (details at the bottom).
2. `backend/feat: Firebase Analytics as S2S event tracking provider` — the change described below.

## Summary

Wires **Firebase Analytics** (GA4 Measurement Protocol) in as a third S2S event provider for rider-app, next to MoEngage and CleverTap. The provider itself lives in shared-kernel (nammayatri/shared-kernel#1506, merged; pinned on `main` as `8e25413`); this PR persists the rider's GA4 app instance id and feeds it to the dispatcher.

## What changes

- **Identity.** New nullable `person.firebase_app_instance_id` (migration 1573, `Person.yaml` field + `updateFirebaseAppInstanceId` query). The client sends it as `firebaseAppInstanceId` in the existing `POST /profile` on every launch; it is updated whenever it differs (it changes on reinstall). `GET /profile` echoes it. This is the **GA4 app instance id** from the Analytics SDK (`getAppInstanceId`), not the Firebase installation id — they are different values and only the former is accepted by `/mp/collect`.
- **Dispatcher.** `Tools.EventTracking.trackEvent` looks the rider up only when `FirebaseAnalytics` is among the target providers and fills the two new `EventTrackingReq` fields (`appInstanceId`, `platform` from `clientDevice.deviceType`). All seven call sites are unchanged. A failed lookup degrades to "no Firebase metadata" rather than blocking the other providers.
- **Tests.** New Newman suite `dev/integration-tests/collections/EventTrackingFlow`, plus `run-tests.sh event-tracking` which fails the run when seeding fails, restarts `rider-app-exe` after seeding so the ConfigPilot in-process cache cannot serve the pre-seed provider list, and is listed in `--list`.

Provider registration (the `FirebaseAnalyticsConfig` arm in the four service-config converters) and the `flake.lock` bump already landed on `main` in #16775, so they are not in this diff.

## Config shape

```json
{
  "baseUrl": "https://www.google-analytics.com",
  "apps": [
    { "platform": "ANDROID", "firebaseAppId": "1:123:android:abc", "apiSecret": "<passetto>" },
    { "platform": "IOS",     "firebaseAppId": "1:123:ios:def",     "apiSecret": "<passetto>" }
  ],
  "enabled": true,
  "debug": false
}
```

## Tests

`./run-tests.sh event-tracking` seeds all three providers for every city pointed at the mock server, sets the Firebase id through `POST /profile`, asserts `GET /profile` echoes it, runs a full auto ride so every tracked event fires with all three providers targeted, then rotates the id and asserts the new value. Every rider-side request carries `x-device` (rider-app re-syncs the stored client device from headers, and a request without it clears the device; the platform selects the Firebase app).

Delivery to the providers is not asserted: the mock server exposes recorded hits only as an SSE stream Newman cannot read, and adding mock-server code was ruled out for this change — rider-app logs are the evidence.

Verified locally on the dev stack: the rider-side slice (auth, profile set/echo, search, rotate/echo) runs 7 requests / 9 assertions / 0 failures. The rider-app log for the search event shows all three providers dispatched, and for Firebase `FirebaseAnalytics event ny_user_source_and_destination sent to app 1:1:android:test` — platform selected, secret decrypted, `/firebase/mp/collect` called with `firebase_app_id` and `api_secret` and accepted by the mock. The onboarding event, which fires at OTP verify before the client has sent the id, skips Firebase with the expected `no app instance id` debug line.

The full-ride portion of the suite needs a working provider-dashboard, which does not bootstrap on a fresh DB on current `main`. That is what the first commit fixes.

## First commit: local bootstrap migrations

Unrelated to Firebase, but it blocks running this suite (and any fresh local DB), and nothing on `main` fixes it as of `7efd6d9`. rider-app, driver-app and provider-dashboard each abort their migration transaction:

- **Not re-runnable.** `ADD COLUMN` without `IF NOT EXISTS` in rider 1566, 1567, 1569 and driver 0869, 0876, 0877 aborts the whole transaction on a second run. Driver 0867 re-adds a constraint 0862 already added; since `ADD CONSTRAINT` has no `IF NOT EXISTS`, it is wrapped in a `DO` block swallowing `duplicate_table` / `duplicate_object`.
- **Missing capability row.** provider-dashboard fails on `capability_endpoint_capability_id_fkey`: the generator emits the `capability_endpoint` and `role_capability` links from the spec's `migrate: capability:` line but not the `capability` row itself, so `city-operations.driver_management.write` is absent. Seeded in `0009-capability-driver-management.sql`, following 0006 and 0007.
- **Comment-only generated files.** `Local_API_Management_Customer.sql` and `Local_API_Management_Offer.sql` contain nothing but comments, and the runner rejects an empty query. A `DO` block is appended rather than a `SELECT` because the runner uses `execute`, which refuses a result set.

The last one is a stopgap worth calling out in review: both files are generated, so a regeneration drops the block. The correct fix is for the runner to skip an empty statement, but that code is in mobility-core, not this repo.

**Verification.** Every one of the 11 files was applied twice in a row against a live PostgreSQL 14.9 (the dev stack's own DB): 22 applications, 0 failures. The resulting schema was then checked rather than just the exit codes — all 12 columns exist with the right types and their `DEFAULT`s preserved (`'{RIDE_RELATED,SAFETY}'::text[]`, `false`, `0.1`), `geohash_area_unique_idx_merchant_operating_city_id_geohash` exists exactly once rather than being duplicated, and the seeded capability row is present with its `capability_endpoint` foreign key resolving.

Earlier, on a fresh DB: rider, driver and provider-dashboard all reach 200 with no SQL errors and `, import-config-data` completes clean.

Migration numbering is a moving target on this branch: the Firebase migration has been renumbered three times (1565 → 1567 → 1568 → 1573) as other people's migrations merged, and the capability seed moved 0008 → 0009 after #16799 took 0008. #16851 cleaned up the pre-existing duplicate indices upstream, so as of `7efd6d9` there are no overlapping numbers on this branch at all. Worth re-checking immediately before merge.

## Client contract

The app sends `firebaseAppInstanceId` (from `FirebaseAnalytics.getInstance().appInstanceId` / `Analytics.appInstanceID()`) in `POST /profile` on every launch and after login, keeps sending `x-device` on every request as it does today, and calls the SDK's `setUserId` with the rider id so `user_id` stitches. The client side is merged (ny-react-native #7747). For a rider whose app predates it, Firebase is skipped with a debug log and the other providers are unaffected.


witcher-shailesh:
Rebased onto current main and addressed the review:

- **run-tests.sh `--list`**: EventTrackingFlow added.
- **Seeding failures**: `seed_event_tracking_config` now runs psql with `ON_ERROR_STOP=1`, returns non-zero, and `run_event_tracking` aborts instead of running Newman.
- **ConfigPilot in-process cache**: after seeding, run-tests restarts `rider-app-exe` (process-compose brings it back) and waits for `/v2` before running the suite; Rules.md updated.
- **Rider lookup blocking other providers**: the Firebase-only `findById` is wrapped in `try`; a failure logs a warning and degrades to "no Firebase metadata", so MoEngage and CleverTap still receive the event.

Not changed:
- **Assert provider delivery via the mock server**: the mock only exposes recorded hits as an SSE stream, which Newman cannot consume, and adding mock-server code was explicitly ruled out for this change. rider-app logs show the Firebase call and its acceptance; this is documented in the suite's Rules.md.
- **Comments on `docs/superpowers/...`**: those files are not part of this PR any more (they were unrelated local commits that rode along before the rebase).

witcher-shailesh:
Client half is up: nammayatri/ny-react-native#7747 sends `firebaseAppInstanceId` via `POST /profile` and sets the Firebase user id to the rider id, so the `user_id` this PR sends stitches on the GA4 side. It reuses the existing device-field reconciliation in the session-mount effect, so there is no extra request.

Either order works: the client decoder treats the response field as optional, so it is safe against a backend that does not yet return it.

witcher-shailesh:
Rebased onto current `main` (62 commits) and re-pointed `flake.lock` at the rebased shared-kernel branch. Also fixed two review comments: `Rules.md` no longer links the design spec that was dropped from this PR, and the rider-app restart requirement is now stated on the manual run paths (dashboard and raw newman), not just the scripted one.

**Heads up: `main` currently fails to build `dynamic-offer-driver-app`, independent of this PR.**

```
src/Domain/Action/UI/Driver.hs:1790:36: error:
    Could not deduce (HasField "kafkaProducerTools" r KafkaProducerTools)
      arising from a use of 'TPayment.fetchLegalEntityName'
```

`fetchLegalEntityName` is constrained by `ServiceFlow`, which shared-kernel commit `142b092a` extended with `HasKafkaProducer`. That commit is already inside the shared-kernel revision `main` pins (`d7123b1135`), and `makeDriverInformationRes` was never given the constraint.

Why it is not from this PR:
- `Domain/Action/UI/Driver.hs` here is byte-identical to `origin/main`.
- Neither commit in this stack touches any driver-app file, or `Kernel/External/Types.hs` where `ServiceFlow` is defined.
- The only shared-kernel delta this PR introduces over what `main` already pins is the single Firebase provider commit.

`rider-app` (the package this PR changes) builds and links cleanly, and `Tools.EventTracking` and `Domain.Action.UI.Profile` compile. Someone should add the missing constraint to `makeDriverInformationRes` on `main` so CI goes green again.

PR #16653, opened 2026-09-08
backend/refactor: rework dashboard person access checks — remove call…

yanab-1:
## What

Reworks the authorization model for admin-by-person-id mutations in the dashboard
(`lib-dashboard`, `Domain/Action/Dashboard/Person.hs`):

- **Removed `assertPersonInCallerMerchant` entirely** (function + all call sites).
  It rejected mutations on persons not "owned" by the caller's merchant, throwing a
  disguised `PersonDoesNotExist` for persons that do exist — which blocked legitimate
  same-merchant access grants for users provisioned under other merchants.

- **`assignMerchantCityAccess`**: same-merchant grants are now open to any admin of
  that merchant, including onto persons provisioned under other merchants.
  Cross-merchant grants remain SUPER_ADMIN-only via `assertMayGrantAccessToMerchant`.

- **`changePasswordByAdmin` / `deletePerson`**: now SUPER_ADMIN-only (previously any
  admin, restricted to their own merchant's users). Password resets grant control of
  the target account across every merchant it can reach (one person row = one
  credential), and deletion destroys the person across all merchants — so both are
  gated on the seeded SUPER_ADMIN tier instead of merchant ownership.

- **`assignRole` / `resetMerchantAccess`**: ownership check removed; existing guards
  unchanged (`DCap.guardAdminMutation` still makes admin-tier role changes
  SUPER_ADMIN-only; route-level `DashboardAuth 'DASHBOARD_ADMIN` still applies).

- **`assertMayGrantAccessToMerchant`** simplified from `m Bool` to `m ()` — its
  return value's only consumer was the deleted ownership check.

## Why

Merchant admins were unable to grant their own merchant's access to existing users
provisioned under other merchants — the request failed with a misleading
`PersonDoesNotExist` even though the person existed. The ownership model is replaced
with a simpler split: same-merchant operations are open to that merchant's admins;
operations with cross-merchant blast radius (password reset, hard delete,
cross-merchant grants, admin-tier role changes) require SUPER_ADMIN.

## Behavior summary

| Operation | Before | After |
|---|---|---|
| Grant own-merchant access to own user | any admin | any admin |
| Grant own-merchant access to another merchant's user | ❌ blocked | ✅ any admin |
| Grant cross-merchant access | SUPER_ADMIN | SUPER_ADMIN |
| Reset password by admin | admin, own users only | SUPER_ADMIN only |
| Delete person | admin, own users only | SUPER_ADMIN only |
| Assign admin-tier role | SUPER_ADMIN | SUPER_ADMIN |

## Notes for reviewers

- An admin can now attach another merchant's user to their own merchant. The
  password-takeover chain this previously enabled is closed at the password-reset
  step instead (SUPER_ADMIN-only).
- `resetMerchantAccess` / `resetMerchantCityAccess` currently have no merchant-target
  scoping (any admin can revoke any person's access to any merchant) — flagged for a
  follow-up if we want `assertMayGrantAccessToMerchant` added there too.

## Summary by CodeRabbit

- **Access Control**
  - Updated merchant access checks for administrative actions.
  - Restricted password changes and person deletion to super administrators.
  - Adjusted merchant and city access assignment and reset workflows.
  - Removed certain merchant-boundary checks from role, contact, and account-status updates.

### DIFF
--- a/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
+++ b/Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs
@@ -256,9 +256,8 @@ adminEmailDomainError = "Administrator accounts must use an approved organizatio
 --
 -- One person row carries one email across every merchant that person can reach, so checking only
 -- the caller's merchant is too weak: an admin of a permissive merchant could set an address that
--- violates a stricter merchant the same person administers. Reachability is defined exactly as in
--- 'assertPersonInCallerMerchant' — access rows when there are any, the provisioning merchant
--- otherwise — so the two guards cannot drift apart. The caller's own merchant is always included,
+-- violates a stricter merchant the same person administers. Reachability is access rows when
+-- there are any, the provisioning merchant otherwise. The caller's own merchant is always included,
 -- which is what preserves today's behaviour for a person with neither access rows nor a
 -- provisioning merchant.
 --
@@ -293,46 +292,6 @@ assertAdminEmailDomainForPerson tokenInfo person role mbEmail =
     merchants <- policyMerchantsForPerson tokenInfo person
     forM_ merchants $ \merchantId -> assertAdminEmailDomain merchantId role mbEmail
 
--- | Admin mutations that address a person directly by id must not reach across merchants.
--- Without this an admin of any merchant could act on an arbitrary person id.
---
--- merchant_access rows are authoritative whenever the person has any: those merchants, and only
--- those, may act on them. A person shared across merchants therefore stays manageable by each.
---
--- The access rows cannot be the whole story though, because they are deletable. An earlier
--- version of this function used them alone and was bypassable: resetMerchantAccess and
--- resetMerchantCityAccess delete access rows and leave the person row alive, so an attacker could
--- empty a victim's rows and then claim them as "unowned". That state is also reachable with no
--- attack at all — a merchant revoking its own user's last access produces it.
---
--- person.merchantId, the merchant the person was provisioned under, closes that. It is written
--- once at creation, never updated, and not settable from any admin endpoint, so it holds the
--- claim when the rows are gone. It is deliberately consulted ONLY as a fallback rather than
--- unioned in: unioning would leave the provisioning merchant with authority forever, including
--- over a person whose access has since moved entirely to somebody else. Access, once granted,
--- decides; provisioning only decides when there is no access to speak of.
---
--- No claimant at all is still permitted, and now means one of two things: the person was created
--- moments ago and has not been granted access yet (createPerson writes no access row;
--- createUserForMerchant grants it on the next line), so rejecting would strand an admin who
--- typo'd an email at creation; or the row predates this column and the backfill found no access
--- row to derive one from. Neither is forgeable by a caller.
-assertPersonInCallerMerchant ::
-  BeamFlow m r =>
-  TokenInfo ->
-  Id DP.Person ->
-  m ()
-assertPersonInCallerMerchant tokenInfo personId = do
-  person <- QP.findById personId >>= fromMaybeM (PersonDoesNotExist personId.getId)
-  allAccess <- QAccess.findAllMerchantAccessByPersonId personId
-  let claimants =
-        if null allAccess
-          then maybe [] (: []) person.merchantId
-          else map (.merchantId) allAccess
-  unless (null claimants) $
-    unless (tokenInfo.merchantId `elem` claimants) $
-      throwError (PersonDoesNotExist personId.getId)
-
 -- | Granting a person access to a merchant is how somebody becomes a user of that merchant, so
 -- leaving this open undoes every other cross-merchant guard: an admin of B could grant their own
 -- user access to merchant A. Callers are held to their own merchant, with an escape hatch for a
@@ -340,14 +299,13 @@ assertPersonInCallerMerchant tokenInfo personId = do
 --
 -- Unconditional, matching DCap.guardAdminMutation: the SUPER_ADMIN tier is seeded (seed-migration
 -- 0018), so the existence guard that once kept these rules dormant no longer has anything to wait
--- for. Returns True when this is a cross-merchant grant that was permitted.
-assertMayGrantAccessToMerchant :: BeamFlow m r => TokenInfo -> Id DMerchant.Merchant -> m Bool
+-- for.
+assertMayGrantAccessToMerchant :: BeamFlow m r => TokenInfo -> Id DMerchant.Merchant -> m ()
 assertMayGrantAccessToMerchant tokenInfo targetMerchantId
-  | targetMerchantId == tokenInfo.merchantId = pure False
-  | otherwise = do
+  | targetMerchantId == tokenInfo.merchantId = pure ()
+  | otherwise =
     unlessM (isSuperAdmin tokenInfo.personId) $
       throwError AccessDenied
-    pure True
 
 -- | Record an admin-initiated mutation against another person. Mirrors the shape deletePerson
 -- already uses: who did it (requestorId), to whom (request), and when. The target's id is the
@@ -475,7 +433,6 @@ assignRole ::
   Id DRole.Role ->
   m APISuccess
 assignRole tokenInfo personId roleId = do
-  assertPersonInCallerMerchant tokenInfo personId
   person <- QP.findById personId >>= fromMaybeM (PersonDoesNotExist personId.getId)
   oldRole <- QRole.findById person.roleId >>= fromMaybeM (RoleDoesNotExist person.roleId.getId)
   newRole <- QRole.findById roleId >>= fromMaybeM (RoleDoesNotExist roleId.getId)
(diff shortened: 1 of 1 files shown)


RitikaHanish (dismissed):
---

## CRITICAL

### [C-1] Four endpoints lose *all* target scoping, not just the ownership relaxation this PR set out to make

`Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs`

The motivating bug ("merchant admins can't grant their own access to a user provisioned elsewhere") lives only in `assignMerchantCityAccess`. But `assertPersonInCallerMerchant` is also deleted from `assignRole`, `changeMobileNumberByAdmin`, `changeEnabledStatus`, and `changeEmailByAdmin` with **no replacement check**.

The PR body says these still have route-level `DashboardAuth 'DASHBOARD_ADMIN` and `DCap.guardAdminMutation`. Neither is merchant-scoped:

- `DashboardAuth 'DASHBOARD_ADMIN` only proves the caller is an admin of *some* merchant.
- Layer A (`Tools.Auth.Capability.enforce`) checks capability ids only — never the target person's merchant.
- Layer C (`enforceResourceScopeFromRequest`) can't help: `ResourceType` is `ROUTE | SPECIAL_ZONE | SPECIAL_LOCATION | TICKET_PLACE`. There is no `PERSON` type, so the `personId` path capture is never scoped.

So the stated model ("same-merchant operations open to that merchant's admins") is not what ships for these four — they become **any-merchant** operations.

Target ids are trivially discoverable: `listPerson` ignores its `TokenInfo` entirely (`listPerson _ mbSearchString …`) and `QP.findAllWithLimitOffset` has no merchant predicate, so every admin can already enumerate every person in every merchant.

Concrete reachable consequences, given `person` is one global row:

1. **Cross-merchant privilege grant via `assignRole`.** `QP.updatePersonRole` is keyed on `personId` alone and writes both `roleId` and `dashboardAccessType`. `guardAdminMutation` only covers `DASHBOARD_ADMIN` and `DASHBOARD_RELEASE_ADMIN`, but `isAdminTier` also lists `MERCHANT_ADMIN` and `TICKET_DASHBOARD_ADMIN`. An admin of merchant B can now promote an arbitrary merchant-A person to `MERCHANT_ADMIN` of A, or demote A's existing `MERCHANT_ADMIN` to a no-capability role (admin lockout in A). Neither path needs SUPER_ADMIN.
2. **Cross-merchant identity mutation via `changeEmailByAdmin` / `changeMobileNumberByAdmin`.** Email is the primary login identifier (`QP.findByEmailAndPassword`). A B-admin can repoint any A-person's email — including A's SUPER_ADMINs — locking them out, and can rewrite another tenant's users' mobile PII. `changeMobileNumberByAdmin`'s only remaining guard is the config-driven `updateRestrictedBppRoles` name list.
3. The full account-takeover chain does appear closed today, but only **incidentally, not by any guard**: the password path is now SUPER_ADMIN-only, and the `mobileNumber + tokenNo` login branch is effectively dead because nothing in the codebase ever writes a non-`Nothing` `tokenNoHash` (all three sites set `Nothing`). There is also no unauthenticated email-recovery flow. Introduce either — populate `tokenNoHash`, or add email-based password reset — and repointing email/mobile becomes cross-tenant takeover. Relying on that is fragile.

**Recommendation:** keep `assertPersonInCallerMerchant` on `assignRole`, `changeMobileNumberByAdmin`, `changeEnabledStatus`, and `changeEmailByAdmin`, and relax it **only** in `assignMerchantCityAccess` — the one call site the reported bug touches. If a single unified model is wanted, add a `PERSON` resource type (or an explicit `assertPersonReachableByCallerMerchant`) rather than leaving the target unscoped.

### [C-2] `assignRole`'s admin-tier gate is narrower than `isAdminTier`

`Backend/app/dashboard/Lib/src/Domain/Action/Dashboard/Person.hs` (`assignRole`)

Separable from C-1 and worth its own fix. `guardAdminMutation` guards only `DASHBOARD_ADMIN` and `DASHBOARD_RELEASE_ADMIN`; `isAdminTier` additionally treats `MERCHANT_ADMIN` and `TICKET_DASHBOARD_ADMIN` as admin tiers. That gap was survivable while the ownership guard confined it to the caller's own merchant. With the guard gone it is a cross-tenant escalation primitive. Align `guardAdminMutation` with `isAdminTier`.

---

## IMPORTANT

### [I-1] `changePasswordByAdmin` / `deletePerson` becoming SUPER_ADMIN-only is a live functional regression, not just a tightening

Every merchant admin who does routine password resets and user deletion today now gets `AccessDenied`. The capability seed still grants those roles the capability behind the endpoint, so the capability matrix becomes misleading: `enforce` passes, then the handler hard-denies. Control-center nav/buttons are capability-driven, so these actions stay visible and 403.

Needs explicit product sign-off, an ops comms plan, and frontend coordination before merge. Cleaner: model this as a dedicated high-blast-radius capability (or extend `guardAdminMutation`) so the capability matrix stays the single source of truth, instead of hardcoding `isSuperAdmin` in the handler.

### [I-2] PR description contradicts the diff on `resetMerchantAccess` / `resetMerchantCityAccess`

"Notes for reviewers" says they "currently have no merchant-target scoping — flagged for a follow-up", but the diff *does* add `assertMayGrantAccessToMerchant tokenInfo merchant.id` to both. The new scoping is good (revocation is confined to the caller's merchant), but an admin who could previously revoke a *different* merchant's access for their own user now gets `AccessDenied` — a behavior change absent from the PR's summary table. Please update the description and the table.

### [I-3] `policyMerchantsForPerson` invariant is broken by the removal

It unconditionally prepends `tokenInfo.merchantId`, justified in its comment as preserving behaviour for a person with no access rows. That held because the caller's merchant was always one of the person's merchants. Now `changeEmailByAdmin` can run against a foreign person, so the caller's allow-list is imposed on a person unrelated to it — a spurious restriction (B's stricter domain policy blocks editing an A-person's email). The comment was edited to drop the `assertPersonInCallerMerchant` reference, but the rationale itself is now stale.

### [I-4] Stale comments now describe guards that no longer exist — actively misleading

- `changeEmailByAdmin`: *"Authorization first, and specifically before the uniqueness probe below…"* now sits above no authorization at all. (The enumeration oracle itself is not a regression — an admin could already probe arbitrary emails using their own `personId` — but either the comment goes or the guard comes back.)
- `changeEnabledStatus`: *"Guarding anyway turns a silent no-op into an explicit error"* — the code now does the opposite. A cross-merchant call returns `Success` having written nothing, which is exactly what the comment says to avoid.

---

## GOOD TO HAVE

- **[G-1]** `unlessM (isSuperAdmin tokenInfo.personId) $ throwError AccessDenied` is now inlined in three places (`assertMayGrantAccessToMerchant`, `changePasswordByAdmin`, `deletePerson`). Extract `assertSuperAdmin :: BeamFlow m r => Id DP.Person -> m ()`.
- **[G-2]** `assertMayGrantAccessToMerchant` is misnamed at its new `resetMerchantAccess` / `resetMerchantCityAccess` call sites — it now means "may administer this merchant's access rows", not "may grant". Suggest `assertMayAdministerMerchantAccess`.
- **[G-3]** Audit rows from `recordAdminActionOnPerson` / `buildDashboardAuthTransaction` carry `tokenInfo.merchantId` (the actor's), so cross-merchant mutations are invisible in the victim merchant's own audit trail. Consider recording the target's merchant too.
- **[G-4]** Guard-before-lookup ordering in `changePasswordByAdmin` and `deletePerson` is correct (`isSuperAdmin` precedes `QP.findById`), so neither leaks person existence. Good.

---

## Impact Summary

| Aspect | Status |
|---|---|
| Ride booking flow | NO |
| DB migrations | NO |
| API changes | NO (wire-compatible; **authorization semantics change**) |
| Generated files touched without spec | NO |
| PII handling | Cross-merchant writes to email/mobile now possible |
| Commits | 1 (squash: NO) |
| Risk level | **HIGH** |

## Recommendation: REQUEST CHANGES

The `assignMerchantCityAccess` relaxation is correct and well-reasoned — that is the actual bug. But the guard removal is applied to four more handlers the reported bug doesn't touch, and nothing in the auth stack replaces it for them, so they become cross-tenant primitives (role/privilege changes, email and mobile rewrites on other merchants' users).

Before merge:
1. Narrow the guard removal to `assignMerchantCityAccess` only.
2. Align `guardAdminMutation` with `isAdminTier`.
3. Fix the three stale comments.
4. Get explicit sign-off on the `changePasswordByAdmin` / `deletePerson` SUPER_ADMIN gate, and confirm the UI doesn't offer the action to users who will now be denied.

## QA should test

1. Merchant-A admin grants A-access to a person provisioned under merchant B → succeeds (the fix).
2. Merchant-B admin calls `assignRole` / `change/email` / `change/mobile` / `changeEnabledStatus` with a merchant-A person's id → must be denied.
3. Cross-merchant grant as a non-super-admin → `AccessDenied`.
4. Non-super-admin merchant admin calls `change/password` and `delete` → confirm the 403 is intended and that the UI does not offer the action.

hkmangla (approved):

PR #16711, opened 2026-09-10
backend/enh: Add the state in allowDestination states while creating  new city

vinit-j13:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [x ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added merchant configuration for managing allowed destination states.
  - Merchants can update permitted destination states for each operating state.
  - Added retrieval of configured destination states by state name.
  - New operating cities automatically synchronize with the merchant’s state configuration.
  - Destination-state settings are validated against the merchant’s operating states.
  - Separate access controls support viewing and updating these settings.


AyushMittal42 (commented):
## Review: PR #16711 — Add the state in allowDestination states while creating new city

**CI**: `hlint` ×2, `Scan code with HLint`, `check-conditions`, `Incremental Repl Check — Dashboard`, CodeRabbit — all green. **But** `cabal-pr-repl-check-rider.yaml` and `cabal-pr-repl-check-driver.yaml` are `on: workflow_dispatch` only and did **not** run, and that's where ~140 of the new hand-written lines live. Nothing in CI compiled `rider-app` or `dynamic-offer-driver-app`.

---

### Critical

**1. Creating an operating city now silently widens cross-state serviceability, with no flag and no opt-out.**

`rider-app/Main/src/Domain/Action/Dashboard/Merchant.hs:943-946`, `dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs:4164-4167`

`syncMerchantStateMesh` is called unconditionally inside `postMerchantConfigOperatingCityCreate` — not gated on `cityAlreadyCreated`, not gated on `req.enableForMerchant`. It **mutates existing `merchant_state` rows** (`row.allowedDestinationStates <> [newState]`), and those rows are live inputs to the ride flow:

- `dynamic-offer-driver-app/.../Domain/Action/Beckn/Search.hs:1060` — BPP destination-state gating on `search`
- `rider-app/.../SharedLogic/Serviceability.hs:18,28` and `.../Domain/Action/UI/Serviceability.hs:178,191` — BAP serviceability

So a city launch in state T makes every already-configured state of that merchant accept destinations in T. No RIDE_FLOW *file* changed, which is why no automated check flags this, but the data those files read does.

**QA should test**: (a) merchant with an existing `merchant_state` row restricted to same-state only → create a city in a new state → confirm the restriction change is intended; (b) `search` from the old state to the new state on BPP; (c) rider-side serviceability for the same pair; (d) re-running city-create for an existing city (idempotency); (e) city-create with `enableForMerchant: false`.

Beckn actions affected: `search` / `on_search` — a rejected destination surfaces as "no quotes" to the rider.

**2. The comment asserts a guarantee the code doesn't provide.**

`Dashboard/Merchant.hs:1596-1597` (and driver-app equivalent):

> `-- If the state already has a row it is already part of the mesh (and any manual override via the upsert API must be respected), so there is nothing to do and we skip the findAll entirely.`

The early-return only protects the **new** state's own row. Every *other* state's manually-curated row gets `newState` appended unconditionally in the `Nothing` branch. An admin who deliberately narrowed KA to `[KA]` will find it silently become `[KA, MH]` after the next city launch. Either honour overrides (e.g. an `isManualOverride` marker) or fix the comment — right now it reads as a safety property that isn't there.

---

### Important

**3. The mesh is seeded from `merchant_state` rows, not from states the merchant actually operates in — so the headline feature is likely a no-op in production.**

`Dashboard/Merchant.hs:1601-1602`:
```haskell
existingRows <- QMerchantState.findAllByMerchantId merchantId
let siblingStates = DL.nub $ map (.state) existingRows
```

`merchant_state` only has rows where someone explicitly configured one. For a merchant already operating in KA with no row, adding a city in TN produces `TN → [TN]` — identical to the `maybe [state]` default every reader already falls back to. Nothing changes until a *second* post-PR state is added.

Meanwhile `postMerchantConfigAllowedDestinationStates` (line 1619) validates against `QMOC.findAllByMerchantId` (**operating cities**). Two sources of truth for "states of this merchant" in the same feature.

Suggestion: seed `siblingStates` from `QMOC.findAllByMerchantId` and keep `merchant_state` as override storage only — that matches both the validation logic and the PR title. Note this makes issue #1 wider, which is exactly why the intent needs to be stated in the (currently empty) PR description.

**4. The upsert API is unidirectional while the auto-sync is bidirectional.**

`Dashboard/Merchant.hs:1579-1592` — `upsertMerchantAllowedDestinationStates` writes only `req.state`'s row. `POST {state: KA, allowedDestinationStates: [TN]}` allows KA→TN but not TN→KA; the admin must call the endpoint twice and nothing in the API or response says so. `syncMerchantStateMesh` always writes both directions. Pick one semantic, or document the asymmetry on the request type.

**5. `merchant_state` is per-app; a single dashboard call configures only one side.**

`atlas_app.merchant_state` (BAP serviceability) and `atlas_driver_offer_bpp.merchant_state` (BPP search gating) are separate tables behind separate endpoints (`RIDER_MANAGEMENT/...` vs `PROVIDER_MANAGEMENT/...`). If an operator updates only one, BAP shows the destination as serviceable and BPP drops the search — the rider gets a silent no-quotes result. Worth a runbook note, or gate both from one call.

**6. GET takes `stateName: Text` and parses it with `readMaybe`, while POST takes a typed `IndianState`.**

`spec/.../API/Merchant.yaml` (`mandatoryQuery: - stateName: Text`), handler at `Dashboard/Merchant.hs:1630`. Consequences: parsing is tied to exact Haskell constructor spelling; the OpenAPI schema exposes a free-form string instead of an enum (so the dashboard UI can't render a dropdown); the two endpoints disagree on the wire format for the same value. There's no precedent for `IndianState` as a query param in the specs — if it lacks `FromHttpApiData`, please note that in a comment; otherwise use the typed param.

**7. `# regenerate` left at the top of a committed spec.**

`rider-app/Main/spec/Storage/configs.yaml:1` — a local generator-forcing marker; no other `spec/*.yaml` in the repo has one. Please drop it.

**8. ~370 lines of unrelated generated-file churn from a mismatched toolchain.**

`CommonAPIs/src-read-only/API/Types/ProviderPlatform/Management/Endpoints/Merchant.hs` (nearly all of its 196/175 lines), `.../RiderPlatform/.../Merchant.hs`, `rider-app/.../src-read-only/API/Action/Dashboard/Management/Merchant.hs`, `rider-app/.../src-read-only/Storage/Queries/MerchantOperatingCity.hs`.

Two competing reformats: `'[JSON]` → `('[JSON])` and `fmap show x` → `(fmap show) x` (parens added), but also `Maybe (Int)` → `Maybe Int` and `m [X]` → `m ([X])` in the opposite direction. That's a NammaDSL/ormolu version skew vs. what produced the committed files. It buries the 8 meaningful generated lines and will conflict with every other open PR touching these files. Please regenerate with the pinned toolchain.

**9. ~45 lines duplicated verbatim across rider-app and driver-app.**

`upsertMerchantAllowedDestinationStates`, `syncMerchantStateMesh`, `postMerchantConfigAllowedDestinationStates`, `getMerchantConfigAllowedDestinationStates` are byte-identical in both apps. The per-app `Storage.Queries` make full sharing awkward, but the pure parts (the `nub (state : rest)` mesh computation, the validate-against-operated-states step) could move to a shared helper parameterised on the query functions.

**10. Please dispatch the rider + driver repl checks** (or paste a local `cabal build rider-app dynamic-offer-driver-app`). Neither package was compiled by any check on this PR.

---

### Minor

- PR description is the untouched template — no description, motivation, or testing notes, and the "modifies the database schema" / "API breaking changes" boxes are unticked even though two new dashboard endpoints and four capability migrations are added. Given issue #1, the intended serviceability semantics should be written down here.
- `findByPrimaryKey` (used in upsert/sync) and `findByMerchantIdAndState` (used in the GET) generate identical SQL — the PK *is* `(merchant_id, state)`. Pick one.
- `AllowedDestinationStatesResp` and `UpsertAllowedDestinationStatesReq` are structurally identical (`Dashboard/Common/Merchant.hs:865-877`); one type would do.
- In `provider-dashboard`/`rider-dashboard` `Domain/Action/RiderPlatform/Management/Merchant.hs:1045,1129` the return type is unqualified (`AllowedDestinationStatesResp`) while the request type on the line above is fully qualified. Cosmetic, but inconsistent with its neighbours.
- No list endpoint — an operator can only read one state at a time, so verifying a merchant's whole mesh means N calls with no way to discover N.

---

### Checks that came back clean

| Check | Result |
|---|---|
| UI backward compat | Clean — purely additive: 2 new endpoints, 2 new types, 2 new `MerchantUserActionType` constructors. No field renamed, removed, retyped, or made mandatory. |
| DB backward compat | Clean — no DDL. Only idempotent `INSERT ... ON CONFLICT DO NOTHING` into `capability_endpoint` / `role_capability`. Verified `city-config.launch.read`/`.write` are **pre-seeded** in `atlas_dashboard` (`dev/seed-migrations/provider-dashboard/0003-capability-seed.sql:143-144`) and `atlas_bap_dashboard` (`dev/seed-migrations/rider-dashboard/0004-capability-seed.sql:137-138`), so no FK `23503` startup crash. New YAML queries add no columns. |
| Redis pattern matching | Clean — `clearCache` uses `Hedis.del` on an exact key; no `KEYS` / `deleteAllWithPattern` / wildcards. |
| PII encryption | N/A — new fields are `IndianState` enums. |
| Index coverage | Fine — `merchant_state` PK is `(merchant_id, state)`, so `findAllByMerchantId`'s `WHERE merchant_id = ?` rides the PK index prefix. Both `merchant_state` and `merchant_operating_city` are in `disableForKV`, so `findAllWithKV` degrades to plain Postgres and the "no KV secondary key" concern doesn't bite here. `updateByPrimaryKey` does bump `updated_at`. |
| Commits | 1 — ideal, no squash needed. |

---

**Risk**: MEDIUM–HIGH — dashboard-only surface, but a city-create side effect mutates live serviceability gating.

The plumbing is clean and correctly additive, but the core logic has an unresolved fork: `syncMerchantStateMesh` seeds siblings from `merchant_state` (so the feature is probably a no-op today), and if that's fixed to read operated states it silently widens cross-state serviceability on every city launch — including over manually-narrowed rows the comment claims are respected. That intent needs stating before merge; the `# regenerate` marker and the generator-version churn should go in the same pass.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Merchant.hs
```diff
@@ -4928,6 +4972,32 @@ postMerchantConfigOperatingCityWhiteList _ _ req = do
 postMerchantConfigMerchantCreate :: ShortId DM.Merchant -> Context.City -> Common.CreateMerchantOperatingCityReqT -> Flow Common.CreateMerchantOperatingCityRes
 postMerchantConfigMerchantCreate = postMerchantConfigOperatingCityCreate
 
+postMerchantConfigAllowedDestinationStates :: ShortId DM.Merchant -> Context.City -> Common.UpsertAllowedDestinationStatesReq -> Flow APISuccess
+postMerchantConfigAllowedDestinationStates merchantShortId _ req = do
+  merchant <- findMerchantByShortId merchantShortId
+  now <- getCurrentTime
+  merchantStates <- DL.nub . map (.state) <$> QMOC.findAllByMerchantId merchant.id
```
AyushMittal42, line 4979: Can we send this merchantState to the upsert method instead of query there again

AyushMittal42 (approved):

hkmangla (approved):

PR #16758, opened 2026-09-15
fix: FY earnings

Yukesh-R:
## Type of Change

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

### Additional Changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a driver earnings API for viewing financial-year earnings by quarter.
  * Displays quarterly net earnings and TDS deductions, along with totals for the selected period.
  * Drivers can view their own earnings, while administrators can access authorized records.
* **Bug Fixes**
  * Added validation to reject quarter values outside the supported range.
  * Earnings results now consistently apply quarter filtering and chronological ordering.


RitikaHanish (changes requested):
CRITICAL
[C1] The fleet-association block is a current-state check applied to historical data — it hides a driver's own past earnings.
[DriverFyEarnings.hs:52-53](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverFyEarnings.hs#L52-L53)


mbFleetAssoc <- QFDA.findByDriverId personId True
whenJust mbFleetAssoc $ \_ -> throwError AccessDenied
findByDriverId _ True returns only the currently active association ([FleetDriverAssociationExtra.hs:183-188](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Storage/Queries/FleetDriverAssociationExtra.hs#L183-L188)). But rows are attributed at ride time: fromMaybe ride.driverId ride.fleetOwnerId ([EndRide/Internal.hs:553](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/Ride/EndRide/Internal.hs#L553)).

Failure scenario: a driver runs independently Apr–Sep (real Q1/Q2 rows keyed on their own personId), then joins a fleet in October. From that moment the endpoint returns 403 for the entire financial year — including the Q1/Q2 earnings that are unambiguously theirs. This is exactly the data a driver needs at filing time, and it disappears the day they sign with a fleet.

The guard is asking "are you in a fleet now?" when the only sound question is "which rows exist under your personId?". Fetch the rows first; the emptiness of the result is itself the correct signal.

[C2] Backward-compatibility break on a live dashboard endpoint: OPERATOR and FLEET_BUSINESS lose access.
[Dashboard/Management/Driver.hs:1555](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Driver.hs#L1555)


canReadFyEarnings requestor target
  | requestor.id == target.id = return True
  | otherwise = return (requestor.role == DP.ADMIN)
The old isAssociationWithDriver permitted OPERATOR → DRIVER, FLEET_OWNER → DRIVER, FLEET_BUSINESS → DRIVER. The new rule permits none of them.

Denying FLEET_OWNER → DRIVER is defensible — those rows live on the owner, so the old path returned misleading zeros. But OPERATOR → DRIVER is a real regression: an operator's independent drivers do have their own rows, and the operator could read them yesterday. Same for FLEET_BUSINESS. With an empty PR description and every checklist box unticked, there's nothing indicating this was intended.

Also note requestor.role == DP.ADMIN is effectively dead code on this path: the comment two lines above states admins aren't present at BPP, so whenJust never fires for them and they're allowed by fall-through. DP.ADMIN exists in the enum ([Person.hs:201](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src-read-only/Domain/Types/Person.hs#L201)) but not as a BPP person row.

Suggested shape — keep association-based reads, reject only the genuinely-misleading case:


canReadFyEarnings requestor target
  | requestor.id == target.id = pure True
  | target.role == DP.DRIVER, requestor.role == DP.FLEET_OWNER = pure False  -- rows live on the owner
  | otherwise = isAssociationWithDriver requestor target
IMPORTANT
[I1] AccessDenied is the wrong error for a legitimate driver. A fleet driver asking for their own earnings is not an authorization failure. AccessDenied is the kernel auth error used for token/merchant-access rejections; the driver app will render a generic "access denied" toast, which is actively misleading — the truthful statement is "your earnings are recorded under your fleet owner." This also contradicts the convention of not throwing for expected cases. Since the API is new, the compatible fix is free right now and expensive later: return 200 with quarters: [] plus a discriminator (attributedTo :: Maybe FleetOwnerRef or earningsAttribution :: SELF | FLEET_OWNER), or add a dedicated error code (FLEET_DRIVER_EARNINGS_NOT_AVAILABLE, 409/422) the app can branch on.

[I2] totalNetEarnings / totalTdsDeducted silently become quarter totals when quarter is passed. Both endpoints compute totals over the filtered list. A client that sends quarter=2 and renders totalNetEarnings next to financialYear displays Q2's number labelled as the FY total. Either always total the full FY (filtering only quarters), or rename to filteredNetEarnings. Fixed now while there are no clients.

[I3] The response-building logic is copy-pasted between the two handlers. The filter → map → sortOn (.quarter) → sum block and the quarter must be between 1 and 4 validation are duplicated verbatim in [Dashboard/Management/Driver.hs:1556-1575](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Dashboard/Management/Driver.hs#L1556-L1575) and [UI/DriverFyEarnings.hs:54-63](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverFyEarnings.hs#L54-L63). SharedLogic.DriverFyEarnings already exists and is the natural home — extract validateQuarter and a (quarter, net, tds) builder; the two response record types can be constructed at the edges. Without this, I2 will get fixed in one place and not the other.

[I4] The "requestor not found at BPP ⇒ allow" fallback is an open door on tax data. whenJust (find ...) means any dashboard principal whose personId has no BPP row reads any driver's FY earnings and TDS. Pre-existing (mirrored in getDriverEarnings), but this PR is rewriting precisely this check, and TDS figures are more sensitive than the ride stats the pattern was built for. Prefer fail-closed with an explicit allowance for the known admin case.

[I5] A load-bearing comment states the wrong source of truth. Both [DriverFyEarnings.yaml:13-15](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/spec/API/DriverFyEarnings.yaml#L13-L15) and the module header say "ride.fleetOwnerId is taken from the vehicle's RC." It isn't — Ride.fleetOwnerId comes from QFDA.findByDriverId driver.id True ([Confirm.hs:138](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/Beckn/Confirm.hs#L138), [SharedLogic/Ride.hs:526](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Ride.hs#L526)). The RC-derived field is RideDetails.fleetOwnerId ([SharedLogic/Ride.hs:474](vscode-webview://1m414gl6femapspd6fbgcnkqqfnuqtqipe5f29bnishfjf8gg027/Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/Ride.hs#L474)), a different column. The code happens to be right — the guard and the write path use the same query — but the next reader will reason from the wrong premise.

PR #16764, opened 2026-09-16
backend/hotpush: rate limit pass select API

Cyanid0:
Cherry-pick of #16705 onto prodHotPush-BAP.

Adds a per-person sliding-window rate limit on the user-facing pass select endpoints:
- POST /multimodal/pass/{passId}/select
- POST /multimodal/pass/v2/{passId}/select

Each select creates a Juspay payment order, and the existing purchase lock only blocks concurrent selects, not a sequential loop.

- New env config passSelectAPIRateLimitOptions (10 hits / 300s)
- Checked before any DB read; keyed per person so switching passId doesn't bypass it; dashboard purchases are exempt
- Exceeding the limit returns 429 HITS_LIMIT_EXCEED

cc: Vignesh S


Cyanid0:
Superseded by #16786 — same cherry-pick of #16705, retargeted to prodHotPush-Common.

PR #16804, opened 2026-09-18
backend/feat: NY-82256 show the bank a payout lands in

karan2cmd:
A driver cannot see where their money goes. The profile's `bankAccountNumber` is actually the Stripe connected-account id, and `payout_request` recorded no destination at all, so payout history could say `"BANK"` and nothing more.

## Bank destination

Stripe now reports the connected account's bank name and last4. Both are stored on `driver_bank_account` and reach the profile and every payout history row.

They are fetched in the `bankAccount/status` endpoint, which already talks to Stripe. Its guard stops asking once an account is linked — which would leave existing drivers empty forever — so it also refreshes while either detail is missing. A response carrying no external account leaves stored values alone rather than blanking details already shown.

The payout **snapshots** the pair onto its own row rather than referencing the bank account, so history keeps naming the bank actually paid even after the driver changes banks. For a fleet driver that is the fleet owner's account — where the money genuinely went.

## Also in this PR

- **Withdrawals dropped from the transaction breakdown.** A payout settles money already earned, so counting it as a deduction against earnings double-counted it. Still visible in transaction and payout history.
- **`payoutMethod`** was hardcoded to `"UPI"`; now inferred from what the request records (cash / VPA / bank).
- **Invoice recipient and supplier.** The ride-assigned ledger context carried the `buildRiderFinanceCtx` defaults, so the invoice the PDF actually selects — newest non-voided — had no recipient and no supplier. It now resolves both the way the ride-end paths do, seeding supplier fields from config before the booking's own values overlay them.
- **Payment charge VAT.** An override reported its whole VAT-inclusive amount as net, dropping the VAT line from every invoice it touched.

## Not in this PR

The redeemable / non-redeemable rework was reverted per review — another developer owns that change. `WalletSummaryResponse` is unchanged from `main`.

## Dependency

Builds on shared-kernel #1520 (**merged**); `flake.lock` is bumped to kernel `main` (`2c258561`).

Kernel `main` also carries `c731d203` ("fix: face detection api"), which made face detection asynchronous: `FaceDetectionSummary.status` is now `Maybe`. This PR includes **only the two-line type change** needed to compile against it (`DetectImageResp.status :: Maybe DetectImageStatus`, `toStatus <$> status`), so the detect response now returns `status: null` rather than a fabricated value. The async result endpoint (`GET …/detectFace`) is left to #16817; those two lines are identical to its edits, so the two PRs won't conflict on them.

## Notes for review

- Nothing is backfilled: `bank_name` / `bank_account_last4` populate on the driver's next status call, and payout rows only from the next payout onward.
- The profile needed no new code — `bankDetails` already projects `DriverBankAccount` through `RecordWildCards`.
- VPA payouts and the Idfy penny-drop path record neither field: one has no bank account, the other reports the account holder rather than the bank.
- **Verified against a live connected account:** `GET /v1/accounts/{id}` returns `external_accounts` with no `expand` param, e.g. `{object: "bank_account", bank_name: "STRIPE TEST BANK", last4: "0785", default_for_currency: true}` — the exact shape the parser expects, including the `default_for_currency` preference.
- Both apps build clean (`cabal build dynamic-offer-driver-app rider-app`); not runtime-tested.

## Summary by CodeRabbit

- **New Features**
  - Driver bank-account details and payout history now include the bank name and masked account digits.
  - Wallet summaries now show redeemable and non-redeemable balances.
  - Payout records now display the applicable payout method and destination details.

- **Bug Fixes**
  - Wallet transaction breakdowns no longer include payout entries.
  - Invoice charges now correctly separate applicable VAT.
  - Invoices now include rider and supplier details where applicable.
  - Previously stored bank details are preserved when payment-provider responses omit them.
  - Image detection responses now handle missing status information gracefully.


Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverWallet.hs
```diff
@@ -244,57 +254,47 @@ fetchWalletRowsFromCH ::
...
-          )
-          (Map.toList dedRefMap)
-      dedTotal = sum (Map.elems dedRefMap)
-      netEarnings = addTotal - dedTotal
-   in ( DriverWallet.WalletItemGroup {totalAmount = addTotal, items = addItems},
-        DriverWallet.WalletItemGroup {totalAmount = dedTotal, items = dedItems},
-        topLevelNonRedeemable,
-        netEarnings
-      )
+      addGroup = mkGroup addEntries
```
khuzema786: revert this change

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverWallet.hs
```diff
@@ -175,10 +175,18 @@ getWalletTransactions (mbPersonId, _merchantId, mocId) mbFromDate mbToDate mbAgg
...
-      let (additions, deductions, nonRedeemableBalance, netEarningsBalance) =
+      let (additions, deductions, netEarningsBalance) =
             aggregateWalletRows (isVatMerchant transporterConfig) accountIds cutoff rows
-          redeemableBalance = max 0 (currentBalance - nonRedeemableBalance)
           agg = bucketizeRows accountIds (generateBucketWindows aggBy timeDiff fromDate toDate) rows
+      -- These two split the CURRENT balance and ignore the requested window, so they hold
+      -- currentBalance = redeemable + nonRedeemable and stay put as the date range changes.
+      -- Deliberately the wallet (Liability) account alone, which is where currentBalance comes
+      -- from; the breakdown below spans Liability + Control and answers a different question.
+      nonRedeemableBalance <-
```
khuzema786: revert these changes

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/SharedLogic/PersonBankAccount.hs
```diff
@@ -237,8 +241,13 @@ getPersonRegisterBankAccountStatus mbForceRefresh personId merchantOpCityId = do
       forceRefresh = fromMaybe False mbForceRefresh
       currentlyDue = fromMaybe [] (bankAccount.requirements >>= (.currentlyDue))
       isBankAccountSuccessfullyLinked = bankAccount.chargesEnabled && null currentlyDue
+      -- A driver linked before bank details were stored would otherwise never refresh, since
+      -- the guard below stops asking Stripe once the account is healthy. Ask once more while
+      -- the details are missing so existing drivers backfill themselves; the sliding-window
+      -- limit still caps how often an account with no bank attached retries.
+      needsBankDetails = isNothing bankAccount.bankName
```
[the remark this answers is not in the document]
karan2cmd: Good catch — fixed. `needsBankDetails` now checks both fields:

```haskell
needsBankDetails = isNothing bankAccount.bankName || isNothing bankAccount.bankAccountLast4
```

An account with a name but no last4 would otherwise have stayed half-populated forever, since the healthy-account guard skips the Stripe call.

Backend/app/rider-platform/rider-app/Main/src/Domain/Action/Beckn/Common.hs
```diff
@@ -595,7 +595,17 @@ rideAssignedReqHandler req = do
...
+          -- carries the buildRiderFinanceCtx defaults (issuedToName = Nothing), and since a
+          -- re-price here voids the ride-end invoice and writes a fresh one, the invoice the
+          -- PDF actually renders ends up with no recipient and no supplier.
+          riderPerson <- QP.findById booking.riderId >>= fromMaybeM (PersonNotFound booking.riderId.getId)
+          let assignedRiderName = listToMaybe $ catMaybes [riderPerson.firstName, riderPerson.middleName, riderPerson.lastName]
+              ledgerCtx =
+                RidePaymentFinance.applyBookingProviderFieldsToCtx booking $
+                  (RidePaymentFinance.buildRiderFinanceCtx booking.merchantId.getId merchantOperatingCityId.getId booking.estimatedFare.currency True booking.riderId.getId ride.id.getId Nothing Nothing (listToMaybe $ catMaybes [booking.fromLocation.address.area, booking.fromLocation.address.street, booking.fromLocation.address.city]))
+                    { issuedToName = assignedRiderName
+                    }
```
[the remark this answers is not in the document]
karan2cmd: Fixed, and you were right that this mattered — `applyBookingProviderFieldsToCtx` only overlays supplier fields the BPP actually sent, so a booking without them left the invoice with no supplier at all. Confirmed in prod data: the ride-assigned invoice had `supplier_name = ''` while the ride-end one had the configured name.

Supplier fields are now seeded from `riderConfig.invoiceConfig` before the booking's own values overlay them, matching the ride-end paths.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverWallet.hs
karan2cmd: Reverted. `aggregateWalletRows` and the handler's balance computation are byte-identical to `main` again, and the two new `wallet*` response fields are gone — `WalletSummaryResponse` is unchanged from `main`, so that surface is free for the redeemable/non-redeemable work.

What remains in this PR on the wallet side is only the withdrawal exclusion from the breakdown (payouts settle money already earned, so counting them as a deduction double-counted), plus the bank name/last4 additions to payout history.

Backend/app/provider-platform/dynamic-offer-driver-app/Main/src/Domain/Action/UI/DriverWallet.hs
karan2cmd: Reverted. `aggregateWalletRows` and the handler's balance computation are byte-identical to `main` again, and the two new `wallet*` response fields are gone — `WalletSummaryResponse` is unchanged from `main`, so that surface is free for the redeemable/non-redeemable work.

What remains in this PR on the wallet side is only the withdrawal exclusion from the breakdown (payouts settle money already earned, so counting them as a deduction double-counted), plus the bank name/last4 additions to payout history.

khuzema786 (approved):

PR #16806, opened 2026-09-18
provider-platform/perf: fork BPP search-stage histogram emission off on_search hot path (fixes ~2x search latency)

RoString36222:
## What
Moves the BPP **search-stage** ride-value histogram emission into a `fork`, off the synchronous on_search path.

## Why (prod regression)
After the ride-value histograms (PR #16798) began emitting, BPP search round-trip **p90 latency ~doubled: ~0.8s → ~1.5–1.9s**, in lockstep with the histogram ramp (verified on VM). The `search`-stage block ran **inline** in `Domain/Action/Beckn/Search.hs` — before `addNearestDriverInfo` and the on_search callback — so `SML.getCityLabel` plus the per-estimate observes on a Label8 histogram vector (millions of `search`-stage label combinations, ~1.1M observes/hr) were added directly to the response path.

## Change
Wrap the search-stage emission (incl. `getCityLabel`) in `fork "BPP search-stage ride-value histograms"`, matching the existing fork pattern in the same handler (`fork "Add Namma Tags"`, `fork "Updating Demand Hotspots on search"`). pre_ride/completed emissions are untouched (bounded, cheap, and not on the search path).

## Note
This removes the latency regression. It does **not** reduce the `search`-stage cardinality itself (VM still holds millions of series for that stage and can 502 on wide queries) — a follow-up may drop the `search` stage or cut its label cardinality (distance_bucket / pickup_zone / drop_zone). Not compiled locally (no nix); please let CI gate.


RoString36222:
Superseded: built on a pre-merge base so it re-proposed the whole feature. Reopening a clean branch off latest prodHotPush-Common that only forks the quantile emissions.

PR #16852, opened 2026-09-21
backend/fix: drop pickup_zone/drop_zone from BPP ride-value histograms

piyushKumar-1:
## Type of Change

- [x] Bugfix

## Description

Removes `pickup_zone` and `drop_zone` from the `RideValueHistogram` label set (`P.Label8` → `P.Label6`), which covers all four ride-value histograms:

- `BPP_price_per_km`
- `BPP_congestion_charge`
- `BPP_ride_distance_meters`
- `BPP_pickup_distance_meters`

The two labels are dropped from the histogram registration, from the four `observe*` helpers in `Tools/Metrics/ARDUBPPMetrics.hs`, and from their 8 call sites. Nothing else changes — no schema, no dhall, no API.

### Additional Changes

## Motivation and Context

**driver-app has been silently missing from monitoring for hours at a time, and completely invisible overnight.**

`pickup_zone`/`drop_zone` are special-location ids, and the pair is an unbounded cross-product — ny-prod currently has ~436 distinct `pickup_zone` and ~483 distinct `drop_zone` values. On a counter that is affordable. On a histogram it is not, because every label combination costs one series **per bucket** (41 for the two `linearBuckets 0 5 40` metrics).

A pod keeps discovering new zone pairs for its entire life, so its `/metrics` payload grows monotonically and never plateaus:

```
~4,170 zone combos  →  ~390k samples  →  >100 MB
```

That crosses vmagent's `-promscrape.maxScrapeSize` (100 MB) after roughly **7–8 hours of pod uptime**, at which point:

```
cannot scrape target "http://10.60.160.2:9999/metrics"
  ({job="beckn-driver-offer-bpp-production", pod="...-759f5d4f48-5r5x4"})
  the last error: the response from ".../metrics" exceeds -promscrape.maxScrapeSize
  or max_scrape_size in the scrape config (100000000 bytes)
```

vmagent discards the **entire scrape** — it is all-or-nothing per target. So every `BPP_*` series from that pod vanishes, not just these histograms.

Measured on ny-prod on 2026-09-21:

- **75 of 125** driver-app pods were dark simultaneously. Counters aggregated across the fleet were reading low all day, biased toward the youngest pods.
- Failure tracks pod age almost perfectly:

  | Pod started (UTC) | failing | ok |
  |---|---|---|
  | ≤ 03:00 | 74 | 1 |
  | 06:00–07:00 | 1 | 3 |
  | ≥ 08:00 | 0 | 46 |

- **Total `BPP_*` blackout 01:35–04:10 IST.** HPA scales down to ~6 pods overnight, and the survivors are the *oldest* pods, so 100% were over the limit and `sum()` was a true zero.
- Metrics go **absent**, not to zero, so the gap is easy to mistake for genuinely quiet overnight traffic. `BPP_rider_acceptance_count` had no series at all for that window.

Per-pod series measured on a pod sitting right at the cliff:

```
BPP_congestion_charge_bucket     171,626
BPP_price_per_km_bucket          170,929
BPP_ride_distance_meters_bucket   12,142
BPP_congestion_charge_{sum,count}  4,186 each
BPP_price_per_km_{sum,count}       4,169 each
BPP_search_try_count                 866
BPP_rider_acceptance_count           604
BPP_booking_created_count            366
BPP_ride_created_count               356
BPP_ride_cancelled_count             320
BPP_ride_completed_count             285
```

The histograms are **~96% of the payload**. The zone-labelled plain counters are only ~2,800 series combined and are perfectly fine, so they keep their labels. This change takes a pod from ~390k samples to roughly ~20k — about a 20x margin under the limit.

### What we lose, and why that's the right trade

Zone breakdowns of *fare per km*, *congestion charge* and *distance distributions* go away. Zone breakdowns of the funnel remain, on `BPP_search_try_count`, `BPP_rider_acceptance_count`, `BPP_booking_created_count`, `BPP_ride_created_count`, `BPP_ride_started_count`, `BPP_ride_completed_count` and `BPP_ride_cancelled_count`.

That is a far better position than today, where the zone dimension exists in theory but the metrics are absent for most of the fleet most of the time — and entirely absent overnight.

### Alternatives considered

- **Raise `-promscrape.maxScrapeSize`** — does not fix it. The growth is unbounded, so this only moves the cliff further out while inflating vmagent/vminsert memory and VM cardinality.
- **Drop the metrics in a vmagent `metricRelabelConfig`** — *cannot* work. vmagent enforces the size limit while reading the response body, before any parsing or relabeling, so the drop rule never runs.
- **Scheduled rolling restarts to keep pods under the cliff** — a band-aid we can use until this ships, not a fix.

## How did you test it?

- `treefmt` and the repo pre-commit hook set pass.
- Change is mechanical: every edit was applied by exact-count string replacement (4 helper signatures, 4 label tuples, 8 call sites, 1 type, 1 registration), and the script fails loudly rather than half-applying if any count is off.
- `P.Label6` confirmed present in the pinned `prometheus-client`.
- The now-unused `(searchPickupZone, searchDropZone)` binding in `Domain/Action/Beckn/Search.hs` is removed; `mbAreaForVST` is still used further down, and the zone bindings in `EndRide/Internal.hs` and `SharedLogic/Ride.hs` are retained because the counters there still need them.

**Not compiled locally** — a cold nix eval wasn't feasible in this environment, so I'm relying on CI for the build. Worth a reviewer's eye on the arity changes in particular.

Verification once deployed: `count(up{job="beckn-driver-offer-bpp-production"} == 0)` should stay at 0 regardless of pod age, and `max(scrape_samples_scraped{job="beckn-driver-offer-bpp-production"})` should settle well under 100k instead of climbing to ~390k and censoring.

### Follow-ups (not in this PR)

- There is no alert on this. `up{job=~"beckn-.*"} == 0` has been firing silently — worth adding.
- Same class of failure on `envoy-stats-ingressgateway` (>100 MB) and `envoy-stats` (istio sidecars hitting `sample_limit=2000`) for `beckn-osrm` and `beckn-location-tracking-service`.

## Summary by CodeRabbit

* **Monitoring**
  * Ride-value metrics now use streamlined dimensions, omitting pickup and drop-off zone labels.
  * Search, pre-ride, and completed-ride metrics continue tracking pricing, congestion, ride distance, and pickup distance with consistent stage and service details.


piyushKumar-1:
Companion PR for `prodHotPush-Common`: #16853 — identical diff, verified byte-for-byte. Both should land together.

PR #16867, opened 2026-09-22
fix: refuse document upload and register while the driver has a live ride

TejaswiniT167:
Issue: an upload re-derives the entity's onboarding flags, so replacing a document mid-trip could take an approved driver back to pending and disable them while they were carrying a passenger. Unlinking a document was already barred during a live ride; uploading one was not. Today the recompute silently freezes instead (`OnboardingFlagMutation` is already a live-ride verb), which leaves the document pending while the flags stay stale until some later recompute.

Fix: a new `UploadDocument` verb on the onboarding guard, added to `liveRideVerbs`, so the existing "no change under a live ride" rule covers uploads. It is applied at two points, which together cover every surface:
- `validateImageHandler` — every image upload, from the admin dashboard, the fleet dashboard and the driver app/webview; the guard runs before the S3 write, so a refused upload leaves nothing behind.
- `postDriverRegistrationDocumentRegisterWithVerifiedBy` — a field-only document registers with no image and so never reaches the upload path (BusinessLicense, LegalEntityCompanyDetails and LegalEntityTAXDetails are configured that way, all mandatory and approval-supporting). The driver webview's `/ui/driver/document/register` delegates into this same handler.

Scoping: a vehicle document resolves to its RC, so the check follows the driver actually on the trip rather than the fleet owner named in the request. A fleet owner's own documents are exempt, because that check spans every active driver in the fleet and one driver on a trip would otherwise block the owner all day.

`noLiveRide`'s message drops "cannot change association", since it now also reaches an operator uploading a document.

## Summary by CodeRabbit

* **Bug Fixes**
  * Document uploads now consistently respect onboarding permissions during driver registration and image uploads.
  * Unauthorized document uploads are blocked instead of bypassing onboarding restrictions.
  * Document uploads are rejected while a live ride is in progress.
  * Fleet-owner identity document uploads without a registration certificate continue to be supported.
* **User Experience**
  * Live-ride restriction messages now provide clearer, more concise wording.


TejaswiniT167:
Superseded: instead of guarding uploads during a live ride, the admin auto-approve moves into the backend so register and approve run inside one withOnboardingAction — the flags are recomputed once, with the document already approved, so they never dip. New branch to follow.

